PR #46, opened 2014-03-22
Branch containing all 3.0 changes

mathiasverraes:


src/Currency.php
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Checks whether this currency is available in the passed context
+     *
+     * @param AvailableCurrencies $currencies
+     *
+     * @return boolean
+     */
+    public function isAvailableWithin(AvailableCurrencies $currencies)
+    {
+        return $currencies->exists($this);
```
mitchellolsthoorn: The method *exists()* does not exist in *AvailableCurrencies*. This should be *contains()*

sagikazarmark:
These two branches are quite far from each other now. Should we just leave it that way, move next release to master and master to 1.x, or try to rebase?

I would still create an 1.x branch anyway, so not sure merging this worths the hassle.

Thoughts?

sagikazarmark:
I think this PR became large and completely unmergable, so I am closing it.

PR #81, opened 2015-02-03
Implemented \JsonSerializable

tylercrompton:
This addresses #5 partially or fully depending on what you had in mind. This works great for JSON API's that use this library.

### DIFF
--- a/lib/Money/Money.php
+++ b/lib/Money/Money.php
@@ -303,11 +303,11 @@ class Money implements \JsonSerializable
      *
      * @return array
      */
-    function jsonSerialize()
+    public function jsonSerialize()
     {
         return array(
             'amount' => $this->amount,
-            'currency' => strval($this->currency),
+            'currency' => $this->currency,
         );
     }
 }
--- a/lib/Money/Money.php
+++ b/lib/Money/Money.php
@@ -10,7 +10,7 @@
 
 namespace Money;
 
-class Money
+class Money implements \JsonSerializable
 {
     const ROUND_HALF_UP = PHP_ROUND_HALF_UP;
     const ROUND_HALF_DOWN = PHP_ROUND_HALF_DOWN;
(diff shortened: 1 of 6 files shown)


sergeyklay:
Hi!

@tylercrompton Can you please add `JsonSerializable` interface class for passing php-5.3 tests?

```php
<?php
interface JsonSerializable {
    public function jsonSerialize();
}
```

lib/Money/Money.php
```diff
@@ -297,4 +297,17 @@ public static function stringToUnits( $string )
 
         return (int) $units;
     }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @return array
+     */
+    function jsonSerialize()
```
sergeyklay: also use visibility here, like `public function jsonSerialize()`  to follow the general coding style and PSR standard

tylercrompton:
Right. I forgot `JsonSerializable` isn't in PHP 5.3. Eh. I would actually consider upgrading the PHP dependency to 5.4+ instead. [PHP 5.3 hasn't even have security support for nearly six months now.](http://php.net/supported-versions.php)

sergeyklay:
To pass the php-5.3 test is sufficient to add the `\JsonSerializable` stub and using `spl_autoload_register` function in phpunit bootstrap file:

```php
spl_autoload_register(function ($class) {
    include 'money/tree/master/tests/' . $class . '.php';
});
```

tylercrompton:
Personally, I prefer not to add support for a version of a language that the implementers no longer support. Besides, the test would pass with aforementioned change but the code would still fail in production. In order to add support, we'd have to provide a custom `json_encode` function that would handle objects implementing `\JsonSerializable`. That just seems like too much to me.

mathiasverraes:
Thanks, see https://github.com/mathiasverraes/money/pull/86

mathiasverraes:
The minimum requirement for the `nextrelease` branch will be php 5.4

sergeyklay:
:+1:

PR #83, opened 2015-02-06
fix division by zero issue

### LINKED ISSUE #82: Division by zero
So...

    <?php

    require_once "vendor/autoload.php";

    $tenDollars = Money\Money::USD(10);

    $divided = $tenDollars->divide(0);

    echo $divided->getAmount();

... produces 0 and a warning, no exception. Is that by design?

lbialy:
Fixes #82, pushing against master, as nextrelease will break BC and this fixes current issue.

### DIFF
--- a/lib/Money/Money.php
+++ b/lib/Money/Money.php
@@ -182,12 +182,18 @@ class Money
     }
 
     /**
+     * @param int|float $operand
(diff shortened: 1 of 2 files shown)


mathiasverraes:
thanks, merged.

PR #90, opened 2015-03-07
Move currencies load out of construct

fabiocarneiro:
it will be nice to have the currencies available in a static method, so we can integrate the same currency list elsewhere by using the ```Currency``` API rather than loading the file manually (which would cause path problems).

__If we add this method, the include will be always relative to this module directory, no matter if the module is the /vendor/foo/bar/Money or /vendor/foo/bar/baz/Money__, plus it is fully backwards compatible, so... why not?

### DIFF
--- a/lib/Money/Currency.php
+++ b/lib/Money/Currency.php
@@ -24,9 +24,7 @@ class Currency
      */
     public function __construct($name)
     {
-        if(!isset(static::$currencies)) {
-           static::$currencies = require __DIR__.'/currencies.php';
-        }
+        static::getCurrencies();
 
         if (!array_key_exists($name, static::$currencies)) {
             throw new UnknownCurrencyException($name);
@@ -34,6 +32,17 @@ class Currency
         $this->name = $name;
     }
 
+    /**
+     * @return array
+     */
+    public static function getCurrencies()
+    {
+        if(!isset(static::$currencies)) {
+            static::$currencies = require __DIR__.'/currencies.php';
+        }
+        
+        return static::$currencies;
+    }
 
     /**
      * @return string
--- a/lib/Money/Currency.php
+++ b/lib/Money/Currency.php
@@ -24,11 +24,12 @@ class Currency
      */
     public function __construct($name)
     {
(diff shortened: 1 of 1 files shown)


fabiocarneiro:
@mathiasverraes ping :dancers:

fabiocarneiro:
@mathiasverraes I've been watching your discussion with @danizord in twitter about this PR, and I totally agree with your point that you should define the application currencies somewhere instead of using the ones provided by Money. But...

The use case we're working on is a module that integrates Money with doctrine and zf2, and we would like to provide in this case the list of all available currencies in Money. So if this module is working directly with money, it makes sense to it know the money currency limitations/possibilities. It doesn't necessarily means they'll be all used in the application, and i'm not sure this limitation is up to you to force. Having a feature doesn't mean it'll be even used. The application developer will know he doesn't accept all those currencies and will probably prefer to have a list of supported ones instead of having a full list of useless things.

Although there is a method in nextrelease that makes available to ask if some currency is available or not, there is no method in the current one that does that and that is the one we're using right now. I propose you to merge this and release 1.3. I still think this could be included in both, but maybe it could be removed for the next one.

That said, i'll say again that it would be nice to have this list available so we could integrate with the money form element in our module and it would be a fallback in case the developer doesn't specify the currency list.

sagikazarmark:
Currency lists have been completely removed from Money. No more currency code checking. If you need to work with a set of currencies, you have to implement `Currencies` interface.

fabiocarneiro:
@sagikazarmark That is still valid for the current release.

sagikazarmark:
@fabiocarneiro Indeed it is :)

sagikazarmark:
The current master is likely to be renamed to 1.x and maintained for some time. Since this PR makes sense on that branch, I am merging it.

Thank you @fabiocarneiro

danizord:
@sagikazarmark many thanks! Can you tag a release for it?

sagikazarmark:
Will do as soon as I get some time. Please follow #109 for details.

PR #92, opened 2015-04-08
IDE hints for magic methods in Money

davidkalosi:
Hi,

saw the issue and was tired of waiting and nervous of getting IDE warnings so here it is ;)

david


sagikazarmark:
Currencies are not necessarily ISO currencies. They can be anything. Maybe this makes things more comfortable, but don't think it should be added to the core class.

davidkalosi:
can you please elaborate on this ? alter all they are just doc blocks so I really do not see a reason. and again they could be even auto generated in the build process by iterating over the list of known currencies

sagikazarmark:
The point is that the possible types of currencies are not limited. You can use 'gold', 'bitcoin' or anything else as "currency code". ISO currencies are bundled with this repository because they are the most common forms of currency code. The library would work just fine without the included currency list.

So you can't even auto generate them, since there is no limitation on the currency code.

davidkalosi:
the only side effect is that you won't get IDE hints for them. ISO currencies are used mostly anyway and they seldom change.

sagikazarmark:
I am reluctant to accept that as an argument. It maybe uncomfortable, but I don't think including it is right.

davidkalosi:
never mind. I would at least expect that the maintainers of the library to have aligned opinions.   
see https://github.com/mathiasverraes/money/issues/70

sagikazarmark:
For the record: I am not a maintainer of this repository, so it is not my call.

mathiasverraes:
@sagikazarmark 

>> For the record: I am not a maintainer of this repository, so it is not my call.

You now have push permissions. Thanks for your efforts in this project.

sagikazarmark:
@mathiasverraes Yeah, got the notification, came out of nowhere. :smile: 

Thanks for the trust.

mathiasverraes:
I feel I'm more of a bottleneck than an asset to this project, so I feel it's better to share responsibilities. Let me know if you need any assistance or advice.

sagikazarmark:
Will do, thanks. I will try to move forward with the big money idea if I have the time.

sagikazarmark:
@mathiasverraes I stated above I would be happier without this. We have been accused with "not having aligned opinions" (which is not a major sin I think). So what's your opinion before closing this without having our opinions aligned?

davidkalosi:
Your call, you can reject this PR. I am not using this library anymore due to numerous other reasons (like fractions of cents etc)

sagikazarmark:
Thanks anyway @davidkalosi, contribution to OSS is always appreciated.

PR #108, opened 2015-09-23
Simplify if statement

beernjazz:
Hi,

This if statement can be simplified for php.

### DIFF
--- a/lib/Money/Money.php
+++ b/lib/Money/Money.php
@@ -225,7 +225,7 @@ class Money
      */
     public function divide($divisor, $rounding_mode = self::ROUND_HALF_UP)
     {
-        if ($divisor === 0 || $divisor === 0.0){
+        if ($divisor == 0){
             throw new InvalidArgumentException('Division by zero');
         }
 


mathiasverraes:
True, but it is not explicit that we specifically accept both.

camspiers:
I don't think this should be changed. The existing code will give a much better error message in the case that you pass through something that `== 0`. The assertOperand function covers the type check.

As an example, if you pass through `false` with your code it will say 'Division by zero'.

camspiers:
Here are some more examples:

```
cam-mbp:~ cam$ php -r "var_dump('' == 0);"
bool(true)
cam-mbp:~ cam$ php -r "var_dump(false == 0);"
bool(true)
cam-mbp:~ cam$ php -r "var_dump(null == 0);"
bool(true)
```

camspiers:
The woes of type juggling

camspiers:
@mathiasverraes I just noticed something that should be fixed though :). On line 232 we are still passing in `true` from when we used to pass in the division boolean.

```
$this->assertOperand($divisor, true);
```

PR #112, opened 2015-12-08
Add script to update currency list

thewilkybarkid:
This adds a quick-and-dirty script to update the currency list from the source (and so shows some differences).

I moved the now-generated file out of `src` to save having to ignore it from code coverage too.


thewilkybarkid:
If #92 is accept, that this should be expanded to generate the list.

bin/update_currencies.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
```
sagikazarmark: Can you please add file headers? I have the feeling they are going to be removed from the project entirely, but should be consistent until then.

bin/update_currencies.php
```diff
@@ -0,0 +1,80 @@
...
+use Buzz\Client\ClientInterface;
+use Buzz\Message\Request;
+use Buzz\Message\Response;
+
+require_once __DIR__.'/../vendor/autoload.php';
+
+final class CurrencyUpdater
```
sagikazarmark: Although I am a big fan of OOP, do we really need an updater object here?

bin/update_currencies.php
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private function request($uri, $method = Request::METHOD_GET)
+    {
+        $parts = parse_url($uri);
+
+        $request = new Request($method, $parts['path'], $parts['scheme'].'://'.$parts['host']);
+        $response = new Response();
+
+        $this->client->send($request, $response);
```
sagikazarmark: Not sure how Buzz actually works, but shouldn't we handle some possible errors here? No internet connection, and so on.

sagikazarmark:
This is nice @thewilkybarkid thanks. Can you please answer to the few comments above?

bin/update_currencies.php
thewilkybarkid: Buzz throws an exception at this point. Anything other than a valid JSON response will cause an exception in `loadJson()`. Of course if they return a JSON response with some other content then... don't commit the change. :)

bin/update_currencies.php
thewilkybarkid: Not really (and I didn't put much thought into it). Just there to separate things out a bit.

bin/update_currencies.php
sagikazarmark: Yeah, I meant this: not sure if we want to see exceptions, but an error message stating what happened.

thewilkybarkid: Well, it's only something that a maintainer would run, so I felt it was clear enough.

bin/update_currencies.php
sagikazarmark: Yeah, probably a little but over separated. Since there isn't any control in the flow.

bin/update_currencies.php
thewilkybarkid: Sorted (complete with out-of-date year :wink:)

bin/update_currencies.php
sagikazarmark: Fair enough. Okay, leave it this way.

sagikazarmark:
I think this is OK.

sagikazarmark:
Thanks @thewilkybarkid

PR #114, opened 2015-12-11
[FEATURE] Describe integration with Doctrine

afoeder:


doc/Integration.rst
```diff
@@ -0,0 +1,67 @@
...
+
+.. code-block:: yaml
+
+    Money\Money:
+      type: embeddable
+      table: money # the persistence table name you'd like to use
+
+      embedded:
+        'currency':
+          class: '\Money\Currency'
```
sagikazarmark: I wonder if the quotes are necessary. Sometimes you use them, sometimes not. Also, in this particular case: the slash in the class name is inconsistent with the rest of the docs.

doc/Integration.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Doctrine integration
+--------------------
+
+Independently from your preferred framework's characteristics, it is also possible to use plain
+Doctrine with the Money VO, using its Embeddables_ feature.
+
+.. _Embeddables: http://doctrine-orm.readthedocs.org/projects/doctrine-orm/en/latest/tutorials/embeddables.html
+
+Since you should not use annotations on the third party's library files, you'll want to use the `YAML mapping`_.
```
sagikazarmark: Although I prefer YAML as well, user can still use XML as well, right? Not sure why, but it seems XML is some kind of standard out there.

doc/Integration.rst
afoeder: usually I use the quotes whenever a value is "arbitrary" and not somehow schema-defined... but you're right, I don't do that consistently here...

sagikazarmark:
Thank you.

A few more things I noticed are commented. Also, the indentation seems to be inconsistend: sometimes 2 space, sometimes 4.

doc/Integration.rst
afoeder: I am totally not familiar with the XML mapping type (see http://doctrine-orm.readthedocs.org/en/latest/reference/xml-mapping.html#example for an example); this would be a typical case where I'd point my finger to another contributor to augment this documentation when he stumbles upon this and adapts it on his own for XML.

Additionally, I tend to argue that for Doctrine mapping configuration, the usual way is Annotations / YAML / XML...

afoeder:
The indendation is because of in the first case (2 indents) it's yaml standardd, in the other case (4) it's Symfony standard / convention... wdyt?

doc/Integration.rst
sagikazarmark: Sorry, after rereading my comment I relized that I was totally unclear: I meant that XML could also be mentioned, the configuration itself is perfectly fine in YAML.

sagikazarmark:
Makes sense. Decent IDE will reindent it anyway.

doc/Integration.rst
afoeder: hehe yep I see, makes also sense now when knowing your POV... I'll adjust it.

afoeder:
Adjusted everything @sagikazarmark

sagikazarmark:
Cool, thanks. And even rebased. :+1:

afoeder:
Thanks too, nice *joy*. Have a fine weekend!

Issue #111, opened 2015-10-07
How to deal with integer limits?


mnapoli:
Hi, we have to deal with Money and store it in database. In that process we came up with a few issues that I'd like to discuss so that maybe we can find "best practices" on how to deal with all that.

First there's the problem of integer limits in memory. On a 32bits system the amount will be stored as an integer up to something like 21 million $/€/… (because it's stored as cents). That's not much and can be a real problem in some applications, I wonder if that should be mentioned in the README? Is the `Money` class only appropriate for 64b systems (where the limit is much less of a problem for *most* cases)? Should the amount be stored differently on 32b? (for our team it won't be a problem because we use only 64 bits systems)

Then there's the question of how to store it in database. For example in MySQL you can use the integer type which has the same limit as a 32b PHP integer. So obviously it's much better to use BIGINT which is the equivalent of a 64b integer.

However if you use Doctrine (quite popular in the PHP world), the entity manager will map a SQL BIGINT to a PHP **string** in order to be compatible with 32b systems… So Money will end up with a string instead of an int as the amount, which is an issue I guess because the class expects to have an int there (I expect things to break/not behave correctly). So we have several options, like storing the whole Money object serialized (we see that the class even implements `Serializable` in the next version) but it comes with a lot of other issues, creating custom Doctrine types to map SQL BIGINT to actual PHP integers (i.e. it assumes a 64b system is used), store the object as a string (e.g. `12 EUR`), etc. Is there some kind of best practices for that? Obviously storing as integer/bigint is great because it allows filtering, ordering, etc in SQL queries, but is that a big "no no" (just like storing money in float is a big "no no")?

camspiers:
You are right in everything you say. Unrestricted integer support has been discussed before and would be possible with something like bcmath (https://github.com/mathiasverraes/money/issues/7), but it just hasn't been implemented.

I think that we should add a clear message in the README that money sizes are limited by integer size in your environment (32bit vs 64bit), and if you are using a database the same holds true for that.

There are also some other things to be aware of, if you look over at the @sebastianbergmann implementation of Money, I added a bunch of failure states when arithmetic on money would cause loss of precision (and sometimes completely crazy values).

https://github.com/sebastianbergmann/money/commits?author=camspiers

These changes should probably be ported over to this library too. These problems become particularly pernicious when dealing in currencies like bitcoin, but they are also a huge problem if people don't validate ecommerce inputs well.

[...]

PR #115, opened 2015-12-23
deal with integer limits

frederikbosch:
Solves #111.

Probably has quite some impact on the library. Review carefully. Rather see some more comments on the design decisions than on some PSR-2 compliance. The latter will be solved - if there is any compliance oversights - when there is agreement on the design.

Hard aspect of this PR was rounding in GMP and BC Math libraries. Had a look at some other libraries for this to implement.

### DIFF
--- a/src/Calculator.php
+++ b/src/Calculator.php
@@ -2,13 +2,16 @@
 namespace Money;
 
 /**
- * Interface Calculator
- * @package Money
+ * Calculator Interface
+ * @author Frederik Bosch
  */
 interface Calculator
 {
 
     /**
+     * Returns whether the calculator is supported in
+     * the current server environment
+     *
      * @return bool
      */
     public static function supported();
--- a/src/Calculator.php
+++ b/src/Calculator.php
@@ -14,6 +14,8 @@ interface Calculator
     public static function supported();
 
     /**
+     * Compare a to b
+     *
      * @param int|string $a
      * @param int|string $b
      * @return int
@@ -21,53 +23,68 @@ interface Calculator
     public function compare($a, $b);
 
     /**
+     * Add added to amount
+     *
      * @param int|string $amount
      * @param int|string $addend
-     * @return mixed
+     * @return int|string
      */
     public function add($amount, $addend);
 
     /**
+     * Subtract subtrahend from amount
      * @param int|string $amount
      * @param int|string $subtrahend
-     * @return mixed
+     * @return int|string
      */
     public function subtract($amount, $subtrahend);
 
     /**
+     * Multiply amount with multiplier
+     *
      * @param int|string $amount
      * @param int|float $multiplier
-     * @return mixed
+     * @return int|string
      */
     public function multiply($amount, $multiplier);
 
     /**
+     * Divide amount with divisor
+     *
      * @param int|string $amount
      * @param int|float $divisor
-     * @return mixed
+     * @return int|string
      */
     public function divide($amount, $divisor);
 
     /**
+     * Round number to following integer
+     *
      * @param int|string|float $number
-     * @return mixed
+     * @return int|string
      */
     public function ceil($number);
 
     /**
+     * Round number to preceding integer
+     *
      * @param int|string|float $number
-     * @return mixed
+     * @return int|string
      */
     public function floor($number);
 
     /**
+     * Round number, use rounding mode for tie-breaker
+     *
      * @param int|string|float $number
      * @param int|string $roundingMode
-     * @return mixed
+     * @return int|string
      */
     public function round($number, $roundingMode);
 
     /**
+     * Share amount among ratio / total portions
+     *
      * @param int|string $amount
      * @param int|float $ratio
      * @param int|float $total
--- a/src/Calculator.php
+++ b/src/Calculator.php
@@ -1,0 +1,78 @@
+<?php
+namespace Money;
+
+/**
+ * Interface Calculator
+ * @package Money
+ */
+interface Calculator
+{
+
+    /**
+     * @return bool
+     */
+    public static function supported();
+
+    /**
+     * @param int|string $a
+     * @param int|string $b
+     * @return int
+     */
+    public function compare($a, $b);
+
+    /**
+     * @param int|string $amount
+     * @param int|string $addend
+     * @return mixed
+     */
+    public function add($amount, $addend);
+
+    /**
+     * @param int|string $amount
+     * @param int|string $subtrahend
+     * @return mixed
+     */
+    public function subtract($amount, $subtrahend);
+
+    /**
+     * @param int|string $amount
+     * @param int|float $multiplier
+     * @return mixed
+     */
+    public function multiply($amount, $multiplier);
+
+    /**
+     * @param int|string $amount
+     * @param int|float $divisor
+     * @return mixed
+     */
+    public function divide($amount, $divisor);
+
+    /**
+     * @param int|string|float $number
+     * @return mixed
+     */
+    public function ceil($number);
+
+    /**
+     * @param int|string|float $number
+     * @return mixed
+     */
+    public function floor($number);
+
+    /**
+     * @param int|string|float $number
+     * @param int|string $roundingMode
+     * @return mixed
+     */
+    public function round($number, $roundingMode);
+
+    /**
+     * @param int|string $amount
+     * @param int|float $ratio
+     * @param int|float $total
+     * @return int|string
+     */
+    public function share($amount, $ratio, $total);
+
+}
--- a/src/Calculator/BcMathCalculator.php
+++ b/src/Calculator/BcMathCalculator.php
@@ -26,8 +26,6 @@ final class BcMathCalculator implements Calculator
 
     /**
      * {@inheritdoc}
-     *
-     * @return bool
      */
     public static function supported()
     {
@@ -36,10 +34,6 @@ final class BcMathCalculator implements Calculator
 
     /**
      * {@inheritdoc}
-     *
-     * @param $a
-     * @param $b
-     * @return int
      */
     public function compare($a, $b)
     {
@@ -48,10 +42,6 @@ final class BcMathCalculator implements Calculator
 
     /**
      * {@inheritdoc}
-     *
-     * @param $amount
-     * @param $addend
-     * @return string
      */
     public function add($amount, $addend)
     {
@@ -72,10 +62,6 @@ final class BcMathCalculator implements Calculator
 
     /**
      * {@inheritdoc}
-     *
-     * @param $amount
-     * @param $multiplier
-     * @return string
      */
     public function multiply($amount, $multiplier)
     {
@@ -84,10 +70,6 @@ final class BcMathCalculator implements Calculator
 
     /**
      * {@inheritdoc}
-     *
-     * @param $amount
-     * @param $divisor
-     * @return string
      */
     public function divide($amount, $divisor)
     {
@@ -96,9 +78,6 @@ final class BcMathCalculator implements Calculator
 
     /**
      * {@inheritdoc}
-     *
-     * @param $number
-     * @return string
      */
     public function ceil($number)
     {
@@ -112,9 +91,6 @@ final class BcMathCalculator implements Calculator
 
     /**
      * {@inheritdoc}
-     *
-     * @param $number
-     * @return string
      */
     public function floor($number)
     {
@@ -123,10 +99,6 @@ final class BcMathCalculator implements Calculator
 
     /**
      * {@inheritdoc}
-     *
-     * @param $number
-     * @param $roundingMode
-     * @return string
      */
     public function round($number, $roundingMode)
     {
@@ -181,11 +153,6 @@ final class BcMathCalculator implements Calculator
 
     /**
      * {@inheritdoc}
-     *
-     * @param $amount
-     * @param $ratio
-     * @param $total
-     * @return string
      */
     public function share($amount, $ratio, $total)
     {
--- a/src/Calculator/BcMathCalculator.php
+++ b/src/Calculator/BcMathCalculator.php
@@ -1,0 +1,175 @@
+<?php
+namespace Money\Calculator;
+
+use Money\Calculator;
+use Money\Money;
+use Money\Number;
+
+/**
+ * Class BcMathCalculator
+ * @package Money\Calculator
+ */
+final class BcMathCalculator implements Calculator
+{
+    /**
+     * @var string
+     */
+    private $scale;
+
+    /**
+     * @param int $scale
+     */
+    public function __construct($scale = 14)
+    {
+        $this->scale = $scale;
+    }
+
+    /**
+     * @return bool
+     */
+    public static function supported()
+    {
+        return extension_loaded('bcmath');
+    }
+
+    /**
+     * @param $a
+     * @param $b
+     * @return int
+     */
+    public function compare($a, $b)
+    {
+        return bccomp($a, $b);
+    }
+
+    /**
+     * @param $amount
+     * @param $addend
+     * @return mixed
+     */
+    public function add($amount, $addend)
+    {
+        return bcadd($amount, $addend);
+    }
+
+    /**
+     * @param $amount
+     * @param $subtrahend
+     * @return mixed
+     */
+    public function subtract($amount, $subtrahend)
+    {
+        return bcsub($amount, $subtrahend);
+    }
+
+    /**
+     * @param $amount
+     * @param $multiplier
+     * @return mixed
+     */
+    public function multiply($amount, $multiplier)
+    {
+        return bcmul($amount, $multiplier, $this->scale);
+    }
+
+    /**
+     * @param $amount
+     * @param $divisor
+     * @return mixed
+     */
+    public function divide($amount, $divisor)
+    {
+        return bcdiv($amount, $divisor, $this->scale);
+    }
+
+    /**
+     * @param $number
+     * @return mixed
+     */
+    public function ceil($number)
+    {
+        $decimalSeparatorPosition = strpos($number, '.');
+        if ($decimalSeparatorPosition === false) {
+            return $number;
+        }
+
+        return bcadd($number, '1', 0);
+    }
+
+    /**
+     * @param $number
+     * @return mixed
+     */
+    public function floor($number)
+    {
+        return bcadd($number, '0', 0);
+    }
+
+    /**
+     * @param $number
+     * @param $roundingMode
+     * @return mixed
+     */
+    public function round($number, $roundingMode)
+    {
+        $number = new Number($number);
+        if ($number->isDecimal() === false) {
+            return (string) $number;
+        }
+
+        if ($number->isHalf() === false) {
+            return $this->roundDigit($number);
+        }
+
+        if ($roundingMode === Money::ROUND_HALF_DOWN) {
+            return $this->floor((string) $number);
+        }
+
+        if ($roundingMode === Money::ROUND_HALF_UP) {
+            return $this->ceil((string) $number);
+        }
+
+        if ($roundingMode === Money::ROUND_HALF_EVEN) {
+            if ($number->isCurrentEven() === true) {
+                return $this->floor((string) $number);
+            } else {
+                return $this->ceil((string) $number);
+            }
+        }
+
+        if ($roundingMode === Money::ROUND_HALF_ODD) {
+            if ($number->isCurrentEven() === true) {
+                return $this->ceil((string) $number);
+            } else {
+                return $this->floor((string) $number);
+            }
+        }
+
+        throw new \InvalidArgumentException('Unknown rounding mode');
+    }
+
+    /**
+     * @param $number
+     * @return string
+     */
+    private function roundDigit(Number $number)
+    {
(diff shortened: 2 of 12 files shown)


frederikbosch:
In the next few days I will write something on the design decisions I made. This will help reviewing the PR (I guess).

sagikazarmark:
Sounds good. Merry Christmas :smile:

frederikbosch:
While creating this PR, I had the following things in mind.

1. It must be possible to change money calculations per math library. Therefore there should be an interface that has to be implemented per library: `CalculatorInterface`.
2. The calculators we are looking for are: PHP, BC and GMP.
3. A calculator is a static property of the `Money` value object. I cannot see any scenario where you want to mix calculators.
4. It should be possible to change the (static) calculator.
5. All calculations inside `Money` class should be replaced by calls to the active calculator
6. String are now allowed inside `Money` because libraries like GMP and BC express numbers as strings.
7. Checks that are related to native php calculations, like `assertIntegerBounds`, should be moved to the PhpCalculator.
8. When possible all calculators must implement all functionality that is now in use by the Money class

Things I came across into when creating this PR.

1. Both BC and GMP libraries do not have every rounding method build-in. The GMP library does not have rounding tie-breakers. But since we do not need extreme precision - we want (big) integers without decimals - rounding can be done by simple string conditionals.
2. The BC library does not convert '0.500000' to '0.5'
3. The GMP library does not recognize decimals at all, calculations like multiple should convert to numbers to non-decimals before calculation, and convert back after finishing calculations. This is done by removing the decimal sign (multiply by eX) and the reinsert the decimal sign after calculation.
4. Divisions in GMP are done by multiplying with 1 / divisor.

Please let me know whether you need more clarification.

sagikazarmark:
> `CalculatorInterface`

No need for the `Interface` suffix.

> It should be possible by using the API to change the (static) calculator.

Is there a reason for that? Does it make sense to allow the user to change the priority order we define?

> String are now allowed inside `Money` because

Soes it mean strings are allowed from the public API as well? I am not a big fan of the idea, because it requires more validation logic than necessary.

> When possible all calculators must implement all functionality that is now in use by the Money class

If we have an interface then this assumption is moved to the fact that the interface must be implemented by all calculators.


> The BC library does not convert '0.500000' to '0.5'

Is this a problem? With don't need decimal precision.

src/Calculator/BcMathCalculator.php
```diff
@@ -0,0 +1,188 @@
...
+namespace Money\Calculator;
+
+use InvalidArgumentException;
+use Money\CalculatorInterface;
+use Money\Money;
+use Money\Number;
+
+/**
+ * Class BcMathCalculator
+ * @package Money\Calculator
```
sagikazarmark: Please add yourself as an author (if you are OK with that) and remove the rest from here.

src/Calculator/BcMathCalculator.php
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * Copyright (c) 2011-2013 Mathias Verraes
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Money\Calculator;
+
+use InvalidArgumentException;
```
sagikazarmark: Classes from root namespace are OK to be used in the form of `\InvalidArgumentException`. Import statements like this does not make things easier to use/read.

src/Calculator/BcMathCalculator.php
```diff
@@ -0,0 +1,188 @@
...
+use Money\Money;
+use Money\Number;
+
+/**
+ * Class BcMathCalculator
+ * @package Money\Calculator
+ */
+final class BcMathCalculator implements CalculatorInterface
+{
+
```
sagikazarmark: -extra line

src/Calculator/BcMathCalculator.php
```diff
@@ -0,0 +1,188 @@
...
+    /**
+     * BcMathCalculator constructor.
+     * @param int $scale
+     */
+    public function __construct($scale = 14)
+    {
+        $this->scale = $scale;
+    }
+
```
sagikazarmark: -extra line

src/Calculator/BcMathCalculator.php
```diff
@@ -0,0 +1,188 @@
...
+final class BcMathCalculator implements CalculatorInterface
+{
+
+    /**
+     * @var string
+     */
+    private $scale;
+
+    /**
+     * BcMathCalculator constructor.
```
sagikazarmark: Unnecessary comment

src/Calculator/BcMathCalculator.php
```diff
@@ -0,0 +1,188 @@
...
+     */
+    public function supported()
+    {
+        return extension_loaded('bcmath');
+    }
+
+    /**
+     * @param $a
+     * @param $b
+     * @return int
```
sagikazarmark: Please add a new line before return statements

sagikazarmark: In case of interface implementations, use `{@inheritdoc}` please.

src/Calculator/BcMathCalculator.php
```diff
@@ -0,0 +1,188 @@
...
+        return bcdiv($amount, $divisor, $this->scale);
+    }
+
+    /**
+     * @param $number
+     * @return mixed
+     */
+    public function ceil($number)
+    {
+        $decimal_separator_position = strpos($number, '.');
```
sagikazarmark: Please use camelCase variables and add a new line after this statement to increase readability.

src/Calculator/GmpCalculator.php
```diff
@@ -0,0 +1,198 @@
...
+     * @param $amount
+     * @param $ratio
+     * @param $total
+     * @return mixed
+     */
+    public function share($amount, $ratio, $total)
+    {
+        return $this->floor($this->divide($this->multiply($amount, $ratio), $total));
+    }
+}
```
sagikazarmark: Missing newline

src/Calculator/BcMathCalculator.php
```diff
@@ -0,0 +1,188 @@
...
+     * @param $ratio
+     * @param $total
+     * @return mixed
+     */
+    public function share($amount, $ratio, $total)
+    {
+        return $this->floor(bcdiv(bcmul($amount, $ratio, $this->scale), $total, $this->scale));
+    }
+
+}
```
sagikazarmark: Missing newline

src/Calculator/PhpCalculator.php
```diff
@@ -0,0 +1,181 @@
...
+     *
+     * @param  numeric $amount
+     */
+    private function assertInteger($amount)
+    {
+        if (!is_int($amount)) {
+            throw new UnexpectedValueException('The result of arithmetic operation is not an integer');
+        }
+    }
+}
```
sagikazarmark: Missing newline

src/CalculatorInterface.php
```diff
@@ -0,0 +1,78 @@
...
+ * Interface CalculatorInterface
+ * @package Money
+ */
+interface CalculatorInterface
+{
+
+    /**
+     * @return bool
+     */
+    public function supported();
```
sagikazarmark: Can't we make this a static function? In that case we don't need to instantiate all calculators when detecting the calculator.

src/CalculatorInterface.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @param $a
+     * @param $b
+     * @return int
+     */
+    public function compare($a, $b);
+
+    /**
+     * @param $amount
```
sagikazarmark: types from `@param` phpdocs are missing

src/CalculatorInterface.php
```diff
@@ -0,0 +1,78 @@
+<?php
+namespace Money;
+
+/**
+ * Interface CalculatorInterface
+ * @package Money
+ */
+interface CalculatorInterface
```
sagikazarmark: Only Calculator, not CalculatorInterface.

src/Number.php
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * @return string
+     */
+    public function __toString()
+    {
+        return $this->number;
+    }
+
+}
```
sagikazarmark: Missing newline.

src/Money.php
```diff
@@ -539,4 +477,29 @@ public function jsonSerialize()
...
+    /**
+     * @param CalculatorInterface $calculator
+     */
+    public static function setCalculator(CalculatorInterface $calculator)
+    {
+        self::$calculator = $calculator;
+    }
+
+    /**
+     *
```
sagikazarmark: Empty phpdoc.

src/Money.php
```diff
@@ -539,4 +477,29 @@ public function jsonSerialize()
...
+     */
+    private static function assignCalculator()
```
sagikazarmark: Something like detect, initialize or find would fit better here IMO.

src/Money.php
```diff
@@ -539,4 +477,29 @@ public function jsonSerialize()
...
+    {
+        /** @var CalculatorInterface[] $calculators */
+        $calculators = [new BcMathCalculator(), new GmpCalculator(), new PhpCalculator()];
```
sagikazarmark: If the supported method would be static, we could even add possibility for custom calculators. Somthing like `Money::registerCalculator` with our calculators in the list out of the box.

src/CalculatorInterface.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @param $amount
+     * @param $ratio
+     * @param $total
+     * @return mixed
+     */
+    public function share($amount, $ratio, $total);
+
+}
```
sagikazarmark: Missing newline.

tests/BcMathCalculatorTest.php
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Money\Tests;
+
+use Money\Calculator\BcMathCalculator;
+use Money\Money;
+use PHPUnit_Framework_TestCase;
```
sagikazarmark: Unnecessary import

tests/BcMathCalculatorTest.php
```diff
@@ -0,0 +1,92 @@
...
+
+class BcMathCalculatorTest extends PHPUnit_Framework_TestCase
```
sagikazarmark: Can we create an abstract test case with an abstract `getCalculator` method to avoid duplication?

tests/BcMathCalculatorTest.php
```diff
@@ -0,0 +1,92 @@
+<?php
+/**
```
sagikazarmark: Missing newline.

sagikazarmark:
Nice PR @frederikbosch.

I added some comments, mostly related CS thing, but there are some design questions as well.

frederikbosch:
@sagikazarmark Thanks for the comments, fixed most of the issues you brought up. Regarding your questions.

1. No need for the Interface suffix. **Removed it.**

2. Does it make sense to allow the user to change the priority order we define? **Changed it to `registerCalculators` as you suggested**.

3. Does it mean strings are allowed from the public API as well? I am not a big fan of the idea, because it requires more validation logic than necessary. **The string is validated**

4. If we have an interface then this assumption is moved to the fact that the interface must be implemented by all calculators. **Right, but libraries have limitations themselves, like I said in my previous comments, e.g. GMP does not have rounding build-in. I tried to create a workaround for the limitations.**

5. Is this a problem? With don't need decimal precision. **It is not, but it should be [taken into account.](https://github.com/frederikbosch/money/blob/nextrelease/src/Number.php#L57)**

src/Money.php
```diff
@@ -539,4 +482,28 @@ public function jsonSerialize()
             'currency' => $this->currency,
         );
     }
+
+    /**
+     * @param Calculator[] $calculators
+     */
+    public static function registerCalculators(array $calculators)
```
sagikazarmark: I think having a singular method would be better. I would also add a check whether the class exists at all and implements our Calculator interface. Also, why default? It could be just calculators IMO

sagikazarmark: Also, a new calculator should be added to the beggining of the list to make sure custom calculators are checked first.

sagikazarmark:
> The string is validated

But is it necessary to be a public detail? IMO we can do it internally, but from a public API point of view, accepting integers is fine.

frederikbosch:
@sagikazarmark No, that is whole point of no integer limits, right? Big integer (bigger than php's max int) can only be expressed in string, not in integers.

sagikazarmark:
Sorry, missed the big integer point. You are right of course.

sagikazarmark:
> If we have an interface then this assumption is moved to the fact that the interface must be implemented by all calculators. Right, but libraries have limitations themselves, like I said in my previous comments, e.g. GMP does not have rounding build-in. I tried to create a workaround for the limitations.

If necessary, we can create interfaces to indicate supported features and act according to it. For example: RoundingCalculator has a round method. If calculator is not a RoundingCalculator fallback to something. This is just an example, not sure if in this case this is the right solution.

frederikbosch:
I dropped php 5.4 support, and added php 7. Is that alright?

sagikazarmark:
No, it's not. :)

frederikbosch:
Just restored it. My advice would be to drop it. Libraries should encourage people not to use a PHP version that has [no security support any more](http://php.net/supported-versions.php). Moreover, this the `nextrelease` branch, not the current master.

sagikazarmark:
I generally agree with you. But PHP 7 is just released and not yet available in the latest Ubuntu LTS. Not to mention, that most of applications will need 5.x support for at least a few years. I don't feel the urge to drop 5.x support right away.

frederikbosch:
But, you are right, that is at least out of the scope of this PR :)

frederikbosch:
The only things that are left are `{inheritDoc}` and `Can we create an abstract test case with an abstract getCalculator method to avoid duplication?`. The rest should be done.

sagikazarmark:
> The only things that are left are {inheritDoc}

Lowercase please to be consistent with the rest of the library.

sagikazarmark:
There are some other inline comments left.

frederikbosch:
@sagikazarmark Ok, I will do it somewhere next week. Regarding the other comments, they should be fixed, I do not know why it is still there.

frederikbosch:
Hmm, my git says new lines after every class are committed, Github thinks different.

frederikbosch:
@sagikazarmark The trailing new line, I do not get it. It's not there when [viewing](https://github.com/frederikbosch/money/blob/nextrelease/src/Calculator.php) the file. It is there when [editing](https://github.com/frederikbosch/money/edit/nextrelease/src/Calculator.php) the file. What's going on?

sagikazarmark:
@frederikbosch Git trims it when you take a look at the file. In diffs, git marks a missing new line with an icon.

frederikbosch:
But is it alright now then?

sagikazarmark:
It is now, yes.

frederikbosch:
hmm, what do you want me to do with the conflicts?

frederikbosch:
I think it would be best that you merge it as is and fix what you want to fix regarding CS.

sagikazarmark:
Rebase the branch please. I think it's mostly comments, or stuff, what you moved to PhpCalculator.

frederikbosch:
Done, rebased.

sagikazarmark:
Have you pulled the nextrelease branch from upstream before rebasing?

frederikbosch:
No.

sagikazarmark:
You should.

frederikbosch:
I have no idea how. Is it not possible that you do this?

sagikazarmark:
It is, but in that case I have to close this PR and there are some things you wanted to do.

Here is how you can rebase the branch:

``` bash
git remote add upstream https://github.com/mathiasverraes/money.git
git fetch upstream
git checkout nextrelease
git rebase upstream/nextrelease
git push origin nextrelease #if this fails, it means you have commits in your own nextrelease branch, where you shouldn't have
git checkout <YOUR FEATURE BRANCH>
git rebase nextrelease
```

As soon as there is a conflict, the rebase will stop and you can fix the conflicts (if you use an IDE, you can use it, otherwise good old plain text). When you are done with all conflicts:

``` php
git add .
git rebase --continue
```

But you have to make sure that every conflict is resolved or you might commit unresolved conflicts into the library.

Another possibility is to pull your branch into the main repo an open a PR from there, but in this case, you still need to open a PR against that branch if you want further modifications.

sagikazarmark:
Ah, realized that I can help if you add me as a collaborator to your forked repo.

frederikbosch:
Just did that.

sagikazarmark:
Ok, rebased. Make sure to checkout your nextrelease branch as a new local branch.

frederikbosch:
@sagikazarmark Thanks! Will see if I can finish the PR today.

frederikbosch:
I think the PR should be ready. Added documentation to the `Calculator` class. Used `{@inheritdoc}` blocks in the calculator implementations.

src/Calculator.php
```diff
@@ -0,0 +1,95 @@
+<?php
+namespace Money;
```
sagikazarmark: Missing newline

src/Calculator.php
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Interface Calculator
+ * @package Money
```
sagikazarmark: Can you please remove package phpdoc and add yourself as an author (If you are OK with that)?

src/Calculator.php
```diff
@@ -0,0 +1,95 @@
...
+ */
+interface Calculator
+{
+
+    /**
+     * @return bool
```
sagikazarmark: I know it's kind of obvious, but can you please add some documentation here as well?

src/Calculator/BcMathCalculator.php
```diff
@@ -0,0 +1,195 @@
...
+     */
+    public static function supported()
+    {
+        return extension_loaded('bcmath');
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @param $a
```
sagikazarmark: Return and param docs are not necessary with inheritdoc.

src/Money.php
```diff
@@ -535,4 +477,28 @@ public function jsonSerialize()
             'currency' => $this->currency,
         ];
     }
+
+    /**
+     * @param Calculator $calculator
+     */
+    public static function registerCalculator(Calculator $calculator)
+    {
+        self::$calculators = $calculator + self::$calculators;
```
sagikazarmark: I would use an array_unshift here. Array additions can behave as a bitch.

src/Money.php
```diff
@@ -535,4 +477,28 @@ public function jsonSerialize()
...
+    }
+
+    /**
+     * @return Calculator
+     */
+    private static function initializeCalculator()
+    {
+        /** @var Calculator[] $calculators */
+        $calculators = self::$calculators;
```
sagikazarmark: I would remove this line and modify the docblock to `Calculator $calculator`

sagikazarmark:
Looks good to me. I added a few more CS related things, can you please correct those. The rest can be automatically fixed by CS fixer.

src/Calculator/BcMathCalculator.php
frederikbosch: I would not agree with this. The [phpDocumentor documentation](http://www.phpdoc.org/docs/latest/guides/inheritance.html) says "The inline tag {@inheritDoc} is used in a description to import the description of a parent element, even if the child element already has a description." So `{@inheritdoc}` is only relevant to the description. The current `Money` class also [contains inherited docblock with return docs](https://github.com/mathiasverraes/money/blob/nextrelease/src/Money.php#L529).

src/Calculator.php
frederikbosch: I don't get why this is necessary, but OK, I will add it.

sagikazarmark: Consistency.

frederikbosch:
Fixed things, except for the docblock. You sure you want to remove `@param` and `@return` there?

sagikazarmark:
For now, yes. I am aware of the inheritdoc recommended usage, but it is a common pattern to use it for all kinds of docblocks. IDEs recognize it, even if phpDocumentor says otherwise. There is demand for complete docblock inheritance, and phpDocumentor is working on it for some years now.

frederikbosch:
OK, removed the `@param` and `@return` blocks where `{@inheritdoc}` is used.

sagikazarmark:
Cool, thanks. Ready for merge then?

frederikbosch:
I would say so :)

sagikazarmark:
Thanks @frederikbosch

frederikbosch:
Great! Thanks @sagikazarmark for reviewing! Maybe we can create an alpha release? I would be happy with some sort of tag.

#111 can also be closed.

sagikazarmark:
#111 should automatically be closed.

Tagging sounds good, but we should first rebase/merge into master.

frederikbosch:
I would just rename branches in such cases ;) But I do not if that has any negative side-effects.

Issue #107, opened 2015-09-22
add 'unitsToString' function to Money\Money


johanwilfer:
For legacy systems I need to convert to/from floats. To convert I do the following:
```php
use Money\Money;
use Money\Currency;
$float = 123.34;
$money = new Money(Money::stringToUnits((string)$float), new Currency('USD'));
var_dump($money);
/*
object(Money\Money)#2 (2) {
  ["amount":"Money\Money":private]=>
  int(12334)
  ["currency":"Money\Money":private]=>
  object(Money\Currency)#3 (1) {
    ["name":"Money\Currency":private]=>
    string(3) "USD"
  }
}
*/

$float = $money->getAmount() / 100;
var_dump($float);
/*
float(123.34)
*/
```

For example money_format() expects the parameter to be a float: http://php.net/manual/en/function.money-format.php

Shouldn't it make sense to have a unitsToString() in `Money\Money`? Maybe even
unitsToFloat() and floatToUnits() .

I can create a pull-request for this if you would consider including this change.

jtallant:
I really think this would make a lot of sense. I always have to implement it myself. When dealing with money, you're going to need to display it in a user friendly way 99% of the time. The ruby lib supports it https://github.com/RubyMoney/money/blob/master/lib/money/money/formatting.rb

johanwilfer:
Ping @mathiasverraes

moon0326:
+1

camspiers:
https://github.com/sebastianbergmann/money/blob/master/src/Money.php#L131-L134
https://github.com/sebastianbergmann/money-intl-formatter/blob/master/src/IntlFormatter.php

When displaying currency values you should take into account locale as well. The php NumberFormatter class takes care of this. You should also take into account the currencies subunit when doing the division. For example with Yen you should be string 100 Yen as `int(100)` and therefore when you format and display you shouldn't be dividing by 100.

https://en.wikipedia.org/wiki/Denomination_(currency)#Subunit_and_super_unit

mathiasverraes:
Apologies for the neglect to this library. Formatting should, imho, be outside of the Money class, for example in a MoneyFormatting class. PRs welcome (and maintainers also welcome).

johanwilfer:
Formatting is one thing, but some way of getting back to a string (or float) should belong in the Money-class, shouldn't it? I can prepare a PR. I can also help with maintaining if you would like.

[...]

PR #123, opened 2016-01-06
MoneyFormatter #107

frederikbosch:
Adds requested feature in #107.

Considerations:
1. No additional dependencies added. If someone wants to use a [ISO4217 package](https://github.com/alcohol/iso4217), it will be easy. Just inject the correct values in the constructor of `IntlMoneyFormatter` (preferably via a factory class).
2. It should work for all different calculators, so we have to work with strings (not integers).
3. `ext/intl` is a suggestion, not a requirement. Though using `IntlMoneyFormatter` without `ext/intl` installed, will throw an exception.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,14 +18,16 @@
         "kriswallsmith/buzz": "~0.15",
         "phpunit/phpunit": "^4.5",
         "ext-bcmath": "*",
-        "ext-gmp": "*"
+        "ext-gmp": "*",
+        "ext-intl": "*"
     },
     "suggest": {
         "TheBigBrainsCompany/TbbcMoneyBundle": "Very complete Symfony2 bundle with support for Twig, Doctrine, Forms, ...",
         "pink-tie/money-bundle": "Pink-Tie's Symfony2 integration with Money library",
         "Sylius/SyliusMoneyBundle": "Sylius' Symfony2 integration with Money library",
         "ext-bcmath": "Calculate without integer limits",
-        "ext-gmp": "Calculate without integer limits"
+        "ext-gmp": "Calculate without integer limits",
+        "ext-intl": "Format Money objects with intl"
     },
     "autoload": {
         "psr-4": {
--- a/src/IntlMoneyFormatter.php
+++ b/src/IntlMoneyFormatter.php
@@ -1,0 +1,51 @@
+<?php
+namespace Money;
+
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -25,7 +25,8 @@
         "pink-tie/money-bundle": "Pink-Tie's Symfony2 integration with Money library",
         "Sylius/SyliusMoneyBundle": "Sylius' Symfony2 integration with Money library",
         "ext-bcmath": "Calculate without integer limits",
-        "ext-gmp": "Calculate without integer limits"
+        "ext-gmp": "Calculate without integer limits",
+        "ext-intl": "Format Money objects with intl"
```
sagikazarmark: should be a dev requirement for unit tests, like gmp and bcmath

frederikbosch:
My implementation is not correct, I just notice. The `NumberFormatter` should help me out with some `setTextAttribute` I think, but my first tries failed. Will look into it later.

frederikbosch:
Problem fixed, also added the `require-dev`. What do you think of implementation? I will update CS, according to my last PR, after the general idea is approved.

src/IntlMoneyFormatter.php
```diff
@@ -0,0 +1,52 @@
...
+     * @var int
+     */
+    private $digits;
+
+    /**
+     * IntlMoneyFormatter constructor.
+     * @param int $digits
+     * @param string $locale
+     */
+    public function __construct($locale, $digits = 2)
```
sagikazarmark: Is digit the same what I called subunit? If so I would rather call it fractionDigits following NumberFormatter convention. Also, I would not set it to 2 as default. Maybe it is common, but it depends on the user: one might not even support subunits (like rounds up cents and doesn't care about the rest)

frederikbosch:
@sagikazarmark Maybe you can merge this one too. It will help reviewing the `MoneyParser` PR. Or do you have any remarks left? I also just squased the commits.

sagikazarmark:
A few new lines are missing from the end of life. Apart from that :+1:

frederikbosch:
Done, and squashed again.

sagikazarmark:
Cool. :+1:

frederikbosch:
Great, thanks! I will create `MoneyParser` soon. Let's get this library to `3.0.0`!

sagikazarmark:
:+1:

PR #127, opened 2016-01-11
isAvailableWithin uses the currencies' contains method instead of exists

rgeraads:


sagikazarmark:
This is already addressed on the currencies branch:

https://github.com/mathiasverraes/money/compare/nextrelease...currencies

Would you mind submitting some tests there?

rgeraads:
Created #129 :)

PR #129, opened 2016-01-11
added test to check whether a currency is available within the list of currencies

rgeraads:


sagikazarmark:
Looks good, thanks.

PR #131, opened 2016-01-12
fix static calculator and some code style

piotrantosik:
Using static properties with eg. doctrine integration is not possible.
Some example: i use Money as embedded class and get it from entity. 
```php
$entity->getMoney()->greaterThanOrEqual($otherEntity->getMoney());
```
Currently it is impossible and generate fatal error because constructior is not triggered:
```php
Fatal error: Call to a member function compare() on null
```

### DIFF
--- a/src/Calculator.php
+++ b/src/Calculator.php
@@ -1,8 +1,10 @@
 <?php
+
 namespace Money;
 
 /**
  * Calculator Interface
+ *
  * @author Frederik Bosch
  */
 interface Calculator
@@ -94,5 +96,4 @@ interface Calculator
      * @return int|string
      */
     public function share($amount, $ratio, $total);
-
 }
--- a/src/Money.php
+++ b/src/Money.php
@@ -3,6 +3,7 @@
 namespace Money;
 
 use InvalidArgumentException;
+use RuntimeException;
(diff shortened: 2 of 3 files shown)


sagikazarmark:
I like the general idea, but I am unsure about instantiating a new calculator every time. Seems to be an overkill to me. How about falling back to static property in the non-static `getCalculator method`?

The object itself can still be stored in the money object, hence it will be the same reference.

I can imagine a static `globalCalculator` property and `getGlobalCalculator` methods.

What do you think?

piotrantosik:
I agree that to return to the static - it is better performance.
Please recheck code and give me a feedback.

sagikazarmark:
Sorry, I caused some confusion, I though you might want to keep the getCalculator method. If there is no separation, I think it is better to rename it back to calculator and getCalculator. Sorry about that.

I think the getCalculator does not have to be static, as we don't want to statically access it. The calculator property can and should be static.

:+1: Nice catch @piotrantosik

piotrantosik:
Ok, soo now is:
- private static calculator property
- private methot to get this property

And looks good for me :smile:

sagikazarmark:
LGTM, thank you.

PR #134, opened 2016-01-12
Money's internal amount should be a string

### LINKED ISSUE #133: Careful with string vs int
Now that there are no more integer limitations (#115), we have to be careful with `int` and `string` usage vs strict comparison. This is something that I did not cover in my PR but now realize that should be taken care of. Especially with these lines.

- [isZero](https://github.com/mathiasverraes/money/blob/nextrelease/src/Money.php#L421)
- [isPositive](https://github.com/mathiasverraes/money/blob/nextrelease/src/Money.php#L431)
- [isNegative](https://github.com/mathiasverraes/money/blob/nextrelease/src/Money.php#L441)
- [equals](https://github.com/mathiasverraes/money/blob/nextrelease/src/Money.php#L133)

These lines are now buggy, because the following will return false.

```php
(new Money(1, new Currency('EUR')))->equals(new Money('1', new Currency('EUR')))
```

I guess we should convert the amount to a string in the constructor with `(string)`. PR is on the way.

frederikbosch:
Fixes #133.

### DIFF
--- a/src/Calculator/PhpCalculator.php
+++ b/src/Calculator/PhpCalculator.php
@@ -34,7 +34,7 @@ final class PhpCalculator implements Calculator
 
         $this->assertInteger($result);
 
-        return $result;
+        return (string) $result;
     }
 
     /**
@@ -46,7 +46,7 @@ final class PhpCalculator implements Calculator
 
         $this->assertInteger($result);
 
-        return $result;
+        return (string) $result;
     }
 
     /**
@@ -58,7 +58,7 @@ final class PhpCalculator implements Calculator
 
         $this->assertIntegerBounds($result);
 
-        return $result;
+        return (string) $result;
     }
 
     /**
@@ -70,7 +70,7 @@ final class PhpCalculator implements Calculator
 
         $this->assertIntegerBounds($result);
 
-        return $result;
+        return (string) $result;
     }
 
     /**
@@ -134,7 +134,7 @@ final class PhpCalculator implements Calculator
     {
         $this->assertIntegerBounds($amount);
 
-        return intval($amount);
+        return (string) intval($amount);
     }
 
     /**
@@ -144,7 +144,7 @@ final class PhpCalculator implements Calculator
      */
     private function assertInteger($amount)
     {
-        if (!is_int($amount)) {
+        if (filter_var($amount, FILTER_VALIDATE_INT) === false) {
             throw new \UnexpectedValueException('The result of arithmetic operation is not an integer');
         }
     }
--- a/src/Money.php
+++ b/src/Money.php
@@ -47,7 +47,7 @@ final class Money implements JsonSerializable
     ];
 
     /**
-     * @param int      $amount   Amount, expressed in the smallest units of $currency (eg cents)
+     * @param int|string $amount   Amount, expressed in the smallest units of $currency (eg cents)
      * @param Currency $currency
      *
      * @throws InvalidArgumentException If amount is not integer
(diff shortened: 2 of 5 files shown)


frederikbosch:
In order to create a `MoneyParser` PR, I need this to be merged. I just improved more docblock and squashed my commit.

src/Calculator/PhpCalculator.php
```diff
@@ -124,6 +124,19 @@ private function assertIntegerBounds($amount)
...
 
     /**
+     * Casts an amount to string ensuring that the internal
+     * amount value of Money is a string
+     *
+     * @param int $amount
+     *
+     * @return int
+     */
+    private function castString($amount)
```
sagikazarmark: Do we really need a separate method for that?

frederikbosch: No, we really do not need it. Removed it, and squashed again.

src/Money.php
```diff
@@ -47,7 +47,7 @@
     ];
 
     /**
-     * @param int      $amount   Amount, expressed in the smallest units of $currency (eg cents)
+     * @param int|string $amount   Amount,expressed in the smallest units of $currency (eg cents)
```
sagikazarmark: I would say we need that space after the comma.

sagikazarmark:
Well, I tend accept this merge, but still need to get used to the fact that amount will be stored as string.

frederikbosch:
@sagikazarmark I agree. It is a big change. It was more comprehensive than I thought at first.

sagikazarmark:
Merging the deprecation PR caused some conflicts.

frederikbosch:
Yeah, and now I need to rebase? :) Still don't know how, sorry about that...

sagikazarmark:
No problem, I will do it.

sagikazarmark:
Conflicts solved in #136

josecelano:
Some months ago I was testing how to deal with bit integers and I did this fork:

https://github.com/josecelano/money/blob/master/lib/Money/BigMoney.php

Instead of using integer or string for the amount I used another class. In this implementation Money has not the responsibility to deal with bit integers and it delegates to another class. The only problem is that library uses only `bcmath` extension. But maybe there is another similar class which uses `calculators`as you did it.

frederikbosch:
@josecelano `bcmath` is the only option for `PreciseMoney`. I rather use that term than `BigMoney`, because we already have `BigMoney` because there is no more integer limit. `GMP` only uses integers, not decimals, see the [RFC for floating points integration](https://wiki.php.net/rfc/gmp-floating-point) in PHP.

The solution I will propose soon will have the following structure.
- `PreciseCalculator` with methods like `roundPrecisely`
- `PreciseMoney`
- `BcMathCalculator` implementing `PreciseCalculator`
- I still have to think have to deal with formatting and parsing.

When the RFC for `GMP` lands, we only have to implement `PreciseCalculator` for GMP.

josecelano:
Sorry,there was definitely a misunderstanding. What I called BigMoney is an implementation of Money which allows big integers, not decimals. I used Decimal class but internally I force the constrain of not using decimals, I mean all values would have no deciaml part even if that class allows it. But the idea I want to expose is you could use another class to store the amount and delegate to that class the responsabiliti of dealing with bit integers. In fact, I was looking for a `BigInteger` class but I do nto remember why finally I decided to use `Decimal` for that test.

sagikazarmark:
Yeah, BigMoney casued a lot of confusion in me as well. But it refers to the precision of the value, not the integer size.

Issue #139, opened 2016-01-16
Umpirsky/currency-list won't work


sagikazarmark:
@umpirsky Will this work when this package is installed as a dependency?

https://github.com/mathiasverraes/money/blob/nextrelease/src/ISOCurrencies.php#L21

I think not.

The path in that case will be:

`__DIR__.'/../../../umpirsky/currency-list/data/en/currency.php'`

umpirsky:
@sagikazarmark Good catch, thanks!

Proposed fix in https://github.com/mathiasverraes/money/pull/140.

PR #140, opened 2016-01-16
Fixes issue #139

umpirsky:


sagikazarmark:
Wonder if require_once and/or caching loaded currencies in a static property could make it faster.

WDYT @umpirsky?

umpirsky:
@sagikazarmark Not sure about `require_once`, but caching loaded currencies can. Even if not static, it is a good idea to do it once.

sagikazarmark:
Ok, I'll open an issue for that, let's get this merged.

PR #142, opened 2016-01-16
Add Money exchange

sagikazarmark:
Support money exchange services with third-party data sources.

Added https://github.com/florianv/swap as a default implementation.

This also seem to solve florianv/swap#15. Although it is not a separate library, but I think still flexible from Money point of view: one is not tied to a specific implementation.


umpirsky:
:+1:

florianv:
:+1:

Issue #141, opened 2016-01-16
Load currency list once in ISO Currencies


sagikazarmark:

umpirsky:
@sagikazarmark Thanks, I am on it. :+1:

sagikazarmark:
:+1:

PR #144, opened 2016-01-16
Fixed issue #141

umpirsky:


sagikazarmark:
Even better what I thought. Currencies are not loaded when not used.

sagikazarmark:
Don't worry about the failing tests, it is caused because of some bad phpunit configuration. I will fix that.

umpirsky:
@sagikazarmark Thanks! :+1:

Issue #97, opened 2015-05-19
Bitcoin not supported by stringToUnits()


thedavidmeister:
Because stringToUnits() explicitly expects only 2 trailing decimals, it cannot support Bitcoin and we see "The value could not be parsed as money"

josecelano:
Hi @thedavidmeister maybe this could be interesting for you:

https://github.com/mathiasverraes/money/issues/7#issuecomment-91863422

thedavidmeister:
yeah, that does look interesting. I'm *just* about to roll a release for something I'm working on that uses the current Money setup, so I don't want to start a refactor now. But it's something worth following up as I had to jump through some hoops along the way that I would have expected Money to handle better.

frederikbosch:
The parser in #148 does not support bitcoin. My suggestion would be to close this issue after #148 is merged and leave implementation up to the user. A very simplistic approach could be as follows.

```php
<?php
use Money\Currency;
use Money\Money;
use Money\MoneyParser;

class BitcoinSupportedMoneyParser implements MoneyParser
{
    private $innerParser;

    public function __construct(MoneyParser $innerParser)
    {
        $this->innerParser = $innerParser;
    }

    public function parse($formattedMoney, $forceCurrency = null)
    {
        if (strpos($formattedMoney, "\0xC9\0x83") !== false) {
            // bitcoin parsing here
            return new Money($formattedMoney, new Currency('XBT'));
        } else {
            return $this->innerParser->parse($formattedMoney, $forceCurrency);
        }
    }
}
```

frederikbosch:
@sagikazarmark This one can be closed too

sagikazarmark:
Hm, should we include this in the repo?

frederikbosch:
Why would we want include it? By using the intl parser together with the parser interface, the user has all the possibilities to implement such a bitcoin parser of his own.

If we include this one, what is then the boundary what to include and what not?

frederikbosch:
And if we include a bitcoin parser, we should - by means of consistency - also include a formatter. 

Maybe we avoid getting same questions by including one. I am happy to PR both formatter and parser. Can you let me know what your decision is @sagikazarmark?

sagikazarmark:
Well, it seems bitcoin is a widely requested feature. Further more, it is plain PHP code without any dependencies. Given these two details, I don't think it's a problem to include it.

frederikbosch:
Ok, will PR next week.

frederikbosch:
[...]

PR #148, opened 2016-01-18
parse a string into a money object

### LINKED ISSUE #98: $ not supported by stringToUnits()
I tried to get '$100' to be some Money units using stringToUnits() but got "The value could not be parsed as money".

frederikbosch:
Fixes #98 and provides a possible solution for #97.

### DIFF
--- a/src/IntlMoneyParser.php
+++ b/src/IntlMoneyParser.php
@@ -1,0 +1,56 @@
+<?php
+namespace Money;
+
+/**
+ * Parses a string into a Money object using intl extension.
+ *
+ * @author Frederik Bosch <f.bosch@genkgo.nl>
+ */
+final class IntlMoneyParser implements MoneyParser
+{
+
+    /**
+     * @var \NumberFormatter
+     */
+    private $formatter;
+
+    /**
+     * @param \NumberFormatter $formatter
+     */
+    public function __construct(\NumberFormatter $formatter)
+    {
+        $this->formatter = $formatter;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function parse($formattedMoney, $forceCurrency = null)
+    {
+        $decimal = $this->formatter->parseCurrency($formattedMoney, $currency);
+        if ($decimal === false) {
+            throw new ParserException(
+                'Cannot parse ' . $formattedMoney . ' to Money. ' . $this->formatter->getErrorMessage()
+            );
+        }
+
+        $decimal = (string) $decimal;
+        if (strpos($decimal, '.') !== false) {
+            $decimal = str_replace('.', '', $decimal);
+        } else {
+            $decimal .= str_pad('', $this->formatter->getAttribute(\NumberFormatter::FRACTION_DIGITS), '0');
+        }
+
+        if (substr($decimal, 0, 1) === '-') {
+            $decimal = '-' . ltrim(substr($decimal, 1), '0');
+        } else {
+            $decimal = ltrim($decimal, '0');
+        }
+
+        if ($forceCurrency === null) {
+            return new Money($decimal, new Currency($currency));
+        }
+
+        return new Money($decimal, new Currency($forceCurrency));
+    }
+}
--- a/src/Money.php
+++ b/src/Money.php
@@ -426,35 +426,6 @@ final class Money implements \JsonSerializable
     }
 
     /**
-     * Creates units from string.
-     *
-     * @param string $string
-     *
-     * @return string
-     *
-     * @throws \InvalidArgumentException If $string cannot be parsed
-     */
-    public static function stringToUnits($string)
-    {
-        $sign = "(?P<sign>[-\+])?";
-        $digits = "(?P<digits>\d*)";
-        $separator = '(?P<separator>[.,])?';
-        $decimals = "(?P<decimal1>\d)?(?P<decimal2>\d)?";
-        $pattern = '/^'.$sign.$digits.$separator.$decimals.'$/';
-
-        if (!preg_match($pattern, trim($string), $matches)) {
-            throw new \InvalidArgumentException('The value could not be parsed as money');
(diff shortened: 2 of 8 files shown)


frederikbosch:
Hmm, why is this failing on Travis but succeeding on my machine.

barryvdh:
Could it be that the locale is not available on the Travis setup?

This gives the same (unclear) message: https://3v4l.org/1p9eF

sagikazarmark:
Travis only has en_US installed. The funny thing is that I didn't have this issue with/without nl_NL installed, I had totally different issues.

I would rewrite the test examples to use en_US locale, it is probably commonly installed for others anyway. I would also skip tests on machines where en_US locale is not available.

frederikbosch:
@sagikazarmark Right, we should do that. But I doubt if people will have same results when using the classes. That might lead to too many unexpected results.

sagikazarmark:
> That might lead to too many unexpected results.

What might lead to ...?

I think I missed an important part: we should try to set the locale to en_US at the beginning of the test. If it fails, skip.

frederikbosch:
@sagikazarmark I agree with that. But I do not get why your results with `nl_NL` are so different. That tells me that we have an unstable API.

sagikazarmark:
I didn't pay too much attention, maybe something else was wrong. It is PHP after all, stability is not one of the keywords. Also, intl is especially something vague.

barryvdh:
Hmm, I don't have a lot of experience here, but if results are unreliable, I'm not sure I'd trust this to handly my money stuff.

sagikazarmark:
I think we could preserve the "original" parsing logic in a separate parser, which was less smart, but more stable.

frederikbosch:
@barryvdh Do not get this wrong: `Money` is reliable. We are talking about an unmerged PR on parsing and formatting negative values with `intl` extension.

@sagikazarmark We should look into this more before drawing conclusions.

sagikazarmark:
Agree, I didn't want to give up on intl extension, but for simpler cases, the previous logic could be useful as well, like SimpleParser.

frederikbosch:
@sagikazarmark The PR already contains [the original parser](https://github.com/frederikbosch/money/blob/parser/src/StringToUnitsParser.php).

sagikazarmark:
:+1:

barryvdh:
Yes I understand Money is stable, I was just referring to intl parsing by itself ;)

sagikazarmark:
Okay, but I guess this way you are able to stick to the "default/previous" strategy or come up with a better, since we have an interface for that and no pressure on which to use.

frederikbosch:
@sagikazarmark Why don't we inject `NumberFormatter` into `IntlMoneyParser` and `IntlMoneyFormatter`?

This we place behaviour of the `intl` extension outside of `Money`. As such we only have to guarantee/test the conversion, and not the results of `NumberFormatter` itself like we are doing now.

sagikazarmark:
I remember I had some issues with NumberFormatter itself, and this injection approach, but I can't see any better solution. Although, it would be nice to look around and find some usage examples.

My biggest concern is the required configuration in of the NumberFormatter. It needs some basic config to work well (which is not necessarily configurable later).

frederikbosch:
Yes, if `Money` provides an intl parser and formatter, at least we should know how to use it! Hence, we should know how to configure it.

But, now the intl parser and formatter implementations are instantiating the `NumberFormatter` object. This limits the user in configuring the `NumberFormatter` how he or she wants it. For example, the current does not allow a developer to create a [pattern](http://php.net/manual/en/numberformatter.setpattern.php).

frederikbosch:
Okay. When constructing an `NumberFormatter` as below, the outcomes are different because the Travis is using a different default pattern.

```php
$formatter = new \NumberFormatter('en_US', \NumberFormatter::CURRENCY);
$config = [ .. ];
var_dump($config);
```

Travis:
```
array(9) {
  'locale_valid' =>
  string(5) "en_US"
  'locale_actual' =>
  string(2) "en"
  'pattern' =>
  string(23) "¤#,##0.00;(¤#,##0.00)"
  'attr_positive_prefix' =>
  string(1) "$"
  'attr_positive_suffix' =>
  string(0) ""
  'attr_negative_prefix' =>
  string(2) "($"
  'attr_negative_suffix' =>
  string(1) ")"
  'attr_padding_char' =>
  string(1) "*"
  'attr_currency_code' =>
  string(3) "USD"
}
```

My Ubuntu (14.04) machine:
```
array(9) {
  'locale_valid' =>
  string(5) "en_US"
  'locale_actual' =>
  string(2) "en"
  'pattern' =>
  string(10) "¤#,##0.00"
  'attr_positive_prefix' =>
  string(1) "$"
  'attr_positive_suffix' =>
  string(0) ""
  'attr_negative_prefix' =>
  string(2) "-$"
  'attr_negative_suffix' =>
  string(0) ""
  'attr_padding_char' =>
  string(1) "*"
  'attr_currency_code' =>
  string(3) "USD"
}
```

What is causing the default pattern to be different?

frederikbosch:
So, now I got a green light.

```php
$formatter = new \NumberFormatter('en_US', \NumberFormatter::CURRENCY);
$formatter->setPattern("¤#,##0.00;-¤#,##0.00");

$parser = new IntlMoneyParser($formatter);
$this->assertEquals($units, $parser->parse($string, 'USD')->getAmount());
```

Only setting a string with double quotes which include a pattern for positive and negative amounts via the setter (not via constructor) will give a stable result.

frederikbosch:
@sagikazarmark So we now have two parsers.
- The easy regex based one, called `StringToUnitsParser`. That one only parses a string into an amount. The developer is required to supply the currency.
- The `intl` one for the developer's advanced locale based parsing. This parser also parses a currency symbol, but can behave differently per system if not setting a pattern explicitly.

PR #149, opened 2016-01-18
fix formatter bug for negative amounts

frederikbosch:
Negative amounts were not tested and contains errors. This is now tested and gives correct results.

### DIFF
--- a/src/IntlMoneyFormatter.php
+++ b/src/IntlMoneyFormatter.php
@@ -10,30 +10,16 @@ namespace Money;
 final class IntlMoneyFormatter implements MoneyFormatter
 {
     /**
-     * @var string
+     * @var \NumberFormatter
      */
-    private $locale;
-    /**
-     * @var int
-     */
-    private $fractionDigits;
+    private $formatter;
 
     /**
-     * IntlMoneyFormatter constructor.
-     *
-     * @param int    $fractionDigits
-     * @param string $locale
+     * @param \NumberFormatter $formatter
      */
-    public function __construct($locale, $fractionDigits)
+    public function __construct(\NumberFormatter $formatter)
     {
-        if (extension_loaded('intl') === false) {
-            throw new \RuntimeException(
-                'Cannot initialize IntlMoneyFormatter because intl extension is missing'
-            );
-        }
-
-        $this->locale = $locale;
-        $this->fractionDigits = $fractionDigits;
+        $this->formatter = $formatter;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


frederikbosch:
It seems that both Travis failures (#148 and #149) are related to negative amounts. My machine uses a dash while travis use parenthesis. The weird thing is, this should not differ because we are using the same locale. So is my machine or Travis wrong?

frederikbosch:
Can you run the tests at your machine @sagikazarmark?

sagikazarmark:
Some resources I found:

http://ux.stackexchange.com/questions/1869/preferred-format-to-display-negative-currency-us-english/1875
http://stackoverflow.com/questions/31408633/numberformatterparsecurrency-gives-different-results-on-machines-with-same-php

I think formatting/parsing is not that easy after all.

For example I have this error on my machine:

```
1) Tests\Money\MoneyFormatterTest::testDifferentDigits
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'€ 6,135-'
+'€ -6,135'
```

I think it depends on some setting we haven't think about: machine locale maybe?

If so, I propose setting the locale specifically in each TestCase (or check if phpunit provides an easier way to do so, like env vars?), for example en_US (I think this is the only one installed on travis).
If this is the case, we probably can't control the behavior, which we have to document.

frederikbosch:
You are right, I think it also depends on locale, whether it is available or not. When it is not available, it falls back to the system's locale.

That [failing test](https://github.com/frederikbosch/money/blob/formatter_fixes/tests/MoneyFormatterTest.php#L39) of yours uses locale nl_NL. I would totally get that you do not have that one installed (not even 1% of the world's population is Dutch). What happens if you install that locale?

Suppose that works, then the questions is how to get locales installed on Travis? Or we should figure out what the system's locale of the Travis machine is.

sagikazarmark:
> en_US (I think this is the only one installed on travis)

sagikazarmark:
https://docs.travis-ci.com/user/common-build-problems/#System%3A-Required-language-pack-isn%E2%80%99t-installed

frederikbosch:
Oh sorry, I misread that. But then why is en_US on my machine different from en_US at Travis? I read those links you sent me, but the answer is not clear yet. The example of one those guys uses a pattern. We are not using a pattern.

frederikbosch:
@sagikazarmark I will give that a try.

sagikazarmark:
So the problem is that the intl extension falls back to something, if the given locale is not found, right?

frederikbosch:
@sagikazarmark That is my best guess. Did you only have one failing test? The one you posted here?

sagikazarmark:
Yup.

frederikbosch:
Ok, can you install the nl_NL locale and see what happens? I will try to get Travis to talk.

sagikazarmark:
I installed nl_NL locale, no change. Also, if I set the locale in PHP specifically to nl_NL.UTF-8, I get even more errors. :smile:

frederikbosch:
@sagikazarmark Did not find time to look into this. Will do quickly. Do not want to slow down release.

frederikbosch:
@sagikazarmark The `IntlMoneyFormatter` is now consistent with the `IntlMoneyParser`: `\NumberFormatter` is now injected via the constructor.

Maybe we should also include an easier formatter. I looked at `money_format` but that one expects a `float`. So we cannot use that one. Another method could be to use `sprintf`. We can also leave it at one formatter.

src/IntlMoneyFormatter.php
```diff
@@ -42,17 +28,26 @@ public function __construct($locale, $fractionDigits)
...
-            $subunits .= substr($valueBase, $valueLength - $this->fractionDigits);
+        if ($valueLength > $fractionDigits) {
+            $subunits = substr($valueBase, 0, $valueLength - $fractionDigits).'.';
+            $subunits .= substr($valueBase, $valueLength - $fractionDigits);
         } else {
-            $subunits = '0.'.str_pad('', $this->fractionDigits - $valueLength, '0').$valueBase;
+            $subunits = '0.'.str_pad('', $fractionDigits - $valueLength, '0').$valueBase;
         }
 
-        $formatter = new \NumberFormatter($this->locale, \NumberFormatter::CURRENCY);
```
sagikazarmark: By removing custom instantiation, what if the second parameter of the cunstructor is not that constant? Will this formatter work? If not, we should document it (even if this class's constructor IMO)

frederikbosch:
Good question, will look into it.

frederikbosch:
@sagikazarmark The formatter will still work if the second constructor parameter (style) is different. Look at my last commit to this PR where I test this.

When one chooses to set the style to decimal or percentage, the formatter will actually format it to decimals and percentages. If there also is a pattern set, then the pattern will be chosen over the style. I think this is desired behaviour. I rather wonder what happens with parsing.

sagikazarmark:
@frederikbosch thanks for checking. Actually I meant this for parsing, seems I failed to ask the question in the right issue.

frederikbosch:
@sagikazarmark Then you can merge this PR ;) I will add tests for the parser too in a new PR.

sagikazarmark:
Thanks @frederikbosch

PR #150, opened 2016-01-18
cs fixes in gmp calculator: snake case to camel case

frederikbosch:


sagikazarmark:
Thanks @frederikbosch

PR #152, opened 2016-01-22
fix floating point bug for number constructor

### LINKED ISSUE #151: Casting to (string) relies on PHP `precision` setting to be between 3 and 8 to work as expected
I noticed in the constructor for `Number` you do a `(string) $number`.

https://github.com/mathiasverraes/money/blob/nextrelease/src/Number.php#L25

We had [some problems doing this on OmniPay](https://github.com/thephpleague/omnipay-common/issues/69) when changing a float to a string, as the string it creates will depend on the accuracy that the current PHP system is set up.

To avoid some unexpected results on a system setting that is not under the control of a currency package, you can use `sprintf()` to force a specific accuracy each time, instead of leaving it for PHP to use whatever `precision` setting it happens to have.

frederikbosch:
Fixes #151.

### DIFF
--- a/src/Calculator/BcMathCalculator.php
+++ b/src/Calculator/BcMathCalculator.php
@@ -103,7 +103,7 @@ final class BcMathCalculator implements Calculator
      */
     public function round($number, $roundingMode)
     {
-        $number = new Number($number);
+        $number = new Number((string)$number);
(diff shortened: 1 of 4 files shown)


sagikazarmark:
Nice solution. :+1:

PR #154, opened 2016-02-01
test different style when parsing

frederikbosch:
Different style with pattern always works. Different style without pattern throws an exception. I am fine with that behaviour.


frederikbosch:
No it is different. `NumberFormatter` does not give expected returns when there is no pattern (like before). At my computer exceptions are thrown, apparently not on Travis.

PR #159, opened 2016-02-07
Check zero value

barryvdh:
A string with zero will now fail (see failed test). Convert to int at the end.


frederikbosch:
Solved by commit 766bdc8

frederikbosch:
Thanks @barryvdh for finding problem ad providing solution!

PR #160, opened 2016-02-07
Use alcohol iso4217 list

barryvdh:
This replaces the `umpirsky/currency-list` list with https://github.com/alcohol/iso4217
This is smaller and probably more fitted for this use-case.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     ],
     "require": {
         "php": ">=5.4",
-        "umpirsky/currency-list": "^1.0"
+        "alcohol/iso4217": "^3.1"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.5",
--- a/src/ISOCurrencies.php
+++ b/src/ISOCurrencies.php
@@ -2,6 +2,8 @@
 
 namespace Money;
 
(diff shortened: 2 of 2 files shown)


sagikazarmark:
Thanks for the PR. I am also aware of the alcohol library and I chose to use @umpirsky's implementation instead. Although it is bigger in size, I am more confident of the ISO code update process and it provides a plain list instead of an object. I would wait for @umpirsky's response to the other issue so that we have all the possibilities listed.

frederikbosch:
@barryvdh You can move this code to your own package in your own namespace (I guess it is for Omnipay) because you are using the lib by alcohol to implement the Currencies interface. This is probably the reason why we have chosen to make an interface: it helps people creating their own currency lists according to their own needs. Nobody is requiring you to use the implementation that is embedded in this package.

@sagikazarmark I agree that reliability is first priority in this case. And @barryvdh I think we all agree that having a huge package size is not what we want. So there will be a solution for it.

barryvdh:
No I don't actually need this code. I just wanted to provide an alternative solution to the current package, because of it's size. But moving it to suggest is even better ;)

PR #163, opened 2016-02-10
Also check tests/ folder

lucasmichot:
I suppose tests folder can also be ckecked ?

---

https://styleci.io/analyses/8wjvYZ can be auto-applied


sagikazarmark:
Thank you very much

PR #178, opened 2016-03-02
proposed revision to warn of breaking change

elstamey:
I thought it might be helpful to make the change I requested yesterday.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 3.0.0-beta - 2016-03-01
 
+### Breaking Change in 3.0
+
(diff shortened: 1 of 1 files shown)


frederikbosch:
Thanks @elstamey. This is really appreciated. We might want to add more to the list because there are more BC breaks, e.g. stringToUnits has been moved to a parser.

elstamey:
I have no problem with your pushing those changes to this branch or taking this work to another branch.

sagikazarmark:
Closing in favor of #180 

We will have that list pushed as a whole (once it is complete) instead of one by one.

Thanks @elstamey

PR #179, opened 2016-03-04
Added static constructor for converting numerics into money objects

pinkbigmacmedia:
Added static constructor for converting numeric values into money objects.

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -80,8 +80,9 @@ final class Money implements \JsonSerializable
     /**
      * Convert a given decimal/currency pair into a Money object.
      *
-     * @param float|integer $money
-     * @param string        $currency
+     * @param  float|integer $money
+     * @param  string        $currency
+     * @return Money
      */
     public static function fromDecimal($money, $currency)
     {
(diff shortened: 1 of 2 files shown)


frederikbosch:
@pinkbigmacmedia Why not using a parser for this? I think the `StringToUnitsParser` will help you converting a decimal.

```php
$parser = new StringToUnitsParser();
$money = $parser->parse((string) 999.99, 'EUR');
```

pinkbigmacmedia:
@frederikbosch Fair enough. I just found the one-liner as one way for creating money objects kind of handy.

Could we potentially have something like this on `Money\Money`:
```php
/**
  *
  * @param  string  $money
  * @param  string  $currency
  * @return Money
  */
public static function fromString($money, $currency)
{
    $parser = new StringToUnitsParser();
    return $parser->parse( (string) $money, $currency);
}
```

sagikazarmark:
I am :-1: on this. Although it might be handy, Money should preserve it's identity as a value object and should not be polluted with all the possible instantiation stuff. The reason why we implemented parsers is to allow devs creating money objects whatever way they want.

frederikbosch:
I have nothing to add to that.

PR #182, opened 2016-03-10
Suggest bc breaks

### LINKED ISSUE #180: BC Breaking changes
- Money::getAmount() returns a string instead of an int value
- stringToUnits has been moved to a parser

frederikbosch:
Fixes #180

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## 3.0.0-beta - 2016-03-01
 
-### Breaking Change in 3.0
-
-- Money::getAmount() returns a string instead of an int value  
-
 ### Added
 
 - Bitcoin parser and formatter
@@ -33,6 +29,9 @@
 
 ### Changed
 
+- **[BC break]** Money::getAmount() returns a string instead of an int value
+- **[BC break]** Moved stringToUnits to StringToUnitsParser parser
+- Announced deprecations are removed (Currency::getName, CurrencyPair::getRatio, Money::getUnits)
 - Library requires at least PHP 5.4
 - Library uses PSR-4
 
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Actually I think it would be better to keep the usual keepachangelog style AND:

- add the `[BC]` tag before each change which is BC breaking
- optionally make it bold
- also a big fat warning that most of the library is rewritten, so apart from the big breaks, there might be minor ones (refer to docs plz)

frederikbosch:
Good idea, will change it.

elstamey:
Thank you!  I'm so glad that users can more readily see breaking changes in the changelog.  We had to find the change to GetAmount's returning a string by reading the code because nothing was obvious in the changelog.

frederikbosch:
@sagikazarmark The PR is updated to your suggestions. My suggestion would be to repeat the BC breaks in the changelog when we release 3.0.0.

CHANGELOG.md
```diff
@@ -29,6 +29,9 @@
 
 ### Changed
 
+- **[BC break]** Money::getAmount() returns a string instead of an int value
+- **[BC break]** Moved stringToUnits to StringToUnitsParser parser
+- Announced deprecations are removed (Currency::getName, CurrencyPair::getRatio, Money::getUnits)
```
sagikazarmark: This shoulg go into the Removed section IMO

sagikazarmark:
> My suggestion would be to repeat the BC breaks in the changelog when we release 3.0.0.

Sounds like a sensible idea. :+1:

frederikbosch:
@sagikazarmark Updated. I qualified all removed features as BC breaks. Because any removed feature is breaking, right?

PR #186, opened 2016-03-16
bug fix: behaviour of negative amounts in all calculators

### LINKED ISSUE #185: Rounding of negative numbers is weird
I'm exposing my mathematical ignorance here, but could someone explain to me how the calculator arrives at the following?

```php
$currency = new Currency('GBP');
$money = new Money(158243, $currency);
$percent = -5;
$multiplier = ((100 / (100 + $percentage)) * $percentage) / 100; // -0.052631578947368
$amount = $money->multiply($multiplier);
var_dump($amount->getAmount()); // -8327
```

To my eyes, here's the maths I expect to happen here:

`1582.43 * -0.052631578947368 = -83.285789474`

Which, when rounded to two decimal places using PHP_ROUND_HALF_UP should be:

`-83.29`

Getting `-83.27` instead just doesn't seem right to me, whatever the rounding mode is. Can anyone explain?

frederikbosch:
Fixes #185

### DIFF
--- a/src/Number.php
+++ b/src/Number.php
@@ -54,12 +54,8 @@ final class Number
             return false;
         }
 
-        $decimalPositions = strlen($this->number) - ($this->decimalSeparatorPosition + 1);
+        $decimalPositions = strlen(rtrim($this->number, '0')) - ($this->decimalSeparatorPosition + 1);
         if ($decimalPositions === 1) {
-            return true;
-        }
-
-        if (substr($this->number, $this->decimalSeparatorPosition + 2) === str_pad('', $decimalPositions - 1, '0')) {
             return true;
         }
 
--- a/src/Number.php
+++ b/src/Number.php
@@ -148,6 +148,9 @@ final class Number
         return rtrim(substr($this->number, $this->decimalSeparatorPosition + 1), '0');
     }
 
+    /**
+     * @return string
+     */
     public function getIntegerRoundingMultiplier()
     {
         if ($this->isNegative()) {
--- a/src/Number.php
+++ b/src/Number.php
@@ -147,4 +147,13 @@ final class Number
 
         return rtrim(substr($this->number, $this->decimalSeparatorPosition + 1), '0');
     }
+
+    public function getIntegerRoundingMultiplier()
+    {
+        if ($this->isNegative()) {
+            return '-1';
+        }
+
+        return '1';
+    }
 }
--- a/src/Number.php
+++ b/src/Number.php
@@ -115,4 +115,36 @@ final class Number
 
         return new self(sprintf('%.8g', $floatingPoint));
     }
+
+    /**
+     * @return bool
+     */
+    public function isNegative()
+    {
+        return $this->number[0] === '-';
+    }
+
+    /**
+     * @return string
+     */
+    public function getIntegerPart()
+    {
+        if ($this->decimalSeparatorPosition === false) {
+            return $this->number;
+        }
+
+        return substr($this->number, 0, $this->decimalSeparatorPosition);
+    }
+
+    /**
+     * @return string
+     */
+    public function getFractionalPart()
+    {
+        if ($this->decimalSeparatorPosition === false) {
+            return '';
+        }
+
+        return substr($this->number, $this->decimalSeparatorPosition + 1);
+    }
 }
--- a/src/Number.php
+++ b/src/Number.php
@@ -145,12 +145,6 @@ final class Number
             return '';
         }
 
-        $fractionalPart = substr($this->number, $this->decimalSeparatorPosition + 1);
-
-        if (str_pad('', strlen($fractionalPart), '0') === $fractionalPart) {
-            return '';
-        }
-
-        return $fractionalPart;
+        return rtrim(substr($this->number, $this->decimalSeparatorPosition + 1), '0');
     }
 }
--- a/src/Number.php
+++ b/src/Number.php
@@ -145,6 +145,12 @@ final class Number
             return '';
         }
 
-        return substr($this->number, $this->decimalSeparatorPosition + 1);
+        $fractionalPart = substr($this->number, $this->decimalSeparatorPosition + 1);
+
+        if (str_pad('', strlen($fractionalPart), '0') === $fractionalPart) {
+            return '';
+        }
+
+        return $fractionalPart;
     }
 }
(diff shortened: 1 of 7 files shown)


src/Number.php
```diff
@@ -115,4 +115,36 @@ public static function fromFloat($floatingPoint)
...
+        return substr($this->number, 0, $this->decimalSeparatorPosition);
+    }
+
+    /**
+     * @return string
+     */
+    public function getFractionalPart()
+    {
+        if ($this->decimalSeparatorPosition === false) {
+            return '';
```
frederikbosch: Or should I throw an exception here? My feeling is not: a number can have an empty fractional part.

tests/Calculator/CalculatorTest.php
```diff
@@ -84,6 +84,10 @@ public function testRound()
         $this->assertEquals('2', $calculator->round(2, Money::ROUND_HALF_EVEN));
         $this->assertEquals('2', $calculator->round(2, Money::ROUND_HALF_ODD));
         $this->assertEquals('12', $calculator->round('12.50', Money::ROUND_HALF_DOWN));
+        $this->assertEquals('-2', $calculator->round(-1.5, Money::ROUND_HALF_UP));
+        $this->assertEquals('-8329', $calculator->round(-8328.578947368, Money::ROUND_HALF_UP));
+        $this->assertEquals('-8329', $calculator->round(-8328.5, Money::ROUND_HALF_UP));
```
pamil: In fact, rounding half up should result in `-8328.5` moved towards positive infinity which results in `-8328`. That's the way PHP behaves too, but with introducing our custom constraints, we can fix that issue and provide correct implementations (I did something like that some time ago in [Rounder](https://github.com/pamil/Rounder)).

src/Number.php
```diff
@@ -115,4 +115,36 @@ public static function fromFloat($floatingPoint)
...
+     */
+    public function isNegative()
+    {
+        return $this->number[0] === '-';
+    }
+
+    /**
+     * @return string
+     */
+    public function getIntegerPart()
```
frederikbosch: Regarding the terminology: I am basing in on [this thread on StackExchange](http://math.stackexchange.com/questions/3448/rules-for-rounding-positive-and-negative-numbers).

src/Number.php
pamil: What if I create a `Number('42.00')`, will it return `00`? IMHO it should return the same value for both `42` and `42.00`.

tests/Calculator/CalculatorTest.php
frederikbosch: @pamil Yes, I correct the behaviour as such that it is conform the `round` function in PHP. Why is that behaviour wrong?

src/Number.php
frederikbosch: Just fixed this.

tests/Calculator/CalculatorTest.php
pamil: @frederikbosch the naming is misleading, rounding half up means towards the positive infinity, rounding half down - towards negative infinity. However, PHP (and Java) implements the first one as away from zero and the second as towards zero. 

https://en.wikipedia.org/wiki/Rounding#Round_half_up may be helpful

src/Number.php
```diff
@@ -115,4 +115,42 @@ public static function fromFloat($floatingPoint)
...
+     */
+    public function getFractionalPart()
+    {
+        if ($this->decimalSeparatorPosition === false) {
+            return '';
+        }
+
+        $fractionalPart = substr($this->number, $this->decimalSeparatorPosition + 1);
+
+        if (str_pad('', strlen($fractionalPart), '0') === $fractionalPart) {
```
pamil: We can just do `return rtrim($fractionalPart, '0')` here, so it will be shorter and will also make the number `42.500` return the same fractional part as `42.5`.

tests/Calculator/CalculatorTest.php
frederikbosch: @pamil While I totally agree that it is misleading, I am not sure that this package should move away from that (misleading) behaviour. The PHP behaviour is the one we had in version 1 of this package. It is very well explainable that we follow the PHP behaviour by default. We could introduce new rounding modes: `Money::ROUND_HALF_TOWARDS_POSITIVE_INFINITY` and `Money::ROUND_HALF_TOWARDS_NEGATIVE_INFINITY`. We could even change the default behaviour.

@sagikazarmark What do you think?

src/Number.php
frederikbosch: Done. Good one!

tests/Calculator/CalculatorTest.php
sagikazarmark: Well, I think the logical way of rounding is the way how @pamil described rounding, but I never realized the difference here. For consistency, I think it would be better to maintain the current behavior, as the change would cause as much confusion as the misleading itself. Keyword: documentation.

As compromise, we could have ROUND_HALF_*_NORMAL, or something like you suggested @frederikbosch, but a little bit more concise.

src/Number.php
```diff
@@ -115,4 +115,36 @@ public static function fromFloat($floatingPoint)
 
         return new self(sprintf('%.8g', $floatingPoint));
     }
+
+    /**
+     * @return bool
+     */
+    public function isNegative()
+    {
+        return $this->number[0] === '-';
```
sagikazarmark: Edge case: what happens if someone enters `-0` here? Will it have any effect at all?

tests/Calculator/CalculatorTest.php
frederikbosch: Fully agree: documentation is the keyword. And honestly, I also did not realize the difference. I think many people don't.

My vote would not go to `NORMAL`. I mean, what is normal? But I understand that my own suggestion is also not great. What about these?

`Money::ROUND_HALF_POSITIVE_INFINITY`
`Money::ROUND_HALF_NEGATIVE_INFINITY`
`Money::ROUND_HALF_UP`
`Money::ROUND_HALF_DOWN`

So I removed the word towards from the constants. The major question is: should we change the default behaviour? I can understand that the majority rather sees a negative number being rounded to positive infinity. You can even wonder when you need the default PHP behaviour at all.

frederikbosch: Misread your suggestion on keeping the current behaviour. Agree. Let's not change the API. But shall we add the constants? Do we need to add a possibility to change the default behaviour (e.g. `Money::changeDefaultRoundingMode`).

src/Number.php
frederikbosch: Good question, let's test that.

tests/Calculator/CalculatorTest.php
sagikazarmark: Ah, no. No more responsibilities. The reason why people CAN describe rounding is to avoid problems originating from default behaviors.

All in all: I don't think there is a right solution for this problem. If we change, there will be problems. But this way it is misleading. Again: docs FTW, but apart from that, we can't really make everyone happy.

frederikbosch: Ok good reasoning. We will leave the current behaviour, add two constants and have *no* extra responsibility for change default behaviour.

src/Number.php
frederikbosch: This is tested now, the behaviour was correct.

Issue #189, opened 2016-03-18
Conversion back and forth


fedeisas:
Hi! I'm having some issues converting money back and forth. Is that operation supposed to be idempotent?

Example:

```php
$ratio = 14.834670;
$inverseRatio = round(floatval(1/$ratio), 6);

$USDARSpair = \Money\CurrencyPair::createFromISO('USD/ARS ' . $ratio);
$ARSUSDpair = \Money\CurrencyPair::createFromISO('ARS/USD ' . $inverseRatio);

$numberFormatter = new \NumberFormatter('en_US', \NumberFormatter::CURRENCY);
$formatter = new \Money\Formatter\IntlMoneyFormatter($numberFormatter);
$money = new \Money\Money(4250, new \Money\Currency('ARS'));

var_dump($formatter->format($money));
var_dump($formatter->format($ARSUSDpair->convert($money)));
var_dump($formatter->format($USDARSpair->convert($ARSUSDpair->convert($money))));

string 'ARS42.50' (length=8)
string '$2.86' (length=5)
string 'ARS42.43' (length=8)
```

I know maybe the problem is in how I'm calculating the `$inverseRatio`.

Thanks!

fedeisas:
I think it's because when you can `Money::convert` it rounds the value to cents. Maybe rounding should be done upon formatting?

sagikazarmark:
``` bash
$ php -r 'echo round(floatval(1/14.834670), 6) * 14.834670;'
```

This should end up in exactly one and it does not. I hardly think that this is a CurrencyPair issue, but the result of the rounding. In order to preserve consistency (and hence the immutability) we cannot do rounding upon a specific action. After each one, the rounding is done. But as you can see, this is the mistake of "your"/php's rounding, not Money's one.

frederikbosch:
@sagikazarmark is right.

```
4250×0.067409656 = 286.491038
4243×0.067409656 = 286.019170408
```

Both result in an amount of `286` when rounding. You will need `PreciseMoney`, which will be realized in `3.1.0`, in order to have a money value object with decimals.

However, you could change the value object by changing the fraction of the amount to 5.

```php
$ratio = '14.834670';
$inverseRatio = '0.067409656';

$pair = CurrencyPair::createFromISO('USD/ARS ' . $ratio);
$inversePair = CurrencyPair::createFromISO('ARS/USD ' . $inverseRatio);

$moneyUsd = new Money(286000, new Currency('USD'));
[...]

PR #191, opened 2016-03-22
Remove float from currency pair

frederikbosch:
@sagikazarmark While testing issue #189 I found that `CurrencyPair` is using `(float)`. I think it would be better if we get rid of it, and use `(string)`. This also has consequences for the json representation, because that will be using quotation marks now. What do you think?

### DIFF
--- a/src/CurrencyPair.php
+++ b/src/CurrencyPair.php
@@ -45,7 +45,7 @@ final class CurrencyPair implements \JsonSerializable
 
         $this->counterCurrency = $counterCurrency;
         $this->baseCurrency = $baseCurrency;
-        $this->conversionRatio = (float) $conversionRatio;
+        $this->conversionRatio = (string) $conversionRatio;
     }
 
     /**
@@ -73,7 +73,7 @@ final class CurrencyPair implements \JsonSerializable
             );
         }
 
-        return new static(new Currency($matches[1]), new Currency($matches[2]), $matches[3]);
+        return new self(new Currency($matches[1]), new Currency($matches[2]), $matches[3]);
     }
 
     /**
--- a/tests/CurrencyPairTest.php
+++ b/tests/CurrencyPairTest.php
@@ -78,6 +78,21 @@ final class CurrencyPairTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals(Money::EUR(100), $eur);
     }
 
+    public function testConvertsArsToUsdAndBack()
+    {
+        $ratio = '14.834670';
+        $inverseRatio = '0.067409656';
+
+        $pair = CurrencyPair::createFromISO('USD/ARS '.$ratio);
+        $inversePair = CurrencyPair::createFromISO('ARS/USD '.$inverseRatio);
+
+        $moneyUsd = new Money(286, new Currency('USD'));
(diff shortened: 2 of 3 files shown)


sagikazarmark:
Not sure. We use string for Money as a compromise. It is very unlikely that we run into issues with the float representation.

What's the advantage?

frederikbosch:
@sagikazarmark It is not that I want to use `string`. I think we should we avoid `float` whenever we can. Maybe that a simple conversion to float via `(float)` is harmless, but why do we want a floating point number in our class?

sagikazarmark:
The reason behind using integer is to avoid rounding issues. The reason behind using string for representing numbers is to allow using Calculators handling big integers. Other words: the reason why we don't use float is rounding.

However in this case we usually want to have a decimal value. Storing it as string, just to not store it as float does not come with any advantage that I can see. On the other hand, it is inconvenient, just like in case of Money, but there we have bigger justification for doing so.

frederikbosch:
@sagikazarmark You have a point there. I cannot see the advantage too. We should be aware though.

frederikbosch:
Do we really want to have float in our codebase? Please read the [RFC on higher float precision](https://wiki.php.net/rfc/precise_float_value). Especially the current behaviour of some values. We should avoid float, at almost any cost.

sagikazarmark:
What's the alternative to represent non-integer values (apart from string)? Keep in mind, that casting is almost as bad as rounding issues that might occur. As an alternative, we could do this on the JSON level to preserve higher precision: cast to string when serialization happens, do the reverse when deserialization. This is a one time operation and does not mess up mathematical ones with string representation. See my reasoning above why I think string is actually worse than float.

I see the precision problem, but I highly doubt that it might cause any problems (it didn't in the past). However, I don't think we should leave it, but I don't think string is a solution. Until we find one, we should have a separate issue and mention it in the docs.

frederikbosch:
@sagikazarmark The problem is that we are [casting to float](https://github.com/moneyphp/money/blob/master/src/CurrencyPair.php#L48). We are not securing what is going in, only that the value is numeric, not the type. Still agree though that it is highly unlikely that it causes issues.

sagikazarmark:
But it is an internal, one-time operation. Also, we do that to make sure the input value is float. Ideally nothing happens, as the input should already be float. Internally afterwards we use float as the representation.

sagikazarmark:
@frederikbosch can we consider this discussed and close? For now I am happy with the current solution. If we ever experience any issues regarding this, we can always reconsider this feature, although either as a bugfix or rather with a major version dump.

frederikbosch:
:+1:

frederikbosch:
@sagikazarmark I still think we should remove any reference to float in this library. Take what I had today.

```
php -r "echo (new NumberFormatter('nl_NL', NumberFormatter::DECIMAL))->parse('64,46') * 100;" // result: double(6446)
php -r "echo intval((new NumberFormatter('nl_NL', NumberFormatter::DECIMAL))->parse('64,46') * 100);" // result: int(6445)
```

If I was not convinced enough, now I am really convinced: working with float/double is really harmful. The PR I put forward was not even strict enough. We should only accepts string and throw an exception on float/double. Let the user do the math (conversion to string) outside this library. Float inside `Money` must be avoided at any cost.

PR #193, opened 2016-04-05
Add BYN currency code

mongosoft:
In July 2016 a new Belarusian ruble will be introduced, at a rate of 1 new ruble = 10,000 old rubles.  https://en.wikipedia.org/wiki/Belarusian_ruble

Please, update currency configuration file. Thanks for creating this extension!


sagikazarmark:
Thanks for the PR. We are not maintaining the currency list ourselves, but we use https://github.com/umpirsky/currency-list/ as a source. I contacted @umpirsky about the possibilities. If we can't make this currency in that package I will temporarily merge this one and add a note somewhere not to overwrite it until it makes it to the list.

Are you fine with that?

mongosoft:
@sagikazarmark Yes. That would be cool!

sagikazarmark:
Actually, I am thinking about adding a currency.local.php where these kind of corrections/additions could be done, even if the data source does not support it (yet). It could be merged with the base currency list. The downside is the extra IO, but opcache should mitigate it. Also, it is a relatively small file (It think mostly returning an empty array).

@frederikbosch WDYT?

@mongosoft do you think you can modify the PR accordingly?

mongosoft:
@sagikazarmark Sorry, I think I can not do it correctly and beautifully.

sagikazarmark:
Np, I will take a look at it at the weekend then.

frederikbosch:
@sagikazarmark I think we should not merge this, nor we should add the possibility for currency.local.php. I have the following reasoning. The list of currencies is only used inside `ISOCurrencies`, which implements `Currencies`. That means someone like @mongosoft can easily create an own implementation of Currencies. In this case using a decorator perfectly solves the issue.

```php
<?php
namespace MyNs;

use Money\Currencies;
use Money\ISOCurrencies;

class BynSupportedCurrencyList implements Currencies {

  private $delegatedList;

  public function __construct(Currencies $list) {
    $this->delegatedList = $list;
  }

  public function contains(Currency $currency) {
    if ($currency->getCode() === 'BYR') {
      return true;
    }
    return $this->delegatedList->contains($currency);
  }
}

$currencies = new BynSupportedCurrencyList(new ISOCurrencies));
```

Or am I missing something here?

sagikazarmark:
Good point @frederikbosch, this is actually a good solution. What do you think about it @mongosoft ?

This is a good solution, because it frees us from covering the edge cases.

frederikbosch:
@sagikazarmark We should document, and add it to a FAQ which we can include in our README.

mongosoft:
@frederikbosch @sagikazarmark I agree that this is the best solution at the moment. I like it. That will be enough to solve my problem.

frederikbosch:
Great, good luck!

PR #194, opened 2016-04-07
More descriptive exception message

jongotlin:


sagikazarmark:
Nice addition @jongotlin, thanks.

It seems to me a common practice that in such exceptions when the type is object, the FQCN is given as type instead of "object". WDYT?

jongotlin:
I had to look in the manual when not using annotation and the example uses only the class name. But if you want me to change I'm happy to do so.
https://phpunit.de/manual/4.8/en/phpunit-book.html#writing-tests-for-phpunit.exceptions.examples.ExceptionTest3.php

sagikazarmark:
I think you misunderstood me. I meant [this scenario](https://github.com/moneyphp/money/pull/194/files#diff-a3b23eb259814cf0992e941cd03fd471R132) should return `stdClass` as "type", not `object`.

jongotlin:
Ah. Isn't that the case when you're using `get_class`? `gettype` doesn't care about what object it is.

jongotlin:
Or do mean you want me to check if its an object, return its fqcn?

sagikazarmark:
> Or do mean you want me to check if its an object, return its fqcn?

Exactly.

jongotlin:
Got it :-) 
Even better

src/Money.php
```diff
@@ -244,14 +244,17 @@ public function subtract(Money $subtrahend)
     /**
      * Asserts that the operand is integer or float.
      *
-     * @param float|int|string $operand
```
frederikbosch: I think an operand of '1.5' should be allowed.

frederikbosch:
Looks good, one remark.

src/Money.php
jongotlin: You're right. My misstake.

frederikbosch:
Thanks @jongotlin

PR #198, opened 2016-04-19
Assign default value to variable passed by reference

sagikazarmark:
@frederikbosch can you take a look?

### DIFF
--- a/src/Parser/IntlMoneyParser.php
+++ b/src/Parser/IntlMoneyParser.php
@@ -32,6 +32,7 @@ final class IntlMoneyParser implements MoneyParser
      */
     public function parse($money, $forceCurrency = null)
     {
+        $currency = null;
         $decimal = $this->formatter->parseCurrency($money, $currency);
 
         if ($decimal === false) {


frederikbosch:
@sagikazarmark @Nikolai-Senkevich Good catch. Can be merged after CI is OK.

src/Parser/IntlMoneyParser.php
```diff
@@ -32,6 +32,7 @@ public function __construct(\NumberFormatter $formatter)
      */
     public function parse($money, $forceCurrency = null)
     {
+        $currency = null;
         $decimal = $this->formatter->parseCurrency($money, $currency);
```
pamil: Why not `$this->formatter->parse($money)` instead?

sagikazarmark: Because formatter is `NumberFormatter`: http://php.net/manual/en/numberformatter.parsecurrency.php

frederikbosch: Because `$this->formatter` is an `IntlFormatter`.

pamil: And it also has [`NumberFormatter::parse`](http://php.net/manual/en/numberformatter.parse.php).

pamil: Since we pass null as currency, it seems like a better choice.

sagikazarmark: But it does not give us the currency

pamil: Oh, I see, standard PHP library is so weird.

PR #200, opened 2016-04-19
Bitcoin namespace and currencies

### LINKED ISSUE #197: Bitcoin supported parser/formatter
Do we need the supported word? Simply BitcoinMoneyParser/Formatter?

I don't think we are forced to put every nature/capability of the class in the name.

frederikbosch:
Move bitcoin to own namespace, rename classes, add currencies implementation. Fixed #197.

### DIFF
--- a/src/Currencies/BitcoinCurrencies.php
+++ b/src/Currencies/BitcoinCurrencies.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Money\Currencies;
+
+use Money\Currencies;
+use Money\Currency;
+
+/**
+ * @author Frederik Bosch <f.bosch@genkgo.nl>
+ */
+final class BitcoinCurrencies implements Currencies
+{
+    const CODE = 'XBT';
+    const SYMBOL = "\0xC9\0x83";
+
+    /**
+     * {@inheritdoc}
+     */
+    public function contains(Currency $currency)
+    {
+        return self::CODE === $currency->getCode();
+    }
+}
--- a/src/Formatter/BitcoinMoneyFormatter.php
+++ b/src/Formatter/BitcoinMoneyFormatter.php
@@ -2,19 +2,17 @@
 
 namespace Money\Formatter;
 
+use Money\Currencies\BitcoinCurrencies;
 use Money\Money;
 use Money\MoneyFormatter;
-use Money\Parser\BitcoinSupportedMoneyParser;
 
 /**
  * Formats Money to Bitcoin currency.
  *
  * @author Frederik Bosch <f.bosch@genkgo.nl>
  */
-final class BitcoinSupportedMoneyFormatter implements MoneyFormatter
+final class BitcoinMoneyFormatter implements MoneyFormatter
 {
-    const CODE = 'XBT';
-
     /**
      * @var MoneyFormatter
      */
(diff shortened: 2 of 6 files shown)


frederikbosch:
@sagikazarmark Can you have a look? I realized we were missing the Currencies implementation for Bitcoin so I added that one too. With a third Bitcoin class I decided to move it to a separate namespace. This can be argued.

sagikazarmark:
I will take a look at it.

First thought: I would rather create an aggregate currency repository instead of using decorator here.

frederikbosch:
@sagikazarmark The Currencies implementation for Bitcoin does not withhold us from creating an aggregate repository. Creating a repo is a good idea. It could be added in version 3.1. In this cade a decorator is the quickest way to complete support for bitcoin. And it does not harm anyone.

sagikazarmark:
> And it does not harm anyone.

No, it does not, but is not-so-good design IMO. It would encourage people to create decorators like this.

Decorators perform well when a specific behaviour needs to be added to very well defined interfaces, which is not really true for a repository.

I was thinking about something like this:

``` php
final class AggregateCurrencies implements Currencies
{
    private $currencies;

    public function __construct($currencies)
    {
        $this->currencies = $currency;
    }

    public function contains(Currency $currency)
    {
            if ($currencies->contains($currency)) {
                return true;
            }
        }

        return false;
    }
}
```

This way bitcoin currency repository would only be responsible for bitcoin.

It is also easier to configure in DI (while configuring long decorator-chains can be hard or confusing)

frederikbosch:
@sagikazarmatk Agreed. Let's do the aggregate repo. Maybe close this one then, and reissue this PR with only formatter and parser classes renamed.

What do you think BitcoinMoneyFormatter or BitcoinFormatter. My vote goes to the latter but wonder what you think.

sagikazarmark:
> Maybe close this one then

Not necessary to close. Mark it [WIP] and then rebase when aggregate is added separately.

AFAIK it was your argument to call them MoneyFormatters, because we might want to add other formatters?

What's important for me is to have consistent naming and structure. Here are the options I could imagine:

- Have them in Bitcoin namespace and call them simply MoneyFormatter and MoneyParser.
- Have them in their respective namespace (Formatter, Parser) and call them BitcoinMoneyFormatter and BitcoinMoneyParser
- Call them BitcoinFormatter/Parser, but we should have similar naming in case of other parsers/formatters? (IntlFormatter?)

(BTW consistency and structure: should we have a Currencies namespace? If we put BitcoinCurrencies in Currencies namespace with ISO it would make sense IMO)

frederikbosch:
Ok. I'll update this PR next week.

- No more Bitcoin namespace.
- Class names will be BitcoinMoneyFormatter and BitcoinMoneyParser in the namespaces Formatter and Parser.
- BitcoinCurrencies will be dropped.
- Currencies namespace will be added and will contain ISOCurrencies and AggregateCurrencies.

Ok?

sagikazarmark:
> BitcoinCurrencies will be dropped

I didn't say `BitcoinCurrencies` should be dropped :wink: 

It should be there, but only contain bitcoin. Also, maybe constants from Formatter and Parser might be moved here as the source of all truth.

frederikbosch:
Aight, good ideas!

sagikazarmark:
From my POV this is ready to be merged.

waahhhh:
since June 2017 the symbol has changed. ₿
Source: https://en.bitcoin.it/wiki/Bitcoin_symbol
See: https://en.wiktionary.org/wiki/%E2%82%BF

frederikbosch:
Could you create a PR?

PR #210, opened 2016-04-22
Migrate tests to PHPSpec

sagikazarmark:


sagikazarmark:
@frederikbosch What do you think about this?

frederikbosch:
@sagikazarmark Never used PHP Spec. I have nothing against this because I do not know what we gain. Do you want to replace PHPUnit?

sagikazarmark:
Well, there are some PHPUnit tests remained, mostly the ones with functional nature (testing NumberFormatter and the like). I usually prefer spec tests because it is much more straigthforward about what we input and what we expect, also the mocking is much more easier than PHPUnit (I replaced many tests with using mocks instead of actual code). I am pretty sure it will take 5 seconds for you to understand.

I think you should probably look through some of the specs and if you think you can get familiar with it, then let's go with it.

frederikbosch:
@sagikazarmark I read them, easy read. Please merge. I will adapt.

PR #214, opened 2016-04-25
Clarify why strings are used for the amount in the README.

frederikbosch:
Clarify why strings are used for the amount.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,9 @@ PHP 5.5+ library to make working with money safer, easier, and fun!
 > "If I had a dime for every time I've seen someone use FLOAT to store currency, I'd have $999.997634" -- [Bill Karwin](https://twitter.com/billkarwin/status/347561901460447232)
 
 In short: You shouldn't represent monetary values by a float. Wherever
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,7 +8,9 @@ PHP 5.5+ library to make working with money safer, easier, and fun!
 > "If I had a dime for every time I've seen someone use FLOAT to store currency, I'd have $999.997634" -- [Bill Karwin](https://twitter.com/billkarwin/status/347561901460447232)
 
 In short: You shouldn't represent monetary values by a float. Wherever
-you need to represent money, use this Money value object.
+you need to represent money, use this Money value object. Since version
+3.0 this library uses [strings internally](https://github.com/moneyphp/money/pull/136)
+in order to support unlimited integers.
```
sagikazarmark: Not sure if it strongly belongs to the readme. This should be emphasized somewhere in the docs IMO.

frederikbosch: Disagree. I think we should put it upfront. Clear for anybody. How many times do people read docs? It does not weigh up to the times they read a readme.

PR #219, opened 2016-04-25
Base exception, fixes #207

### LINKED ISSUE #207: Add base exception interface

frederikbosch:


src/Exception/Exception.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Money\Exception;
+
+/**
+ * Base Exception.
+ *
+ * @author Frederik Bosch <f.bosch@genkgo.nl>
+ */
+
+abstract class Exception extends \Exception
```
sagikazarmark: I would rather have this as an interface.

sagikazarmark: That way we can use SPL exceptions as well and extend whichever we want. For example UnresolvableCurrencyPairException should be invalid argument.

frederikbosch:
Why as interface? You mean like a `marker` interface? What is the difference with an abstract class in thise case?

sagikazarmark:
That way we can use SPL exceptions as well and extend whichever we want. For example UnresolvableCurrencyPairException should be invalid argument:

`UnresolvableCurrencyPairException extends \InvalidArgumentException implements Exception`

sagikazarmark:
:+1:

frederikbosch:
@sagikazarmark Good idea. You do not see that many marker interfaces in PHP.

sagikazarmark:
https://github.com/php-http/httplug/blob/master/src/Exception.php
https://github.com/lstrojny/fxmlrpc/blob/master/src/fXmlRpc/Exception/ExceptionInterface.php
https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/Exception/HttpCacheExceptionInterface.php

There are a few :wink:

sagikazarmark:
Can you squash commits?

frederikbosch:
Merge it in?

PR #220, opened 2016-04-25
Use autoloading for "spec" namespace

pamil:


sagikazarmark:
To be honest, I used require statements on purpose. While I also used to do this, some recent changes made this impossible to work. However, looking at the test results, it seems to work again. I am going to look at phpspec issues to see if this issue has been official fixed.

Thanks for the PR.

sagikazarmark:
Here is the actual issue: https://github.com/phpspec/phpspec/issues/825

sagikazarmark:
As @ciaranmcnulty stated [here](https://github.com/phpspec/phpspec/issues/825#issuecomment-216476357), this will only work as long as there are no typehints in the traits.

@pamil can you please place a comment in the traits that currently PHPSpec does not support typehint mocks in traits (and a link to the issue above)?

pamil:
@sagikazarmark totally forgot about this PR, here it comes :)

sagikazarmark:
Thanks @pamil

ciaranmcnulty:
Actually, you can do `require PhpSpec\Loader\StreamWrapper::wrapPath('Path/To/Foo.php')` and it'll work, but it's not really supported behaviour

sagikazarmark:
Well, if it's not REALLY supported, then it is not supported and should not be used IMO. :smile: 

Thanks for the heads up anyway. For now, it doesn't seem to be a big issue not to use typehints in our traits.

PR #223, opened 2016-05-03
Add extra optional argument subunits to IntlMoneyFormatter.

frederikbosch:
With the extra argument the user of formatter can now also control the number of subunits for a currency. This is fundamentally different than the amount fraction digits to be displayed (presentation vs business rule). It is optional because in the majority of the cases the subunits will equal the number of fraction digits.

### DIFF
--- a/src/Formatter/IntlMoneyFormatter.php
+++ b/src/Formatter/IntlMoneyFormatter.php
@@ -16,13 +16,23 @@ final class IntlMoneyFormatter implements MoneyFormatter
      * @var \NumberFormatter
      */
     private $formatter;
+    /**
+     * @var int
+     */
+    private $subunits;
 
     /**
      * @param \NumberFormatter $formatter
+     * @param int              $subunits
      */
-    public function __construct(\NumberFormatter $formatter)
+    public function __construct(\NumberFormatter $formatter, $subunits = 0)
     {
+        if (is_int($subunits) === false) {
+            throw new \InvalidArgumentException('Subunits must be an integer');
+        }
+
         $this->formatter = $formatter;
+        $this->subunits = $subunits;
     }
 
     /**
@@ -38,20 +48,19 @@ final class IntlMoneyFormatter implements MoneyFormatter
             $valueBase = substr($valueBase, 1);
         }
 
-        $fractionDigits = $this->formatter->getAttribute(\NumberFormatter::FRACTION_DIGITS);
         $valueLength = strlen($valueBase);
 
-        if ($valueLength > $fractionDigits) {
-            $subunits = substr($valueBase, 0, $valueLength - $fractionDigits).'.';
-            $subunits .= substr($valueBase, $valueLength - $fractionDigits);
+        if ($valueLength > $this->subunits) {
+            $number = substr($valueBase, 0, $valueLength - $this->subunits).'.';
+            $number .= substr($valueBase, $valueLength - $this->subunits);
         } else {
-            $subunits = '0.'.str_pad('', $fractionDigits - $valueLength, '0').$valueBase;
+            $number = '0.'.str_pad('', $this->subunits - $valueLength, '0').$valueBase;
         }
 
         if ($negative === true) {
-            $subunits = '-'.$subunits;
+            $number = '-'.$number;
         }
 
-        return $this->formatter->formatCurrency($subunits, $money->getCurrency()->getCode());
+        return $this->formatter->formatCurrency($number, $money->getCurrency()->getCode());
     }
 }
--- a/tests/Formatter/IntlMoneyFormatterTest.php
+++ b/tests/Formatter/IntlMoneyFormatterTest.php
@@ -11,8 +11,16 @@ final class IntlMoneyFormatterTest extends \PHPUnit_Framework_TestCase
     /**
      * @dataProvider numberFormatterExamples
      */
-    public function testNumberFormatter($amount, $currency, $result, $locale, $mode, $hasPattern, $fractionDigits)
-    {
+    public function testNumberFormatter(
+        $amount,
+        $currency,
+        $result,
+        $locale,
+        $mode,
+        $hasPattern,
+        $fractionDigits,
+        $subunits
+    ) {
         $money = new Money($amount, new Currency($currency));
 
         $numberFormatter = new \NumberFormatter($locale, $mode);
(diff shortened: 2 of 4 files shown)


src/Formatter/IntlMoneyFormatter.php
```diff
@@ -16,13 +16,19 @@
      * @var \NumberFormatter
      */
     private $formatter;
+    /**
+     * @var int|null
```
sagikazarmark: null is casted to 0 if used in arithmetical expressions. Shouldn't 0 be the default value?

tests/Formatter/IntlMoneyFormatterTest.php
```diff
@@ -50,4 +50,19 @@ public static function numberFormatterExamples()
             [5, 'EUR', '500%', 'en_US', \NumberFormatter::PERCENT, false, 0],
         ];
     }
+
+    public function testSubunits()
```
sagikazarmark: I would prefer adding this to spec tests.

src/Formatter/IntlMoneyFormatter.php
frederikbosch: It should be either `null` (optional) or required in the constructor. We could also make a named constructor for this. So then you have two options.

```php
$formatter = new IntlMoneyFormatter(new \NumberFormatter());
// or
$formatter = IntlMoneyFormatter::withSubunits(new \NumberFormatter(), 2);
```

The advantage is you can eliminate `null` from the constructor, which is a good thing because you want to avoid null as much as you can.

tests/Formatter/IntlMoneyFormatterTest.php
frederikbosch: Will do.

src/Formatter/IntlMoneyFormatter.php
sagikazarmark: `public function __construct(\NumberFormatter $formatter, $subunits = 0)` ?

Subunit is still optional

frederikbosch: It is good to have available, but why should it be required? There was nothing wrong with the old behaviour. Only it does not match everyone's needs. This fixes that.

sagikazarmark: > It is good to have available, but why should it be required?

I don't understand. What is required?

frederikbosch: @sagikazarmark The case: "we are not able to control fraction digits anymore, but only subunits".

Look at the last two commits I added, specifically `testNoFractionDigitsNoSubunitsNoPattern`. The default behaviour that `intl` gives us is fine for the majority of the cases. The case that @bramstroker has added is the case when you want to display a different amount of fractions than the number of subunits that a currency has.

That is what this PR fixes. You now have control over both.

sagikazarmark: Sorry, I missed this check:
`if ($this->subunits === null) {`

I thought there was no check for being set, so using null always casted to integer (0) seemed to be unnecessary.

frederikbosch: Alright, what do you think? Merge it in?

sagikazarmark:
Just for me to understand: this way if subunits are NOT defined, we use NumberFormatter fraction digits as subunits?

frederikbosch:
@sagikazarmark Correct.

sagikazarmark:
Why do we do that? At least I wouldn't expect this behaviour if I were a stranger.

frederikbosch:
@sagikazarmark Because `intl` is perfectly capable of coming up with the correct subunits. At least that is what I think. Maybe that should be tested better.

sagikazarmark:
I am still thinking about this use case:

``` php
$money = new Money(500, new Currency('EUR')); // this means 500 EUR, no subunits
$numberFormat->setAttribute(NumberFormatter::FRACTION_DIGITS, 2);
$formatter->format($money);
//Outputs € 5,00 not € 500,00
```

No subunits, but fraction digits.

frederikbosch:
@sagikazarmark See below the behaviour.

By default, when there is no fraction digit in the number formatter, `intl` takes care of it by using locale or maybe even the currency subunit default.
```php
$money = new Money(500, new Currency('EUR'));
$numberFormatter =  new \NumberFormatter('en_US', \NumberFormatter::CURRENCY);
$formatter = new IntlMoneyFormatter($numberFormatter);
$formatter->format($money); //Outputs € 5,00
```

But you can still set the number of fractions manually.
```php
$money = new Money(500, new Currency('EUR'));
$numberFormatter =  new \NumberFormatter('en_US', \NumberFormatter::CURRENCY);
$numberFormatter->setAttribute(NumberFormatter::FRACTION_DIGITS, 2);
$formatter = new IntlMoneyFormatter($numberFormatter);
$formatter->format($money); //Outputs € 5,00
```

But if you want to set a different number of fractions than the subunits, then you will have to set them both.
```php
$money = new Money(500, new Currency('EUR'));
$numberFormatter =  new \NumberFormatter('en_US', \NumberFormatter::CURRENCY);
$numberFormatter->setAttribute(NumberFormatter::FRACTION_DIGITS, 0);
$formatter = IntlMoneyFormatter::withSubunits($numberFormatter, 2);
$formatter->format($money); //Outputs € 5
```

And 
```php
$money = new Money(500, new Currency('EUR'));
$numberFormatter =  new \NumberFormatter('en_US', \NumberFormatter::CURRENCY);
$numberFormatter->setAttribute(NumberFormatter::FRACTION_DIGITS, 0);
$formatter = IntlMoneyFormatter::withSubunits($numberFormatter, 0);
$formatter->format($money); //Outputs € 500
```

sagikazarmark:
> But if you want to set a different number of fractions than the subunits, then you will have to set them both.

This is what concerns me: why? Why are there any coupling between the two setting? I expect them to be separately configured. Also your use cases covered everything, but not my example. I understand that it could be something like:

``` php
$money = new Money(500, new Currency('EUR'));
$numberFormatter =  new \NumberFormatter('en_US', \NumberFormatter::CURRENCY);
$numberFormatter->setAttribute(NumberFormatter::FRACTION_DIGITS, 2);
$formatter = IntlMoneyFormatter::withSubunits($numberFormatter, 0);
$formatter->format($money); //Outputs € 500.00
```

But if I set the fraction digits, I don't expect it to behave as subunits, but fraction digits. If I don't set the subunit explicitly to 0, then I am unable to use JUST fraction digits.

So what I don't really understand that WHY it is necessary to use fraction digits as fallback if subunits are not set? Why can't we just have subunits defaulting to 0 and leaving fraction digits to the user?

That way:
- User has control over the process
- There is no coupling between configuration
- Easier to understand and document (from my POV)

frederikbosch:
@sagikazarmark Because it is the `IntlMoneyFormatter`. I would expect that the `intl` extension takes care of it these things. That is what the extension is for, right? If we force the subunits to be zero then the user `must` explicitly tell the formatter for every currency it wants to format how many subunits it has. Then we could better skip the extension.

frederikbosch:
@sagikazarmark It would be best if the `Currency` object would carry the number of subunits for that currency.

sagikazarmark:
> I would expect that the intl extension takes care of it these things. That is what the extension is for, right?

No. The term subunit does not exist in the number formatter. Think about the example I showed. EUR value, but no subunit, although there are smaller values than euros (cents).

The intl extension is to format a value according to a locale, not a currency. Of course it formats the currency, but only in terms of sign, fraction digits, etc. Subunit is not part of this.

> It would be best if the Currency object would carry the number of subunits for that currency.

No, because of the same reason. Subunit is a relative, and not strongly tied to currencies in every case. There might be a lowest possible value in a currency (like cents), but you might not always want to represent it. That's why subunit is a configuration of your application.

frederikbosch:
@sagikazarmark PR updated. I understand your points. Though I have to say I still do not fully agree. In my opinion you cannot construct a `Money` with or without subunits. If you construct a `Money` object you must always use the lowest possible denominator. That means in the case of EUR always cents.

```php
new Money(500, new Currency('EUR'));
```

That should always represent the same value, no matter which application you are using. In your case, that *might* mean € 5 but that *might* also mean € 500. That is just plain wrong. In my opinion that is € 5, period. The question how to format it belongs to the formatter, with or without fractions. Subunits should not be part of formatting in the first place. While typing this, I actually still believe that this PR should **not** be merged.

sagikazarmark:
Honestly, I agree with you, but I simply can't see a way to add subunit to the model correctly. This is not a problem on the formatter level, but the domain model level.

frederikbosch:
@sagikazarmark That is right. Let me give it another thought.

frederikbosch:
@sagikazarmark I also found out that there is something like `NumberFormatter::MULTIPLIER`. I do not know if it can help us. That how I also found out about #227.

frederikbosch:
@sagikazarmark By the way, everything that has been said here, is also true for a parser. The parser is also not aware of the amount of subunits.

jyggen:
The problem with using `NumberFormatter::FRACTION_DIGITS` is that it's for the locale and not the currency, so using it to determine the subunit is dangerous. You've kind of agreed on that already, but here's an example anyway where the price is formatted differently based on the user's configured locale which outputs the wrong price for a Japanese user:

```shell
---- 15.99 USD for en_US ----
Fraction Digits:    2
Format:             $15.99

---- 15.99 USD for sv_SE ----
Fraction Digits:    2
Format:             15,99 US$

---- 15.99 USD for fr_CA ----
Fraction Digits:    2
Format:             15,99 $ US

---- 15.99 USD for ja_JP ----
Fraction Digits:    0
Format:             $1,599.00
```

Code used to produce the output above:

```php
<?php
use Money\Currency;
use Money\Formatter\IntlMoneyFormatter;
use Money\Money;

require_once 'vendor/autoload.php';

function debug($price, $currency, $locale) {
    $money = new Money($price, new Currency($currency));
    $numberFormatter = new \NumberFormatter($locale, \NumberFormatter::CURRENCY);
    $moneyFormatter = new IntlMoneyFormatter($numberFormatter);

    echo '---- '.($price/100).' '.$currency.' for '.$locale.' ----'.PHP_EOL;
    echo 'Fraction Digits: '.$numberFormatter->getAttribute(\NumberFormatter::FRACTION_DIGITS).PHP_EOL;
    echo 'Format:          '.$moneyFormatter->format($money).PHP_EOL;
    echo PHP_EOL;
}

debug(1599, 'USD', 'en_US');
debug(1599, 'USD', 'sv_SE');
debug(1599, 'USD', 'fr_CA');
debug(1599, 'USD', 'ja_JP');
```

sagikazarmark:
Thanks for the information, it is indeed very useful. I hope we can solve this issue soon.

frederikbosch:
@sagikazarmark My final thoughts are that we should create a subpackage `minor-units` or `subunits`. Quite similar to the creation of the currencies, we can create an array with the code as key and the number of subunits as value.

```php
<?php return [
  'EUR' => 2,
  'USD' => 2,
 // and so on
]
```

And do something like this.

```
$formatter = new MoneyIntlFormatter(new \NumberFormatter(), new CurrencySubunitPair());
```

sagikazarmark:
Why not an extension of currencies then? `SubunitCurrencies` interface can have that method and we don't need another term within the project and we could even implement it in the aggregate currencies stuff.

Also, I would make this thing optional in formatters: named constructor or something.

frederikbosch:
@sagikazarmark Good idea. Let's do that.

frederikbosch:
Just issued a new PR. Please comment your feedback over there. I would love to hear it.

PR #231, opened 2016-05-03
Fixes #227.

### LINKED ISSUE #227: Money does not allow values higher than PHP_INT_MAX
This does not work, while it should.

```php
(new Money(PHP_INT_MAX, new Currency('EUR')))->add(new Money(1, new Currency('EUR')))
```

Cause.
```php
filter_var($amount, FILTER_VALIDATE_INT) === false
```

Apparently, `filter_var` does have `PHP_INT_MAX` as its upper boundary. So we should validate with another method. Of course, regular expressions will work, but I wonder if there is not a faster method.

frederikbosch:


src/Number.php
```diff
@@ -157,4 +157,25 @@ public function getIntegerRoundingMultiplier()
...
+     *
+     * @return bool
+     */
+    public static function isInteger($number)
+    {
+        if (filter_var($number, FILTER_VALIDATE_INT) !== false) {
+            return true;
+        }
+
+        $invalid = array_filter(
```
sagikazarmark: Not sure this is the fastest, but we can improve that if necessary without breaking API. I will try to write some benchmarks.

tests/MoneyTest.php
```diff
@@ -26,4 +27,13 @@ public function testJsonEncoding()
             json_encode(Money::EUR(350))
         );
     }
+
+    public function testMaxInit()
```
sagikazarmark: Thumb rule: everything, that is not functional nature test should go to specs.

src/Number.php
frederikbosch: Exactly my thoughts, I merged it because I was happy on the API. And actually, I thought this was pretty clever ;).

sagikazarmark: Yeah, I agree, quite clever. :smile:

tests/MoneyTest.php
frederikbosch: @sagikazarmark I was doing that, was fighting with it actually because I got errors that I was not able to fix quickly. Decided to leave it. My bad.

PR #234, opened 2016-07-05
Improves MoneySpec. Makes test of a subtraction more accurate.

Sharom:
Order of operand is important to subtraction.
This commit fixes spec to check order of operands.


sagikazarmark:
Awesome, thank you very much.

PR #236, opened 2016-07-09
Cast amount to integer on JSON conversion.

Garbee:
The amount can be losslessly converted to an integer on the JSON output. This will provide the best experience to developers building APIs using the money object since their users won't need to cast the output on their end.

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -442,7 +442,7 @@ final class Money implements \JsonSerializable
     public function jsonSerialize()
     {
         return [
-            'amount' => $this->amount,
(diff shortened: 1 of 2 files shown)


sagikazarmark:
Well, this is not completely correct. See the following example: `var_dump(json_decode(json_encode(PHP_INT_MAX+1)))`

This will be a float value, not integer.

Garbee:
Ah, right. This would mean only about 21 million could be handled well. If anyone does anything going over that it is going to get wonky.

PR #238, opened 2016-07-17
Added number of sub-units into Currency object, Money::fromFloat introduced

Achse:
Hello, here is my small change. 
* I'd like to introduce `numberOfSubunits` into `Currency` object.
* The purpose for this is ability to create Money object from alternative construct method `Money::fromFloat()`.
 * Usages for this method can be: 
   * Parsing user input into `Money` object.
    * Create `Money` object for Crypto-currencies like bitcoin (Satoshi, `numberOfSubunits => 8`) or Ethereum (Theter, `numberOfSubunits = 12`), etc...

I also:
* Added `__toString` method for `Money` (mainly for test and development reasons - I found it handy.)
* Added `Money::normalize` to make strings like `1.000000` to `1`. (Necesary for `__toString()`)
* Added `pow` method for (`10^n`) into Calculator interfaces and all implementations.

Under the hood:
* I made `getCalculator()` method static to be usable in alternative static constructors (`fromFloat`). Also, because it works only with static properties, I think it should be static anyway.

Thanks for your review and comments.

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -61,6 +61,27 @@ final class Money implements \JsonSerializable
     }
 
     /**
+     * @param float|string $amount
+     * @param Currency     $currency
+     *
+     * @return static
+     */
+    public static function fromFloat($amount, Currency $currency)
+    {
+        if (!is_numeric($amount)) {
+            throw new \InvalidArgumentException('Amount must be a number.');
+        }
+
+        $calculator = self::getCalculator();
+        $amount = $calculator->multiply(
+            (string) $amount,
+            $calculator->pow('10', (string) $currency->getNumberOfSubUnits())
+        );
+
+        return new static(self::normalize($amount), $currency);
+    }
+
+    /**
      * Convenience factory method for a Money object.
      *
      * <code>
@@ -144,7 +165,7 @@ final class Money implements \JsonSerializable
     {
         $this->assertSameCurrency($other);
 
-        return $this->getCalculator()->compare($this->amount, $other->amount);
+        return self::getCalculator()->compare($this->amount, $other->amount);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/Money.php
```diff
@@ -61,14 +62,30 @@ public function __construct($amount, Currency $currency)
     }
 
     /**
+     * @param float|string $amount
+     * @param Currency $currency
+     * @return static
+     */
+    public static function fromFloat($amount, Currency $currency)
+    {
+        if (!is_numeric($amount)) {
```
Achse: `is_numeric` supports "big-numbers" (source: http://php.net/manual/en/function.is-numeric.php#102011)

frederikbosch:
Sorry, but this PR is not going to be merged.

1. __toString was discussed before and refused.
2. normalize is not needed, we already have normalizers for that.
3. why do need pow if there is no `Money` method for it?
4. When do we need `fromFloat`?

Achse:
Hey @frederikbosch, thank you for your replay. I am really sorry about that an I'll try to explain it.

1. About `__toString()`, no problem, I can remove it from pull request.
2. Without`__toString()` there will be no need for `normalize`. (Btw can you point me to thouse normalizers?)

But in those other points I feel really sad about:
3. Method `pow` is needed to transform `numberOfSubUnits` (for example dollar has 2 levels) into amount of smallest values for 1 unit (`100 cents` in this example) and is used in `fromFloat` method to create integer-like string value from float-like string value.
 * If you really dislike this way, it can be rewritten in way that Currency will store directly 100. But I used this log_10 way, because if cryptocurrencies can have realy many subunits (For example cryptocurrency Ethereum have `1 Ether = 1000000000000000000 Wei`.
4. Its not "literary" float. Mostly it is for creating Money from string that contains value like: `1.34` This is really needed in cases
 * you are creating money object from users input (Form mostly),
 * when you working with crypto-currencies and most of APIs and other systems works with exact string in "float"-like format.

Thank you for reconsideration of this feature.

frederikbosch:
You should a parser instead of `fromFloat`. See the StringToUnitsParser. There is a more comprehensive solution towards subunits on the way.

PR #241, opened 2016-07-26
add absolute value support

bendavies:
Adds a method to `Money` and calculators for returning the absolute value.

### DIFF
--- a/spec/Calculator/CalculatorBehavior.php
+++ b/spec/Calculator/CalculatorBehavior.php
@@ -62,6 +62,12 @@ trait CalculatorBehavior
         $this->floor(-2.7)->shouldReturn('-3');
     }
 
+    function it_calculates_the_absolute_value()
+    {
+        $this->absolute(2)->shouldReturn('2');
+        $this->absolute(-2)->shouldReturn('2');
+    }
+
     function testShare()
     {
         $this->share(10, 2, 4)->shouldReturn('5');
--- a/spec/MoneySpec.php
+++ b/spec/MoneySpec.php
@@ -388,4 +388,32 @@ class MoneySpec extends ObjectBehavior
             ['-1', false, false, true],
         ];
     }
+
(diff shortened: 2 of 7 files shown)


frederikbosch:
@bendavies Are you abs for money or other purposes? I have the feeling that you like the classes to be used outside `Money`, right?

frederikbosch:
Oh sorry, I did not see the absolute method on `Money`. It actually looks quite good.

frederikbosch:
@sagikazarmark I think we can merge this one in.

bendavies:
note that i've called it `abs` on the calculators and `absolute` on money.
which do you prefer? shall we name them the same?

spec/Calculator/CalculatorBehavior.php
```diff
@@ -62,6 +62,12 @@ function it_floors_a_value()
         $this->floor(-2.7)->shouldReturn('-3');
     }
 
+    function it_absolutes_a_value()
```
sagikazarmark: it_calculates_the_absolute_value?

spec/MoneySpec.php
```diff
@@ -388,4 +388,32 @@ function comparatorExamples()
             ['-1', false, false, true],
         ];
     }
+
+
+    /**
+     * @dataProvider absoluteExamples
+     */
+    function it_absolutes_the_amount($amount, $result, Calculator $calculator)
```
sagikazarmark: same here

sagikazarmark:
Two small comments, LGTM apart from those.

frederikbosch:
@bendavies I rather name the methods the same: `absolute` for `Money` and `Calculator` implementations.

bendavies:
comments addressed. 
@sagikazarmark went for `it_calculates_the_absolute_amount/value` which follows the convention of the existing specs on the class.

PR #242, opened 2016-07-26
Fix specs

bendavies:
The specs i've fixed weren't asserting the results

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,17 +23,18 @@
         "php": ">=5.5"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.5",
         "ext-bcmath": "*",
         "ext-gmp": "*",
         "ext-intl": "*",
-        "florianv/swap": "^2.3",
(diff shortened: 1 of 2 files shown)


frederikbosch:
Why do need `beberlei/assert` to [check if something is true](https://github.com/beberlei/assert/blob/master/lib/Assert/Assertion.php#L1316)?

sagikazarmark:
We don't. I got the point, I will fix these tests and close this PR.

bendavies:
@frederikbosch because the specs weren't asserting/checking anything at all

sagikazarmark:
No need for assertion: `->shouldReturn(something);`

bendavies:
@sagikazarmark not possible without refactoring as the return from `equal` isn't a prophecy, it's a real `Money` object

bendavies:
@sagikazarmark are you going to fix this yourself?

frederikbosch:
Fixed in PR #260.

Issue #221, opened 2016-04-26
Numberformatter returning wrong formatted amount


bramstroker:
Following you documentation we should construct new `Money` objects using the smallest unit (eg. cents). This means `new Money(100, 'EUR')` is the same as 1 Euro.

When using the IntlMoneyFormatter which uses NumberFormatter I get the wrong amount outputted.

Output:
€ 100

Expected:
€ 1

The `IntlMoneyFormatterSpec` test also is wrong imo.

sagikazarmark:
Can you please try setting the fraction digit attribute of the number formatter?

https://github.com/moneyphp/money/blob/master/tests/Formatter/IntlMoneyFormatterTest.php#L24

bramstroker:
@sagikazarmark 
I get a "correct" value when using fraction digits 2.

€ 1,00

But this is not the same behaviour as using the NumberFormatter without the Money object. i.e.
Fractional digits should only affect the position of dots and comma's and decimal positions, but never affect the base value.

    $numberFormat = new NumberFormatter('nl_NL', NumberFormatter::CURRENCY);
    $numberFormat->setAttribute(NumberFormatter::FRACTION_DIGITS, 2);
    echo $numberFormat->formatCurrency(500, 'EUR') . PHP_EOL;
    //Outputs € 500,00

    $numberFormat->setAttribute(NumberFormatter::FRACTION_DIGITS, 0);
    echo $numberFormat->formatCurrency(500, 'EUR') . PHP_EOL;
    //Outputs € 500

sagikazarmark:
Fraction digit in our case is the same as the subunit. Unfortunately we haven't found a better way to use the core NumberFormatter. Subunit is not part of the money object, that's why you must configure it in the formatter (in our case, in the number formatter).

If you have a better idea, please give us a hint, but I think @frederikbosch investigate the possibilities quite well.

bramstroker:
@sagikazarmark 
Thanks for the explanation. I now understand why you choose to do it like this.
You mean by subunits the amount of decimals which needs to be substracted to get to the "normal" value? I.e.  500 / 100 = 5
Maybe it would be a good idea to add that "factor" to the constructor of the `IntlMoneyFormatter`, That way it would not alter the core NumberFormatter behaviour and the implementation can be much simpler imo by just dividing the amount by the specified factor instead of all the string padding what happens now,

sagikazarmark:
[...]

PR #233, opened 2016-05-24
add method to get money subunits in formatters

iampersistent:
this allows you to get the subunits without the currency formatting

Issue #232, opened 2016-05-13
Simple Numeric Formatting - How?


judgej:
I have a money object with some amount in it. I want to put this into a form field so it can be edited, or put into a report to tabulate. In both cases I need the amount appropriately formatted, with a decimal point and trailing zeroes where needed, but no currency symbol. How would this be achieved?

I'm just looking for the quick conversion between what is stored, and what a human would write the amount down as (without adding the currency symbol) in both directions.

judgej:
Just a simple `(string)$myMoney` would be nice.

sagikazarmark:
@judgej Do you use the 1.x stable or the 3.x upcoming version?

In the upcoming version we are working hard on making Money as easy to use as possible.

What you are trying to achieve is quite hard, since the amount is internally stored in integer format, which means Money (Currency) should be aware of the subunit. We tried to avoid this and make the value objects as simple and flexible as possible. Right now we are working on a formatting solution which involves a separate formatter object and a currency repository.

Casting the Money object to string could be a nice addition, but I guess everyone would expect a different result: just the amount, amount with currency, amount with currency symbol, etc.

judgej:
I started on 1.3 b but have moved to the 3.x beta.

So the currency here is not aware of the size of the subunit? I thought I was going mad when I couldn't find it. To have to keep that information separate from the currency itself seems to defeat half the object of this library.

Humans and currency exchange services rates work in major units and minor units. Computers can work largely in minor units. Accounts like to work in fractions of minor units. Being able to easily move between these forms is what is going to make this library much more useful.

sagikazarmark:
> So the currency here is not aware of the size of the subunit?

No, this library never did that. The main reason was that currencies themselves does not necessarily have minor units (although most physical currencies have).

Right now we are thinking about two possible solutions:

1. We already have the term of currency repository which contains basic and meta information for whatever currency you would like to use (see #223)
2. The other possibility is to add subunit support to the currency object itself. However, there are two problems with that:
    - You would have to create currency objects like `new Currency('EUR', 2)`
    - Convenient `Money::EUR(123)` factories wouldn't work anymore

Also, this makes the creation of a money object rather hard: `new Money(123, new Currency('EUR', 2))` In this case, is the input amount €123 or €1.23?

Minor units are actually useful if you want to format the amount. Using 1 or 100 for calculations doesn't matter, it only matters when you want to output the value.

This is a major problem we are aware of and blocks the 3.0 version, but there is no nice solution for this:

[...]

PR #248, opened 2016-07-28
Extend currencies with subunit

frederikbosch:
Format currencies with subunits by using a subunit provider. Built-in providers are Constant and ISO. A follow-up PR might refactor `ISOCurrencies` and `ISOProvider` to use a single `ISORepository` or `ISOFactory`.

Questions:
1. Is `Provider` the correct name in this context?
2. What do you feel about the architectural decisions?
3. What about namespaces? Is everything in the correct place?

Please give me feedback on those important issues. Afterwards we can discuss secondary style issues. /cc @sagikazarmark 

Fixes: #221 #232 #233 #223.

### DIFF
--- a/spec/Formatter/IntlMoneyFormatterSpec.php
+++ b/spec/Formatter/IntlMoneyFormatterSpec.php
@@ -2,6 +2,7 @@
 
 namespace spec\Money\Formatter;
 
+use Money\Currencies;
 use Money\Currency;
 use Money\Money;
 use Money\MoneyFormatter;
@@ -9,9 +10,9 @@ use PhpSpec\ObjectBehavior;
 
 class IntlMoneyFormatterSpec extends ObjectBehavior
 {
-    function let(\NumberFormatter $numberFormatter)
+    function let(\NumberFormatter $numberFormatter, Currencies $currencies)
     {
-        $this->beConstructedWith($numberFormatter);
+        $this->beConstructedWith($numberFormatter, $currencies);
     }
 
     function it_is_initializable()
@@ -24,12 +25,12 @@ class IntlMoneyFormatterSpec extends ObjectBehavior
         $this->shouldImplement(MoneyFormatter::class);
     }
 
-    function it_formats_money(\NumberFormatter $numberFormatter)
+    function it_formats_money(\NumberFormatter $numberFormatter, Currencies $currencies)
     {
-        $numberFormatter->getAttribute(\NumberFormatter::FRACTION_DIGITS)->willReturn(2);
-        $numberFormatter->formatCurrency('0.01', 'EUR')->willReturn('€1.00');
-
         $money = new Money(1, new Currency('EUR'));
+
+        $numberFormatter->formatCurrency('0.01', 'EUR')->willReturn('€1.00');
+        $currencies->subunitFor($money->getCurrency())->willReturn(2);
 
         $this->format($money)->shouldReturn('€1.00');
     }
--- a/src/Currencies/ISOCurrencies.php
+++ b/src/Currencies/ISOCurrencies.php
@@ -4,6 +4,7 @@ namespace Money\Currencies;
 
 use Money\Currencies;
 use Money\Currency;
+use Money\Exception\UnknownCurrencyException;
 
 /**
  * List of supported ISO 4217 currency codes and names.
@@ -31,6 +32,37 @@ final class ISOCurrencies implements Currencies
         return isset(self::$currencies[$currency->getCode()]);
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function subunitFor(Currency $currency)
+    {
+        if (null === self::$currencies) {
+            self::$currencies = $this->loadCurrencies();
+        }
+
+        if (!isset(self::$currencies[$currency->getCode()])) {
+            throw new UnknownCurrencyException('Cannot find ISO currency '.$currency->getCode());
+        }
+
+        return self::$currencies[$currency->getCode()]['minorUnit'];
+    }
+
+    /**
+     * @return \Traversable
+     */
+    public function getIterator()
+    {
+        if (null === self::$currencies) {
+            self::$currencies = $this->loadCurrencies();
+        }
+
+        return new \ArrayIterator(array_keys(self::$currencies));
+    }
+
+    /**
+     * @return array
+     */
     private function loadCurrencies()
     {
         $file = __DIR__.'/../../vendor/moneyphp/iso-currencies/resources/current.php';
--- a/src/Formatter/IntlMoneyFormatter.php
+++ b/src/Formatter/IntlMoneyFormatter.php
@@ -2,6 +2,7 @@
 
 namespace Money\Formatter;
 
+use Money\Currencies;
 use Money\Money;
 use Money\MoneyFormatter;
 
@@ -16,13 +17,19 @@ final class IntlMoneyFormatter implements MoneyFormatter
      * @var \NumberFormatter
      */
     private $formatter;
+    /**
+     * @var Currencies
+     */
+    private $currencies;
 
     /**
      * @param \NumberFormatter $formatter
+     * @param Currencies       $currencies
      */
-    public function __construct(\NumberFormatter $formatter)
+    public function __construct(\NumberFormatter $formatter, Currencies $currencies)
     {
         $this->formatter = $formatter;
+        $this->currencies = $currencies;
     }
 
     /**
@@ -33,25 +40,25 @@ final class IntlMoneyFormatter implements MoneyFormatter
         $valueBase = (string) $money->getAmount();
         $negative = false;
 
-        if (substr($valueBase, 0, 1) === '-') {
+        if ($valueBase[0] === '-') {
             $negative = true;
             $valueBase = substr($valueBase, 1);
         }
 
-        $fractionDigits = $this->formatter->getAttribute(\NumberFormatter::FRACTION_DIGITS);
+        $subunit = $this->currencies->subunitFor($money->getCurrency());
         $valueLength = strlen($valueBase);
 
-        if ($valueLength > $fractionDigits) {
-            $subunits = substr($valueBase, 0, $valueLength - $fractionDigits).'.';
-            $subunits .= substr($valueBase, $valueLength - $fractionDigits);
+        if ($valueLength > $subunit) {
+            $formatted = substr($valueBase, 0, $valueLength - $subunit).'.';
+            $formatted .= substr($valueBase, $valueLength - $subunit);
         } else {
-            $subunits = '0.'.str_pad('', $fractionDigits - $valueLength, '0').$valueBase;
+            $formatted = '0.'.str_pad('', $subunit - $valueLength, '0').$valueBase;
         }
 
         if ($negative === true) {
-            $subunits = '-'.$subunits;
+            $formatted = '-'.$formatted;
         }
 
-        return $this->formatter->formatCurrency($subunits, $money->getCurrency()->getCode());
+        return $this->formatter->formatCurrency($formatted, $money->getCurrency()->getCode());
     }
 }
--- a/spec/Currencies/AggregateCurrenciesSpec.php
+++ b/spec/Currencies/AggregateCurrenciesSpec.php
@@ -9,6 +9,8 @@ use Prophecy\Argument;
 
 class AggregateCurrenciesSpec extends ObjectBehavior
 {
+    use HaveCurrencyTrait;
+
     function let(Currencies $isoCurrencies, Currencies $otherCurrencies)
     {
         $this->beConstructedWith([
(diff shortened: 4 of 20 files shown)


sagikazarmark:
I would rather create a repository interface extension: `CurrenciesWithSubunit`

``` php
function getSubunitFor(Currency $currency);
```

By definition a subunit should always be an integer value and I think we should fall back to 0 if we have no information.

pamil:
Anyway, we should stick to either `subunit` or `subUnit` :) The first one looks better for me.

frederikbosch:
@sagikazarmark Why do want to extend another interface?

frederikbosch:
@pamil That can easily be renamed. I can do that.

frederikbosch:
@sagikazarmark Actually, that is not completely different from what I am proposing. Just different names.

sagikazarmark:
Well, it does not have to be an extension, but IMO we should implement these interfaces in monolithic repositories. Kind of breaks SRP, but it's repository pattern.

sagikazarmark:
No, it's not different, but fits into the domain better IMHO.

frederikbosch:
@sagikazarmark Refactoring... give me two minutes.

frederikbosch:
@sagikazarmark Is this any better?

frederikbosch:
@pamil What do you think?

frederikbosch:
`CurrenciesWithSubunit` might be better of in `Money` namespace instead of the `Money\Currencies` namespace since `Currencies` is also in the `Money` namespace.

spec/Formatter/IntlMoneyFormatterSpec.php
```diff
@@ -24,13 +25,13 @@ function it_is_a_money_formatter()
...
-    function it_formats_money(\NumberFormatter $numberFormatter)
+    function it_formats_money(\NumberFormatter $numberFormatter, CurrenciesWithSubunit $currenciesWithSubunit)
     {
-        $numberFormatter->getAttribute(\NumberFormatter::FRACTION_DIGITS)->willReturn(2);
-        $numberFormatter->formatCurrency('0.01', 'EUR')->willReturn('€1.00');
-
         $money = new Money(1, new Currency('EUR'));
 
+        $numberFormatter->formatCurrency('0.01', 'EUR')->willReturn('€1.00');
+        $currenciesWithSubunit->getSubunitsFor($money->getCurrency())->willReturn(2);
```
sagikazarmark: Hm, I am not sure what is gramatically correct: subunit or subunitS. The first seems to be more logical to me. Can you check which is correct?

src/Currencies/ConstantSubunits.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    /**
+     * @var int
+     */
+    private $subUnits;
+
+    /**
+     * @param int $subUnits
+     */
+    public function __construct($subUnits)
```
sagikazarmark: As per the spec, shouldn't this default to 0?

src/Currencies/CurrenciesWithSubunit.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Money\Currencies;
+
+use Money\Currency;
+
+interface CurrenciesWithSubunit
```
sagikazarmark: author block and docblock

src/Currencies/ISOCurrencies.php
```diff
@@ -31,6 +32,25 @@ public function contains(Currency $currency)
...
+     * {@inheritdoc}
+     */
+    public function getSubunitsFor(Currency $currency)
+    {
+        if (null === self::$currencies) {
+            self::$currencies = $this->loadCurrencies();
+        }
+
+        if (!isset(self::$currencies[$currency->getCode()])) {
+            throw new UnknownISOCurrencyException('Cannot find ISO currency '.$currency->getCode());
```
sagikazarmark: UnkownCurrencyException would be better? Could be used by other repositories too.

src/Currencies/ISOCurrencies.php
```diff
@@ -31,6 +32,25 @@ public function contains(Currency $currency)
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getSubunitsFor(Currency $currency)
+    {
+        if (null === self::$currencies) {
+            self::$currencies = $this->loadCurrencies();
+        }
+
+        if (!isset(self::$currencies[$currency->getCode()])) {
```
sagikazarmark: Hm, should we do this or always return with zero if a currency is not found?

src/Formatter/IntlMoneyFormatter.php
```diff
@@ -16,13 +17,19 @@
      * @var \NumberFormatter
      */
     private $formatter;
+    /**
+     * @var CurrenciesWithSubunit
+     */
+    private $currenciesWithSubunit;
```
sagikazarmark: IMO $currencies is enough

frederikbosch:
@sagikazarmark Since you are already giving detailed comments I assume you like this design better then the initial one?

sagikazarmark:
Your assumption is correct. :wink:

frederikbosch:
Haha, ok. I will fix those issues and think about the default of zero. Don't know for sure if that is what we want.

frederikbosch:
Just updated the PR. I did not got for the default subunit of zero. Rather I added a `FallbackSubunit` class that can catch errors from a delegated `CurrenciesWithSubunit` a then use a fallback from another `CurrenciesWithSubunit`. Basically, the fallback will act as a decorator.

frederikbosch:
Also went for `subunit`: singular.

frederikbosch:
By the way, I have not created spec tests for the different `CurrenciesWithSubunit` because I want to have an agreement on the architecture. I will add those.

frederikbosch:
Actually I think this solution it far better than others we discussed before. Thanks for the suggestion @sagikazarmark. I will leave this here for some days. Without objections this will be merged somewhere next week.

frederikbosch:
My reasoning for not using a subunit default is as follows.

1. The `ISOCurrencies` class is only aware of the how and what of ISO currencies. A default is only appropriate if there is a ISO default. I do not believe there is one. Moreover, it would be weird when `contains` returns false and `getSubunitFor` actually returns a number. So it should indeed throw an exception.
2. `ConstantSubunit` is a class where you - if you use it - can inject the number zero without any problem. That is the whole idea of the class.

Hence, we need a third class: `FallbackSubunit` that applies a constant when the delegated is failing.
```php
<?php
use Money\Currencies\ConstantSubunit;
use Money\Currencies\FallbackSubunit;
use Money\Currencies\ISOCurrencies;
use Money\Formatter\IntlMoneyFormatter;

$formatter = new IntlMoneyFormatter(
  new \NumberFormatter(...),
  new FallbackSubunit(
    $currencies,
    new ConstantSubunit(0)
  )
);
```

Sharom:
I think, that we can add `CurrencySpecification` value.

`CurrencySpecification` should contain information about subunits, name of subunit and so on.

Also we should provide `CurrencyRepositoryInterface`:

```php
interface CurrencySpecificationRepositoryInterface
{
    /**
     * @param Currency $currency
     * @return CurrencySpecification|null
     */
     public function findCurrencySpecification(Currency $currency);
}
```

After that we can inject an instance of `CurrencySpecificationRepositoryInterface` into the formatter/parser.

And `CurrencyRepository` can provide currency by currency specifications =)

frederikbosch:
@Sharom I took your idea and merged into the branch. It is little different from your proposal but I definitely like the interface `CurrenciesSpecification` better than `CurrenciesWithSubunit`. The new value object `Specification` should throw exception I guess instead of returning null when it does not know a specific value (numeric code or name of a currency). But I'd like to hear first how @sagikazarmark and @pamil think about this compared to what we had before.

frederikbosch:
Now, I think that CurrenciesSpecification should be moved into Currencies. So Currencies interface has two methods: contains and specify. If the repository contains the currency, it must be able to specify its characteristics. Right?

Sharom:
> Now, I think that CurrenciesSpecification should be moved into Currencies. So Currencies interface has two methods: contains and specify. If the repository contains the currency, it must be able to specify its characteristics. Right?

@frederikbosch I am not sure. I think that it's better to have separated interfaces. `Currency` and `CurrencySpecification` are different things. Therefore a repository of currencies and repository of specifications are different too. Two responsibilities of a one interface smell bad.

But currency can be created by a specification. In this case `Currencies` can use repository of specifications.

I suggest not to rush to a final decision. Let's think about it.

src/CurrenciesSpecification.php
```diff
@@ -0,0 +1,20 @@
...
+namespace Money;
+
+use Money\Currencies\Specification;
+
+/**
+ * Specifies information about a currency.
+ *
+ * @author Frederik Bosch <f.bosch@genkgo.nl>
+ */
+interface CurrenciesSpecification
```
Sharom: I think, that it's better to rename this interface like  `CurrencySpecifications`, isn't it?

src/CurrenciesSpecification.php
```diff
@@ -0,0 +1,20 @@
...
+{
+    /**
+     * @param Currency $currency
+     *
+     * @return Specification
```
Sharom: And I would rename `Specification` -> `CurrecySpecification`. I think that it will be clearer.

src/CurrenciesSpecification.php
Sharom: I am not sure in choice of name for this interface.

sagikazarmark:
> Now, I think that CurrenciesSpecification should be moved into Currencies. So Currencies interface has two methods: contains and specify. If the repository contains the currency, it must be able to specify its characteristics. Right?

This is Multiple Responsibility Principle too.

Can you give me some more usecases for a Specification object? Sounds like a generic metadata store, dunno if we need that.

sagikazarmark:
I am going to turn of the automatic MR creation from StyleCI. It's annoying.

frederikbosch:
@sagikazarmark Now I closed the other PR and got back here. And I also tried to work with `CurrenciesWithSubunit` like you are proposing. It simply does not work within our domain to have it as a new interface.

1. How should AggregateCurrencies look like when we have `CurrenciesWithSubunit`?
2. How should CachedCurrencies look like when we have `CurrenciesWithSubunit`?

While trying to make it work I found that the solutions can be.

1. We will end-up with dirty `instanceof` checks if we implement `CurrenciesWithSubunit` in those classes.
2. Or we have to create new classes `AggregateCurrenciesWithSubunit` and `CachedCurrenciesWithSubunit`. That becomes a nightmare.
3. We leave those classes untouched and we accept that there no aggregation and cache for subunit.

Basically decorating `Currencies` | `CurrenciesWithSubunit` becomes very hard if we add another interface for subunits. We should keep it simple. `subunitFor` is a requirement of our domain and belongs to something our repository should provide by default. Now if you have a specific currency without a subunit, you create your own `Currencies` implementation.

```php
namespace My;

use Money\Currencies;

class 8g7bv87b786cv8zuvb786bv7zv76v implement Currencies {

  public function contains(Currency $currency) {
     return $currency->getCode() === '8g7bv87b786cv8zuvb786bv7zv76v';
  }

  public function subunitFor(Currency $currency) {
    return 0;
  }
}
```

@sagikazarmark If you still have time... ;) Please provide some feedback here.

Sharom:
**One other case, related to subunits**

Sometimes we need to accomplish some rounding of money:

1. Let we have some thing with purchase price equal to `10.20 USD`;
2. And we want to sell it with ~12% profit: `10.20 USD * 12% = 11.42 USD`;
3. But price `11.42 USD` looks ugly, `10.50 USD` looks better!

This functionality may be implemented on `Money` in the future.
And we need information about subunits to be able do this.

frederikbosch:
I am going to wait until Travis finishes this one. I will squash it and merge it. There were many iterations and many experiments. I have done as much as I could with all the feedback. Now it is decision time. We have to move on.

frederikbosch:
This PR includes formatting and parsing with rounding. That means parsing $1000.005 for a currency with subunit of 2 will become `Money::US(100001)` and formatting  `Money::US(5055)` with subunit of 2 and formatting fraction digits of 0 will become '$51'.

frederikbosch:
This PR also includes `Currencies` extending `\IteratorAggregate`. So one we can iterate over every currency code inside a currency repository. So `Currencies` actually provides currency codes.

frederikbosch:
This PR does not include a solution to a name of a currency. Iteration via `\IteratorAggregate` allows to see which currency codes are available. The library user can translate a currency code to a currency name via its own objects.

frederikbosch:
Thanks @sagikazarmark @pamil @Sharom and others that helped giving the feedback.

sagikazarmark:
I am not entirely happy to have `subunitFor` in the Currencies interface. Subunit is not a core part of Currencies. I am still on the side of a separate interface. I don't care if both are implemented in the same class, but they seem to be separate responsibilities.

> How should AggregateCurrencies look like when we have CurrenciesWithSubunit

The answer is simple: it shouldn't change. Since they are different responsibilities, you need to use two different objects: one Currencies instance decorated with AggregateCurrencies and an other one is the Subunit repo. (If it was DI, we would have two separate services).

On the other hand, it would probably increase complexity which might not worth the effort.

A few cosmetic points:

- methods names are better when they are in imperative: `getSubunitFor`. Contains is kind of an exception
- document better when UnkownCurrencyException should be thrown (when repo does not contain currency via contains?)

frederikbosch:
@sagikazarmark Thanks for the feedback. I understand it very well. I will have a look at the `UnknownCurrencyException`.

PR #256, opened 2016-08-01
Extend currency with subunit and name

frederikbosch:
Instead of creating new objects like `CurrenciesSpecification` or `CurrenciesWithSubunit`, this solution adds functionality to `Currencies` and `Currency`. First of all: why this PR? I am not necessarily happy with PR #248. I still feel we are not there yet. Therefore I am trying different solutions. This should not be seen as how it must be, rather another shot at the problem. Maybe it contains something to continue on. Maybe it is crap.

But, here is why I think that this PR does contain some valid points.

1. Our `Currency` is coming short in our domain. A subunit is a property of the currency and should be modelled as such. There should not be another object for it. So no `CurrencySpecification`.

2. `Currencies` should - according to its description - 'provide a list of currencies'. But it is not. It only tells whether a currency is available within the context of the object. It is not *providing* a list. That is a problem. In my current applications every time I construct a `Money` object, I also construct a new `Currency`. Why? There is only one Euro or one US dollar currency. Hence my application should have one currency object per currency, not multiple. This implies that Currencies - because its responsibility is to *provide* currencies - should actually start doing that. This is what is achieved with the `find` method on the object. It returns a currency object and always the same object per currency code.

3. But now `Currencies` has more than one method on its interface: that is multiple responsibility! I disagree here. Single responsibility does not imply that an interface should have one method. Of course, functional interface (those with one method) should be preferred over those with many methods. But in this case, the domain requires there to be more than one method. I compare it with [container-interop](https://github.com/container-interop/container-interop/blob/master/src/Interop/Container/ContainerInterface.php#L14). That is not a bad interface, is it? `contains` <-> `has` and `find` <-> `get`.

Maybe you can shoot at the points I am making. Because I think we need some more iterations before we will find the solution we are looking for.

/cc @sagikazarmark @pamil @Sharom

### DIFF
--- a/spec/Currencies/AggregateCurrenciesSpec.php
+++ b/spec/Currencies/AggregateCurrenciesSpec.php
@@ -4,6 +4,7 @@ namespace spec\Money\Currencies;
 
 use Money\Currencies;
 use Money\Currency;
+use Money\Exception\UnknownCurrencyException;
 use PhpSpec\ObjectBehavior;
 use Prophecy\Argument;
 
@@ -35,12 +36,20 @@ class AggregateCurrenciesSpec extends ObjectBehavior
         $this->contains(new Currency('EUR'))->shouldReturn(true);
     }
 
-    function testItDoesNotContainCurrencies()
+    function testItDoesNotContainCurrencies(Currencies $isoCurrencies, Currencies $otherCurrencies)
     {
         $isoCurrencies->contains(Argument::type(Currency::class))->willReturn(false);
         $otherCurrencies->contains(Argument::type(Currency::class))->willReturn(true);
 
         $this->contains(new Currency('EUR'))->shouldReturn(false);
+    }
+
+    function it_throws_an_exception_when_currency_is_unknown(Currencies $isoCurrencies, Currencies $otherCurrencies)
+    {
+        $isoCurrencies->find('XXXX')->willThrow(UnknownCurrencyException::class);
+        $otherCurrencies->find('XXXX')->willThrow(UnknownCurrencyException::class);
+
+        $this->shouldThrow(UnknownCurrencyException::class)->duringFind('XXXX');
     }
 
     function testConstructorThrowsAnException()
--- a/spec/Currencies/BitcoinCurrenciesSpec.php
+++ b/spec/Currencies/BitcoinCurrenciesSpec.php
@@ -4,6 +4,7 @@ namespace spec\Money\Currencies;
 
 use Money\Currencies;
 use Money\Currency;
+use Money\Exception\UnknownCurrencyException;
 use PhpSpec\ObjectBehavior;
 use Prophecy\Argument;
 
@@ -23,5 +24,15 @@ class BitcoinCurrenciesSpec extends ObjectBehavior
     {
         $this->contains(new Currency('XBT'))->shouldReturn(true);
         $this->contains(new Currency('EUR'))->shouldReturn(false);
+    }
+
+    function it_can_find_bitcoin()
+    {
+        $this->find('XBT')->shouldReturnAnInstanceOf('Money\\Currency');
+    }
+
+    function it_throws_an_exception_when_currency_is_unknown()
+    {
+        $this->shouldThrow(UnknownCurrencyException::class)->duringFind('XXXX');
     }
 }
--- a/spec/Currencies/CachedCurrenciesSpec.php
+++ b/spec/Currencies/CachedCurrenciesSpec.php
@@ -59,4 +59,21 @@ class CachedCurrenciesSpec extends ObjectBehavior
 
         $this->contains(new Currency('EUR'))->shouldReturn(true);
     }
+
+    function it_finds_currencies_from_the_cache(
+        CacheItemInterface $item,
+        CacheItemPoolInterface $pool,
+        Currencies $currencies
+    ) {
+        $item->isHit()->willReturn(true);
+        $item->set(Argument::type(Currency::class))->shouldNotBeCalled();
+        $item->get()->willReturn(new Currency('EUR'));
+
+        $pool->getItem('currency|code|EUR')->willReturn($item);
+        $pool->save($item)->shouldNotBeCalled();
+
+        $currencies->find(Argument::type(Currency::class))->shouldNotBeCalled();
+
+        $this->find('EUR')->shouldReturnAnInstanceOf('Money\\Currency');
+    }
 }
--- a/spec/Currencies/ISOCurrenciesSpec.php
+++ b/spec/Currencies/ISOCurrenciesSpec.php
@@ -32,11 +32,6 @@ class ISOCurrenciesSpec extends ObjectBehavior
         $this->shouldThrow(UnknownCurrencyException::class)->duringFind('XXXX');
     }
 
-    function it_provides_a_list_of_currencies()
-    {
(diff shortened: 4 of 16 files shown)


frederikbosch:
Updated the PR a little more. No more entity property on the currency. There is no real use-case for it other than meta data. The currency name however does have a use-case. I can image people wanting to render a currency list, so one should be able to [iterate over `ISOCurrencies`](https://github.com/moneyphp/money/blob/currency-extension/src/Currencies/ISOCurrencies.php#L71) and display the name of currency.

By the way @sagikazarmark, with this setup, there is [a default subunit of zero](https://github.com/moneyphp/money/blob/currency-extension/src/Currency.php#L23).

sagikazarmark:
There are multiple - and I have to say - serious problems with this PR.

> A subunit is a property of the currency and should be modelled as such.

No, it is definitely not. Subunit is something logical which you mostly have for "real" currencies. You cannot buy a half bitcoin. Also, from the Money point of view this subunit doesn't really matter as long as you use the same logical unit for addition/substraction. It only matters when you "present" the value.

Can you tell me a use case when you needed subunits, but it wasn't a presentation?

> Currencies should - according to its description - 'provide a list of currencies'.

This is probably a little bit misleading, but the point is still the same: provide some kind of repository for currencies. I think it fits into the SRP to have a check method in it (we might also want to add other things here in the future), but if subunit is a totally different kind of data, then it should not be included in this interface. My preference is still the CurrenciesWithSubunit interface.

> There is only one Euro or one US dollar currency. Hence my application should have one currency object per currency, not multiple.

Definitely not. We apply the Value Object pattern here where equality is based on the VALUE that's represented, not the object reference. In the Currency, this value is the currency code, in the Money the value is the numeric value AND the currency together (see `equals` functions)

For the moment I don't think having a find method in the repository makes sense. You can simply create a new currency and check validity manually.

frederikbosch:
@sagikazarmark Thanks for your reply. It really helps to see the problem better. I agree with most of the points your make, but disagree with the fact that a subunit is completely different data. Every currency has a subunit. Even a [Bitcoin has a subunit](http://bitcoindenominations.org/). The value (8) is even greater than for any other currency! So you can actually own a half Bitcoin!

Then, let's take another attribute into account: the name of currency. And let's pick a simplified use-case. Suppose we have an application with multiple webshops. Every shop could have a different currency. So I want to render a list of currencies. And on my shop settings page I want to display the currency name.

1. Who is providing the list of names? Probably the `Currencies` implementation by providing an `Traversable` over it, where the currency code is the key, and the name of the currency is the value.

2. So now, we save the currency code with our shop entity.

3. But how do we display it on our shop settings page? The only thing we can do is iterate over `Currencies` again, lookup our currency code in the key and then display that. That is not a practical solution.

Suppose we want to provide a better solution. If we continue on the path of creating a new interface we create the interface `CurrenciesWithName`. So then we have the interfaces `Currencies`, `CurrenciesWithName` and `CurrenciesWithSubunit`. Three concepts in our domain while in real life there is one concept currency with three attributes code, name and subunit.

Now I'd say we add `name` and `subunit` to

1. `Currency` by the `withName` and `withSubunit` which will be inject via `Currencies`
2. `Currencies` via the methods `subunitFor` and `nameFor`.

sagikazarmark:
> Even a Bitcoin has a subunit

Okay, let's close to this problem from another point of view: Currency is a VO and as such, it should only have the minimum internal structure that helps representing that single value. That's the currency code. In case of money, it's the number AND the currency. Think about money in real life: when you hold a ten EUR note, it doesn't tell you that there are cents too.

About having further details in currency: take a look at this standard proposal: https://github.com/e-ig/standards/pull/3

It contains many other proposed fields and I had quite a few fights with people in that discussion to explain why it is a bad idea.

I see a few possible solutions for the problem:

- subunit is a problem for us at the moment, so we need to solve it. CurrenciesWithSubunit is a possible solution. As for the name: everyone should solve it on their own :trollface: The biggest problem in my opinion is that Name specifically is usually a translated field and I really don't want to provide translations. Every framework handles translations separately.
- we can go for the CurrencySpecification solution too (although I would rather call them CurrencyDetail(s)), but we should then probably make it a general metadata storage: require a name, require a subunit, allow adding an optional array with any other meta data (maybe require the Currency object for reference too?). However in this case we have to declare the behaviour for dependent components: what should happen if there is no subunit?

frederikbosch:
> Okay, let's close to this problem from another point of view: Currency is a VO and as such, it should only have the minimum internal structure that helps representing that single value. That's the currency code. In case of money, it's the number AND the currency. Think about money in real life: when you hold a ten EUR note, it doesn't tell you that there are cents too.

Completely agreed. But does that mean an object cannot have any further properties than what it is dependent on? The solution in this PR does not change the behaviour of `Currency`. It leaves behaviour untouched. We need new classes for different behaviour (SRP). In this case we only attach properties that are tight to a currency to the object itself. They are optional.

> what should happen if there is no subunit?

Currencies that do not have a subunit will have a subunit of zero. In this PR it is the default value within the currency object. In the case of the PR that creates `CurrenciesWithSubunit` there have to be more objects `ConstantSubunit` or `DefaultSubunit` in order to provide defaults. Or if we have `subunitFor` as method of `Currencies` that method has to be implemented.

>  As for the name: everyone should solve it on their own :trollface: The biggest problem in my opinion is that Name specifically is usually a translated field and I really don't want to provide translations. Every framework handles translations separately.

Also disagree here. Every currency has an official name. But you are right in the fact that localization might translate the official name differently. Therefore people can perfectly create their own `Currencies` implementation, e.g. `ChineseLocalizedISOCurrencies` and attach a different name to the currency.

sagikazarmark:
> But does that mean an object cannot have any further properties than what it is dependent on?

Yes, it does. :smile:  It is not the behaviour, it's the object identity.

> Every currency has an official name.

No. I only accept a currency on my website which does not have a name only a code: `8g7bv87b786cv8zuvb786bv7zv76v` (<- yes, this is the currency code :trollface:)

The point here: this is a money processing library and having an official way to format these values should be the maximum that we provide as a presentation tool. I think we should rather concentrate on this and leave presentation and everything else to consumers, because there is no generic way of doing the presentation layer.

> Therefore people can perfectly create their own Currencies implementation, e.g. ChineseLocalizedISOCurrencies and attach a different name to the currency.

No-no, currency repositories return currencies which does not have special forms. Translation should completely be out of scope here. Currency details on the other hand could be localized.

frederikbosch:
You are right on the `ChineseLocalizedISOCurrencies`, that is a bad idea. Also updated the PR to actually see it without name, but I am not convinced yet. Actually, I think I still disagree.

1. How should I solve displaying ISOCurrencies names then? I will end up creating a new class loading the currencies again. `ISOCurrencies` is a repository that is still not providing anything.

2. Currencies that have no name, can actually have no name. Moreover, [no name is default](https://github.com/moneyphp/money/blob/8601b1f707c9f8fb83a556577fb2690cd32d2760/src/Currency.php#L27). It is the same as is with the subunit default of zero.

sagikazarmark:
This PR goes against Value Objects. Please check the linked PR, I gave several examples when having such fields/features in a currency object is a bad idea, even can cause problems.

> How should I solve displaying ISOCurrencies names then?

In symfony, I would create a service which accepts a currency and I would use the currency code appended to a trans key and use the translator to translate the currency name.

PR #260, opened 2016-08-03
Do not allocate to negative ratio

### LINKED ISSUE #258: Unspecified behavior of Money::allocate()
What is correct behavior for each of these examples?

``` php
/* Negative amount */
(new Money(-5, new Currency('USD')))->allocate([1, 1]); // Now returns [-2, -3]
(new Money(-5, new Currency('USD')))->allocate([3, 7]); // Now returns [-1, -4]
(new Money(-5, new Currency('USD')))->allocate([7, 3]); // Now returns [-3, -2]

/* Negative ratio */
(new Money(5, new Currency('USD')))->allocate([-3, 7]); // Now returns [-3, 8]
(new Money(5, new Currency('USD')))->allocate([-3, -7]); // Now returns [2, 3]
```

I guess that `ratios` mustn't contain negative value, and we should throw an exception if it does.

What about negative amount of money?

frederikbosch:
No allocations allowed to negative values. When having a negative amount, allocation should be reversed. Fixes #258.

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -357,6 +357,14 @@ final class Money implements \JsonSerializable
         $total = array_sum($ratios);
 
         foreach ($ratios as $ratio) {
+            if ($ratio < 0) {
+                throw new \InvalidArgumentException('Cannot allocate to none. Ratio must be zero or positive');
+            }
+
+            if ($this->isNegative()) {
+                $ratio = $total - $ratio;
+            }
+
             $share = $this->getCalculator()->share($this->amount, $ratio, $total);
             $results[] = $this->newInstance($share);
             $remainder = $this->getCalculator()->subtract($remainder, $share);
--- a/spec/MoneySpec.php
+++ b/spec/MoneySpec.php
@@ -437,7 +437,6 @@ class MoneySpec extends ObjectBehavior
                 foreach ($subject as $key => $money) {
                     $compareTo = new Money($value[$key], $money->getCurrency());
                     if ($money->equals($compareTo) === false) {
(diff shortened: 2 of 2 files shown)


src/Money.php
```diff
@@ -357,6 +357,14 @@ public function allocate(array $ratios)
         $total = array_sum($ratios);
 
         foreach ($ratios as $ratio) {
+            if ($ratio < 0) {
```
Sharom: @frederikbosch Any of ratios can be `0`, but not all.

1. All ratios must be non-negative (>= 0);
2. Sum of ratios must be greater than `0` (division by zero occur if sum is 0).

Sharom:
I am not sure about allocation of negative amount. It's question about domain.

Previous behavior may be correct. If we do allocation of debt (represented as negative amount) for example.

```php
(new Money(-5, new Currency('USD')))->allocate([3, 7]); // => [-1, -4]
```

frederikbosch:
@Sharom Hmm. Now I am doubting too.

pamil:
Negative amounts are all right, but mixed sign ratios surely shouldn't be allowed - anyway, all of the allocated money should have the same sign. Negative ratios aren't as bad as mixed ones, but IMO it's the feature we don't need just for clarity.

frederikbosch:
I reverted to the old behaviour. Let's leave as it is. We could always add another constant to the `Money` class, as with rounding.

So negative ratios are not allowed. And when there is a negative amount (debt), the one with largest share gets most of the debt.

PR #263, opened 2016-08-03
throw exception when allocation total is zero

frederikbosch:
@Sharom There was an extreme delay in notifications from github. You are right, total cannot be zero. In this PR it throws an exception.

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -357,7 +357,7 @@ final class Money implements \JsonSerializable
         $results = [];
         $total = array_sum($ratios);
 
-        if ($total > 0) {
+        if ($total <= 0) {
             throw new \InvalidArgumentException('Cannot allocate to none, sum of ratios must be greater than zero');
         }
 
--- a/src/Money.php
+++ b/src/Money.php
@@ -357,7 +357,7 @@ final class Money implements \JsonSerializable
         $results = [];
         $total = array_sum($ratios);
 
-        if ($total === 0) {
+        if ($total > 0) {
             throw new \InvalidArgumentException('Cannot allocate to none, sum of ratios must be greater than zero');
         }
 
--- a/src/Money.php
+++ b/src/Money.php
@@ -350,15 +350,20 @@ final class Money implements \JsonSerializable
     public function allocate(array $ratios)
     {
         if (count($ratios) === 0) {
-            throw new \InvalidArgumentException('Cannot allocate to none');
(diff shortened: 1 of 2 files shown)


src/Money.php
```diff
@@ -350,15 +350,20 @@ public function divide($divisor, $roundingMode = self::ROUND_HALF_UP)
     public function allocate(array $ratios)
     {
         if (count($ratios) === 0) {
-            throw new \InvalidArgumentException('Cannot allocate to none');
+            throw new \InvalidArgumentException('Cannot allocate to none, ratios cannot be an empty array');
```
sandrokeil: @frederikbosch Why not using a [component exception](https://github.com/moneyphp/money/blob/master/src/Exception/Exception.php#L10) here? The `InvalidArgumentException` must be created too.

frederikbosch: @sandrokeil Could you clarify a bit more? Don't get your point. It's the argument that is wrong here, right? It's not a Money exception.

sandrokeil:
Each exception should rely on `Money\Exception\Exception`. So in this case I would prefer this Exception or is there a special reason why to use `\InvalidArgumentException`?

```php
namespace Money\Exception;

use InvalidArgumentException as PhpInvalidArgumentException;

class InvalidArgumentException extends PhpInvalidArgumentException implements Exception
{
}
```

So you can catch all component exceptions `Money\Exception\Exception`, only the component invalid argument exception `Money\Exception\InvalidArgumentException` or all invalid argument exceptions `\InvalidArgumentException`.

sagikazarmark:
Since this is rather a logic problem, not a domain one I see no reason to have a custom exception, but I see no reason against either.

sandrokeil:
Alright, I was just wondering of the mixed usage.

sagikazarmark:
Well, if this would be the only case where we use an SPL exception and not our custom one then maybe it would make sense for consistency.

frederikbosch:
@sandrokeil Well, there were times I also followed your pattern. Meaning every exception from a library should be thrown from within that library. Now I see things different. What is an exceptional state and who is responsible for creating that state?

Only when the library itself turns into an exceptional state, e.g. cannot parse currency because it is not supported by a parser, it must be a library exception. But when user is providing wrong arguments, the exceptional state is the responsibility of the user of the library, then I am fine with SPL exceptions.

sandrokeil:
@frederikbosch Thanks for your explanation. Sounds interesting to me.

Issue #272, opened 2016-09-10
Conversion with different subunits, not working as expected


muteor:
This is for the 3.0 beta.

When converting between two currencies that have different subunits, the conversion is not returning the correct amount.

Here is test case added to `CurrencyPairSpec.php`

```php
function it_converts_correctly_when_currencies_have_different_subunits() {
    $this->beConstructedWith(new Currency('GBP'), new Currency('BHD'), 0.5);

    $money = new Money(100, new Currency('GBP'));

    $resultMoney = $this->convert($money);

    $resultMoney->shouldHaveType(Money::class);
    $resultMoney->getAmount()->shouldBeLike(500);
    $resultMoney->getCurrency()->getCode()->shouldReturn('BHD');
}
```

Running this, `$resultMoney` has value of `50` where it should be `500` because `BHD` has subunit of `1000` not `100`.

This can be fixed by doing `((subunit_to / subunit_from) * amount) * rate` when doing the conversion. So in this example `((1000 / 100) * 100) * 0.5 = 500` which is `500 fils` or `0.5  dinar` (http://www.xe.com/currencyconverter/convert/?Amount=1&From=GBP&To=BHD). 

I had a quick look at the `convert` method but wasn't sure how you would like getting the subunit value into that method, looks like we would need to add a dependency on currencies.

piku235:
I also hit on that some time ago, when testing conversion between EUR (2 minorUnits) and JPY (no minor units). At this moment I'm using a modified version, I highly needed it in one of my projects, generally works. :) I think I'll be able to share with you about my solution pretty soon when I'll finish doing final tests.

frederikbosch:
As soon as Travis in #280 is done, we will merge that PR and this issue will fixed. Thanks for reporting and your help to make this a better package!

PR #275, opened 2016-09-14
Money conversion

piku235:
This pull request solves a bug described in #272. 

To summarize the conversion bug appears only when converting money between currencies that have different minorUnits. For instance when converting 1 EUR (100 amount, 2 minorUnits) to JPY (no minorUnits) with 115 as the conversionRate it'll give 11500 JPY (11500 amount) instead of 115 JPY.
To solve this I used minorUnits that are holded by Currencies instances to calculate difference in minorUnits between a base currency and a target currency. The calculated difference tells about how many places to shift the decimal point. The whole idea shows the bellow alghorithm:

```
places = targetMinorUnits - baseMinorUnits
number = amount * conversionRate
number = shiftDecimalPoint(number, places)
result = round(number)
```


sagikazarmark:
This is nice, thanks @piku235. I will review it during the weekend.

.gitignore
```diff
@@ -1,3 +1,4 @@
+/.idea
```
sagikazarmark: Please add this to your global gitignore

spec/MovePointExamples.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace spec\Money;
+
+/**
+ * Mocking with typehints usage won't work here as the trait is autoloaded.
+ *
+ * @see https://github.com/phpspec/phpspec/issues/825
+ */
+trait MovePointExamples
```
sagikazarmark: MovePoint doesn't sound declarative enough. MovingDecimalPoint maybe? Or MovingDecimalPlaces?

spec/NumberSpec.php
```diff
@@ -61,6 +63,33 @@ function it_tests_if_the_number_is_integer($number, $decimal)
...
     }
 
+    /**
+     * @dataProvider movePointExamples
+     */
+    function it_moves_point($number, $places, $expected, $integerPart, $fractionalPart)
+    {
+        $this->beConstructedWith($number);
+
+        $this->movePoint($places);
```
sagikazarmark: Move point again. Move decimal places? Not sure, but sounds better for me.

src/Calculator.php
```diff
@@ -114,4 +114,19 @@ public function round($number, $roundingMode);
...
+     *
+     * @param string $number
+     * @param int    $places Number of places to move the
+     *                       decimal point. Negative value
+     *                       moves the decimal point to the left
+     *                       and positive in the opposite direction
+     *
+     * @return string
+     */
+    public function movePoint($number, $places);
```
sagikazarmark: Is it a good idea not to accept a Number instance here? /cc @frederikbosch

src/Money.php
```diff
@@ -36,6 +38,11 @@
     private static $calculator;
 
     /**
+     * @var Currencies
+     */
+    private static $currencies;
```
sagikazarmark: I am not really sure that this is a good idea. But I don't see a better option. Alternative: remove conversion from the Money object and move it to a Converter object.

/cc @frederikbosch

.gitignore
piku235: Thanks for the hint.

spec/MovePointExamples.php
piku235: Ok, I think the MovingDecimalPointExamples it's a good name, thanks.

spec/NumberSpec.php
piku235: At the beggining I couldn't find a good name for it, so I just borrowed it from movePointLeft method of the BigDecimal class (Java). If it's also confusing for others then of course I think we can change it to `moveDecimalPlaces` or something more clear.

src/Calculator.php
piku235: hmm.. what we'll gain by doing so? Looking at the `Currency` interface arguments of each method are of int|float|string type.

src/Money.php
piku235: Yes, initially I was thinking the same, the logic behind conversion for sure will get bigger and maybe it'll be better to move it, but after looking a while on the Money class and seeing calculator instances are also holded by static property I decided to go the samy way to keep the class organization.

muteor:
I am not sure about the optional currencies dependency, I think it would be better to make currency know about subunits, I see it as currency is actually invalid without its subunit because conversion (and other operations) does not work without it, therefore you should not be able construct a currency without specifying its subunit.

```php
$currency = new Currency('USD', 2); // code, minorUnit
```

Might also be easier to store the `subunit_to_unit` like ruby money does: https://github.com/RubyMoney/money/blob/master/config/currency_iso.json

Another approach could be to make `CurrencyPair` depend on `Currencies` this way the conversion could be correct, though I think this still leaves the design open to improper conversions.

TDLR: We should make it impossible to do incorrect conversions.

frederikbosch:
I also feel that a static property is the wrong decision. If you ask me there three options.

1. Money knows about subunits
2. Currency knows about subunits
3. Conversion should take place in a new class

The case of conversion with different subunits is quite similar to the case of the same currency with different subunits (think of gasstations). How to solve that conversion?

That leads me to say that subunits should be part of the Money class. So option 1. Because then you solve both cases. Think about it. The integer of 100 is useless if you do not know the currency and the unit of measurement.

frederikbosch:
@piku235 Thanks for all your effort here. Unfortunately we solved issue in another way. Without your PR however, we would never came up with this solution and the required thinking. I hope you do not mind us solving differently. Moreover, hope you want to contribute again in the future. Thanks again!

PR #277, opened 2016-09-16
Fix integer detection when the number overflows the integer type

### LINKED ISSUE #276: Number::isInteger can return false on numbers greater PHP_INT_MAX
Hi,

we've noticed some Strange behavior when dealing with numbers greater than PHP_INT_MAX in the version v3.0.0-beta.2.

It's easy to reproduce, install version 3.x beta:
`composer require mathiasverraes/money ^3.0-beta`

An run this code:
```php
<?php

namespace TestError;

use Money\Money;
use Money\Number;

require_once 'vendor/autoload.php';

echo "PHP_INT_MAX: ", PHP_INT_MAX, PHP_EOL;
$chunkSize = strlen((string)PHP_INT_MAX)-1;
echo PHP_EOL;

echo "Building Number (Chunk Size ", $chunkSize, "): ", PHP_EOL;
$bogous = str_repeat('1', $chunkSize);
$bogous .= '0123123';

echo "Bogous Number: ", $bogous, PHP_EOL;
echo '$bogous Number::isInteger():', PHP_EOL;
var_dump(Number::isInteger($bogous));

// Will throw InvalidArgumentException: Amount must be an integer
$willThrow = Money::EUR($bogous);
```

I've tracked the problem down to the `Number::isInteger` method, [which splits numbers bigger PHP_INT_MAX into chunks to validate them individually](https://github.com/moneyphp/money/blob/fb5644d929285cbb54c572e436eb1acd77bb5479/src/Number.php#L172-L177).
If one of this chunks has a leading zero the `filer_var` will return false.

The code above results in two chunks:
- '111111111111111111'
- '0123123'

The seconds chunk is an invalid integer.
Every operation (add, multiply, etc..) can result in a number triggering this problem.

Any ideas on how to work around this problem?
- a regex `preg_match` with `^\d+$`
- using `ctype_digit` (requires php to be compiled with ctype support)

sagikazarmark:
When the number overflows integer type it's split into chunks
and checked step by step. When the chunk starts with a zero
the type check fails since an integer cannot start with zero.

Closes #276

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+
+## Unreleased
+
+### Fixed
+
+- Integer detection when the number overflows the integer type and contains zeros
+
+
 ## 3.0.0-beta.2 - 2016-08-03
 
 ### Changed
--- a/spec/NumberSpec.php
+++ b/spec/NumberSpec.php
@@ -53,6 +53,14 @@ class NumberSpec extends ObjectBehavior
         $this->getIntegerRoundingMultiplier()->shouldReturn($negative ? '-1' : '1');
     }
 
+    /**
+     * @dataProvider numberExamples
+     */
+    function it_tests_if_the_number_is_integer($number, $decimal)
+    {
+        $this->isInteger($number)->shouldReturn(!$decimal);
+    }
+
     public function numberExamples()
     {
         return [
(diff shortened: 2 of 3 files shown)


sagikazarmark:
@frederikbosch  can you please take a look?

frederikbosch:
Will do next week, on holidays now

frederikbosch:
Uh, in two weeks

sagikazarmark:
It's only code review, if you have a few minutes. Enjoy your holiday!

frederikbosch:
@sagikazarmark Can be merged if you ask me! Very nice found! Those zero bugs are harder than we think!

PR #280, opened 2016-10-01
Fix converting currencies with different subunits

frederikbosch:
I chose the method of a new class `Converter`. Also renamed `getCurrencyPair` to `quote`, because I feel that is better name. Don't ask but tell. Happy to change it again if that is what we want.

### DIFF
--- a/src/Converter.php
+++ b/src/Converter.php
@@ -21,6 +21,7 @@ final class Converter
      * @param Money        $money
      * @param CurrencyPair $currencyPair
      * @param int          $roundingMode
+     *
      * @return Money
      */
     public function convert(Money $money, CurrencyPair $currencyPair, $roundingMode = Money::ROUND_HALF_UP)
--- a/src/Converter.php
+++ b/src/Converter.php
@@ -3,7 +3,7 @@
 namespace Money;
 
 /**
- * Provides a way to convert Money to Money in another Currency using an exchange rate
+ * Provides a way to convert Money to Money in another Currency using an exchange rate.
  *
  * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
  */
--- a/src/Converter.php
+++ b/src/Converter.php
@@ -38,7 +38,7 @@ final class Converter
     public function convert(Money $money, Currency $counterCurrency, $roundingMode = Money::ROUND_HALF_UP)
     {
         $baseCurrency = $money->getCurrency();
-        $ratio = $this->exchange->quote($baseCurrency, $counterCurrency);
+        $ratio = $this->exchange->quote($baseCurrency, $counterCurrency)->getConversionRatio();
 
         $baseCurrencySubunit = $this->currencies->subunitFor($baseCurrency);
         $counterCurrencySubunit = $this->currencies->subunitFor($counterCurrency);
--- a/src/Converter.php
+++ b/src/Converter.php
@@ -46,9 +46,7 @@ final class Converter
 
         if ($subunitDifference > 0) {
             $ratio = $ratio / pow(10, $subunitDifference);
-        }
-
-        if ($subunitDifference < 0) {
+        } elseif ($subunitDifference < 0) {
             $ratio = $ratio * pow(10, abs($subunitDifference));
         }
 
--- a/src/Converter.php
+++ b/src/Converter.php
@@ -5,7 +5,7 @@ namespace Money;
 /**
  * Provides a way to convert Money to Money in another Currency using an exchange rate.
  *
- * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
+ * @author Frederik Bosch <f.bosch@genkgo.nl>
  */
 final class Converter
 {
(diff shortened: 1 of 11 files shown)


sagikazarmark:
Without actually checking this: if you are curious about terminology, the latest changes in Swap can be a good base. It is actually two library now:

https://github.com/florianv/swap
https://github.com/florianv/exchanger

src/Converter.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace Money;
+
+final class Converter
```
sagikazarmark: Missing author and possible description about this class.

src/Converter.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * @param Money        $money
+     * @param CurrencyPair $currencyPair
+     * @param int          $roundingMode
+     *
+     * @return Money
+     */
+    public function convert(Money $money, CurrencyPair $currencyPair, $roundingMode = Money::ROUND_HALF_UP)
+    {
+        if ($money->getCurrency()->getCode() !== $currencyPair->getBaseCurrency()->getCode()) {
```
sagikazarmark: Maybe use the currency `equals` method? It's kind of Currency domain logic after all.

src/Converter.php
```diff
@@ -0,0 +1,50 @@
...
+            throw new \InvalidArgumentException('Base Currency of currency pair does not equal money currency');
+        }
+
+        $subunitBaseCurrency = $this->currencies->subunitFor($currencyPair->getBaseCurrency());
```
sagikazarmark: `$baseCurrencySubunit` sounds better IMO, the current one is an inverse logic which is harder to understand for me.

src/Converter.php
```diff
@@ -0,0 +1,50 @@
...
+        $subunitCounterCurrency = $this->currencies->subunitFor($currencyPair->getCounterCurrency());
```
sagikazarmark: Same here: `$counterCurrencySubunit`

src/Converter.php
```diff
@@ -0,0 +1,50 @@
...
+        $differenceInSubunit = $subunitBaseCurrency - $subunitCounterCurrency;
```
sagikazarmark: `$subunitDifference`?

src/Converter.php
```diff
@@ -0,0 +1,50 @@
...
+        $ratio = $currencyPair->getConversionRatio();
+
+        if ($differenceInSubunit > 0) {
```
sagikazarmark: The following seems to be the same instead of two ifs:

``` php
        if (0 !== $differenceInSubunit) {
            $ratio = $ratio * pow(10, abs($differenceInSubunit));
        }
```

Even if there is a need to have two separate checks: the difference cannot be less than and more than 0 at the same time, so I would use elseif to avoid evaluating both conditions if not necessary. But if the above works, I would rather use that.

sagikazarmark:
I am thinking about a third wrapper class: Accept a money and a currency only and don't care about internals (exchange, converter). It might be an overkill, but from a consuming point of view: I might not always be interested in currency pair details.

sagikazarmark:
Just for the record: does this fix #275 ?

frederikbosch:
Yes, will update your suggestions tomorrow!

sagikazarmark:
Actually I rather meant a wrapper around Converter than altering it's usecase. I see a valid use case for using the converter with an already existing currency pair.

frederikbosch:
@sagikazarmark I liked your suggestion regarding accepting money and currency only. However, I think we do not need a third class for it. Better to change the `Converter` to your suggested signature. So now there is a `Converter` to change `Money` and `Exchange` for those interesed in `CurrencyPair`.

src/Converter.php
sagikazarmark: Maybe I wasn't clear: I would prefer having the solution above as it's essentially the same, but simpler. Is there any specific use case why we need both branches?

sagikazarmark:
And what about those who want to use a fixed conversion rate? They have to create an Exchange for a static list this way.

frederikbosch:
@sagikazarmark Hmm, true that.

frederikbosch:
@sagikazarmark But why is that a problem? You can implement `Exchange`, right? That is what it is for?

frederikbosch:
This way you have a perfect separation of responsibilities. The fact that someone will be using a fixed exchange rate should be expressed in an Exchange related class, not in the converter. The converter should depend on the exchange.

sagikazarmark:
Okay, it's not a big deal after all. In that case, should we provide a default implementation for a static Exchange list?

frederikbosch:
I think that would be a good idea. Hang on, I will make it part of the PR.

frederikbosch:
Great, I think this made the PR better!

frederikbosch:
@sagikazarmark Merge after Travis also likes it?

frederikbosch:
Hmm docs.

frederikbosch:
Now it is done. Will merge after Travis is happy.

PR #281, opened 2016-10-02
Add USD, EUR and GPB as static methods to the PHPDoc block

hohl:
Documenting dynamically appended static methods helps IDEs (like PHPStorm) to understand that these static methods are valid methods (for autocompletion and error detection). I'd at least add the most common currencies to the PHPDoc header for the `Money` class.

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -7,9 +7,9 @@ namespace Money;
  *
  * @author Mathias Verraes
  *
- * @method static \Money\Money USD()
- * @method static \Money\Money EUR()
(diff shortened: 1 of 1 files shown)


sagikazarmark:
I am sorry, but I won't accept this patch in it's current form. This has already been rejected two times before, which means to me that it's commonly requested feature. However, this library tries to be a generic Money lib which does not emphasize ISO currencies.

Although I don't like this solution, this issue has been raised quite a few times now, so I opened an issue to sort it out: #282

PR #284, opened 2016-10-04
Improve ratio compensation

sagikazarmark:
Multiplication with a negative power is the same as division with a positive power.


sagikazarmark:
Although I recommended this in the original PR, it has been missed. Can you please check @frederikbosch ?

frederikbosch:
Sorry, you are right, can be merged!

PR #285, opened 2016-10-04
Remove conversion from CurrencyPair

sagikazarmark:


frederikbosch:
Good call @sagikazarmark. Obviously I was not aware of these methods. But I should have looked better.

frederikbosch:
Will merge after Travis is happy.

Issue #278, opened 2016-09-29
Something is wrong with the compare method in Calculator


wppd:
when doing `divide`, it'll call the calculator to verify the divisor **(cast to string first)** isn't zero.

from what I see,  the compare method in BcMathCalculator doesn't have scale as third parameter.

`bccomp($a, $b)`

which makes `bccomp("0.5", 0) === 0`: "Returns 0 if the two operands are equal"

if use GmpCalculator instead,

`ErrorException: gmp_cmp(): Unable to convert variable to GMP - string is not an integer`

frederikbosch:
Why would you want that in a `Money` context? It is [impossible to construct a `Money` object with a decimal](https://github.com/moneyphp/money/blob/master/src/Money.php#L55).

frederikbosch:
The same story is true for almost any `gmp_` method. You cannot `substract` or `add` a decimal number with the current GMP calculator. But since adding and subtracting decimals is not required by this package either (for the same reason I already gave that it is impossible to construct a Money object with a decimal), there will not be support for it until we start working on `PreciseMoney`. See #7.

frederikbosch:
Also relevant in this thread is the [GMP Floating Point Support RFC](https://wiki.php.net/rfc/gmp-floating-point).

sagikazarmark:
We could probably throw a logic exception when doing operations with
decimals.

Frederik Bosch <notifications@github.com> ezt írta (2016. október 1.,
szombat):

frederikbosch:
@sagikazarmark I think we should just leave it. If there is no `Money` problem, there is no problem. We already have insufficient hands to get 3.0 out in reasonable time. If we starting fixing these issues...

sagikazarmark:
Well, I agree. Honestly, I am not sure I completely understood the problem. However, if there is a leak in the API which easily leads to errors like this AND fixing them requires an API change we should do something.

Again: not sure if this is the case.

PR #287, opened 2016-10-06
gmp: fix multiplier with leading zero

### LINKED ISSUE #286: GmpCalculator multiply:  gmp_init() error when fractional part with leading 0
```PHP
$phpCalculator = new \Money\Calculator\PhpCalculator;
$gmpCalculator = new \Money\Calculator\GmpCalculator;
$bcCalculator = new \Money\Calculator\BcMathCalculator;
echo $phpCalculator->multiply('100', 0.029) . "\n"; // 2.9
echo $gmpCalculator->multiply('100', 0.029) . "\n"; // .0 gmp_init(): Unable to convert variable to GMP - string is not an integer
echo $bcCalculator->multiply('100', 0.029) . "\n"; // 2.900
```

frederikbosch:
Fixes #286

### DIFF
--- a/spec/Calculator/CalculatorBehavior.php
+++ b/spec/Calculator/CalculatorBehavior.php
@@ -41,6 +41,13 @@ trait CalculatorBehavior
     {
         $this->multiply(1, 1.5)->shouldReturn('1.5');
         $this->multiply(10, 1.2500)->shouldBeLike('12.50');
+        $this->multiply(100, 0.29)->shouldBeLike('29');
+        $this->multiply(100, 0.029)->shouldBeLike('2.9');
+        $this->multiply(100, 0.0029)->shouldBeLike('0.29');
+        $this->multiply(1000, 0.29)->shouldBeLike('290');
+        $this->multiply(1000, 0.029)->shouldBeLike('29');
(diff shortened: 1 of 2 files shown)


sagikazarmark:
So is it actually something which can be fixed? Not a limitation of GMP, like in the referenced issue?

frederikbosch:
@sagikazarmark In the case of issue #278, there was an example that could never be the case inside `Money`. Why would you want to compare non-integers? That is simply impossible with Money because Money requires the amount to be integer. We never compare non-integers.

But issue #286 is different. One should be able to multiply an amount with decimal leading zero value. As you can see in the multiply method I created a workaround the limitation of `GMP`.

My point with the calculators is: they are Money internals. Anything inside the calculators that causes bugs inside Money should be fixed. Every other case should be ignored.

frederikbosch:
@sagikazarmark Decimals within `Money` are only relevant when multiplying or with divisions. That is why I only created the workaround inside the GMP calculator for `multiply` and `divide`. Decimals are irrelevant when adding, subtracting and comparing because `Money` will always require integers for those methods.

sagikazarmark:
@frederikbosch okay, agreed.

Issue #78, opened 2015-01-06
Is there a way to get the float number?


herlon214:
This lib has some function to get the float number? Example: 1270 get -> 12.70 ?

Thanks in advance, and sorry for the newbie question.

mathiasverraes:
getAmount() / 100

teohhanhui:
Now that Money::getAmount returns a string, we can just do

```php
substr_replace($orderSubtotalMoney->getAmount(), '.', '-2', 0)
```

to obtain the actual amount as string (say, perhaps you want to store the value as "decimal"/"numeric" type in the database).

sagikazarmark:
Thanks for the tip @teohhanhui

teohhanhui:
Shouldn't we add this as a convenience method?

sagikazarmark:
@teohhanhui I would prefer keeping money clear from convenience stuff.

PR #289, opened 2016-10-07
Add DecimalMoneyFormatter

teohhanhui:
This formatter outputs a simple decimal string which is always in a consistent format independent of locale.

Related: #78

The workaround in https://github.com/moneyphp/money/issues/78#issuecomment-202262198 is inadequate now that we have subunits.

@sagikazarmark I hope this is not just "convenience stuff"?

### DIFF
--- a/doc/Formatting.rst
+++ b/doc/Formatting.rst
@@ -10,7 +10,7 @@ Money comes with the following implementations out of the box:
 Intl Formatter
 --------------
 
-As it's name says, this formatter requires the `intl` extension and uses ``NumberFormatter``. In order to provide the
+As its name says, this formatter requires the `intl` extension and uses ``NumberFormatter``. In order to provide the
 correct subunit for the specific currency, you should also provide the specific currency repository.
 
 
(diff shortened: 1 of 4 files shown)


src/Formatter/DecimalStringMoneyFormatter.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $this->currencies = $currencies;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function format(Money $money)
+    {
+        $valueBase = (string) $money->getAmount();
```
sagikazarmark: Amount is already string AFAIK.

teohhanhui:
There's duplicated code, because it's just `IntlMoneyFormatter` without for the last step.

frederikbosch:
I have nothing to add here. The PR looks good. Maybe the name of class. I think we could drop String from the name. That results in `DecimalMoneyFormatter` which I think covers the purpose of the class.

sagikazarmark:
Alrighty, up to @frederikbosch then

frederikbosch:
@teohhanhui There is nothing wrong with that code duplication. There is no need for abstraction.

frederikbosch:
Oh, I did agree with @sagikazarmark comment though. Amount is already string. So please change that. I only wanted to say that I have no more comments on the PR than that.

src/Formatter/DecimalStringMoneyFormatter.php
```diff
@@ -0,0 +1,58 @@
...
+use Money\Currencies;
+use Money\Money;
+use Money\MoneyFormatter;
+
+/**
+ * Formats a Money object as a decimal string.
+ *
+ * @author Teoh Han Hui <teohhanhui@gmail.com>
+ */
+final class DecimalStringMoneyFormatter implements MoneyFormatter
```
frederikbosch: Change in `DecimalMoneyFormatter`.

src/Formatter/DecimalStringMoneyFormatter.php
frederikbosch: Agree here.

frederikbosch:
Will merge after Travis is happy.

frederikbosch:
@teohhanhui Great PR, thanks for your help!

bendavies:
before this was merged, i solved it like this:

```
        $numberFormatter = new \NumberFormatter('en_US', \NumberFormatter::PATTERN_DECIMAL, '0.00');
        $decimalFormatter = new IntlMoneyFormatter($numberFormatter, $currencies);
```

frederikbosch:
Not a bad solution either.

bendavies:
yeah, the duplicated code (not a bad thing i know), could probably be removed and the `DecimalMoneyFormatter` could be composed with the `IntlMoneyFormatter`

teohhanhui:
DecimalMoneyFormatter has no dependency on the `intl` component. Can't say the same about IntlMoneyFormatter...

bendavies:
good point

PR #290, opened 2016-10-07
Speed up Travis build

teohhanhui:


.travis.yml
```diff
@@ -8,7 +8,8 @@ addons:
 
 cache:
     directories:
-        - $HOME/.composer/cache/files
+        - $HOME/.cache/pip
+        - $HOME/.composer/cache
```
sagikazarmark: This is not recommended. It might cause some inconsistencies in composer as it does not update the repository information

.travis.yml
```diff
@@ -31,7 +32,7 @@ matrix:
           env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest --ignore-platform-reqs" COVERAGE=true TEST_COMMAND="composer test-ci"
 
 before_install:
-    - travis_retry composer self-update
+    - if [[ ! $COVERAGE = 'true' && ! $TRAVIS_PHP_VERSION = hhvm* ]]; then phpenv config-rm xdebug.ini || echo "xdebug not available"; fi
```
sagikazarmark: Please use this as a boilerplate: https://github.com/php-http/boilerplate/blob/master/.travis.yml#L31-L36

.travis.yml
sagikazarmark: Please put back the original

teohhanhui: Any reference that it might cause inconsistencies?

teohhanhui: Found it: https://github.com/travis-ci/travis-ci/issues/4579

sagikazarmark: Not actual inconsistencies, but rather delays in updated repository information. I got the intel from @xabbuh originally, I don't have any references. The relevand PR: https://github.com/php-http/boilerplate/pull/19

teohhanhui: https://github.com/travis-ci/travis-ci/issues/4579#issuecomment-138548147

> The repo subfolder in the cache contains metadata downloaded from Packagist (and other repos you may have registered in the project), and these would probably have changed for each build (the main files giving access to others are updated every few minutes because of the frequency of changes in packages). I'm not sure the time won thanks to the caching of some of the metadata files (the ones for packages not updated since your previous Travis build) would actually outweight the time spent storing the Travis cache again for each build.

sagikazarmark: Cool, thanks for that.

PR #292, opened 2016-10-11
Add DecimalMoneyParser

teohhanhui:
This parser takes a raw decimal string in a locale-independent format.

Replaces `StringToUnitsParser`.


frederikbosch:
This looks good. I wonder (do not see a test for it): what happens if I parse $ 1.00 like so.

```php
$parser->parse('$ 1.00', 'USD');
```

teohhanhui:
It will throw exception. I'll add the test for that.

frederikbosch:
Okay, so when people want to parse an amount with a currency symbol they will have to use the `IntlMoneyParser`. Or they create a decorator within their own domain that removes any special character from the decimal amount. Right?

Could you highlight what is breaking of this parser comparing to the `StringToUnitsParser`?
1. Only a minus sign is allowed in the `DecimalMoneyParser` while both plus and minus were allowed in the `StringToUnitsParser`.
2. ....

We might want to consider to have both `StringToUnitsParser` and `DecimalMoneyParser` in our library, depending on our purpose. Could you highlight why the former has been removed instead of having both?

teohhanhui:
Differences from `StringToUnitsParser`:
1. No plus sign allowed.
2. Digits (before the decimal point) is compulsory.
3. Comma is not allowed as the decimal separator. Only the dot (`.`) is allowed.
4. Also fixed a bug where `0.`, `1.`, `1234.` were considered valid.

The guiding principle for me is to eliminate all ambiguities. There is no longer the question of whether this alternate format or that shorthand format is supported, because there is only one simple, standard format.

Why remove `StringToUnitsParser`? Because it is redundant and only adds confusion as to which parser to use. If the user needs more control, it's possible to use `IntlMoneyParser` by providing a `\NumberFormatter` which uses a custom pattern. Or implement a custom parser.

frederikbosch:
@sagikazarmark I think the points are valid. However it does mean we will break BC again. WDYT?

sagikazarmark:
@frederikbosch I also think the points are valid, and avoiding confusion should be high prio. Also, I am more concerned about break in behaviour than breaks like this. We are not stable or RC yet, so we can freely change the API. That said we can just deprecate it, but starting a stable with deprecations is not a good sign.

frederikbosch:
Then I am going to merge this.

frederikbosch:
@teohhanhui Many thanks again. Great work!

PR #296, opened 2016-10-18
WIP: Refactor tests

sagikazarmark:
It was probably a mistake using PHPSpec for everything. In this PR I am trying to separate behavior from unit.


frederikbosch:
Agreed, I would it when you're happy with the changes.

frederikbosch:
+merge

sagikazarmark:
Thanks, still WIP and uncovered the GMP division issue which I would like to get done first.

frederikbosch:
I will pickup the GMP problem. If you could do this one, I am not so experienced with phpspec.

sagikazarmark:
Well, I already finalized the tests on this branch, so you can create one from it to solve the issue.

frederikbosch:
@sagikazarmark Done, but I need to refactor. It is quite hard to be honest. I added also comments to this method with some explanations. Since `GMP` does not support floating point precision, we will never be able to give precision on anything resulting in a decimal. But the amount of precision (like in BC math) is something we should be able to control.

frederikbosch:
At this moment there is a precision of 1, which will make sure that the resulting integer will be rounded correctly.

sagikazarmark:
Hm, okay. It might be a problem for large numbers then where we cannot make sure that the precision is correct and in those cases, Money is responsible for rounding. But for smaller numbers we might be able to calculate the fractional part in advance, right?

If so than I think we can just state that this is for internal usage and we cannot guarantee precision 100% and we just do what we can.

frederikbosch:
@sagikazarmark There will not be a problem with `GMP` in the current `Money` context since there will always be a correct amount. But, if we move on to `PreciseMoney` this might be troublesome because we cannot give enough precision on the fractional part.

frederikbosch:
But I will have another look at this. I think I can still improve a lot.

sagikazarmark:
Well, we have to state that in the documentation somewhere.

frederikbosch:
@sagikazarmark  That might not even be necessary. We don't do that for BC math, are we? We can inject a precision integer to the GMP constructor, just like we are doing with BC math.

frederikbosch:
Done. I want to have a second look at this another time. Since it is quite mathematical, I should review the approach and the resulting code.

frederikbosch:
Moreover, we have had some issues with leading and trailing zero in the past. I should be careful and make sure that such problems might not be introduced again.

frederikbosch:
Okay, I am happy now with the result. It is simplified and logic now.

1. If the divisor is decimal, multiply amount and divisor with 10 * (length of divisor fractional part)
2. Calculate the integer part of the result using `gmp_div_qr`
3. Multiply the remainder part - if any - with the injected precision
4. Do another division calculation with the multiplied remainder
5. Concatenate the result of 2 and 4 with keeping in my the precision used at 4.

Example: 2 / 0.5 = 4
1. Divisor is decimal with fractional length of 1, multiply both with 10, results in 20 and 5
2. 20 / 5 = 4
3. No remainder, return 4

Example: 98 / 24 = 4.083333333333333
1. Divisor is not decimal, no need for multiplication
2. 98 / 24 = 4 with a remainder of 2
3. Multiply the remainder with 10 ^ 14 (default scale 14)
4. Fractional part: 8,333333333×10¹²
5. Result: 4.083333333333333

frederikbosch:
I'd say we merge it in. Of course, when Travis is also happy with it.

frederikbosch:
Maybe you can squash your commits too @sagikazarmark

sagikazarmark:
Yes, I will squash, also test refactoring is not ready yet.

PR #300, opened 2016-10-19
throw for currencies that is not bitcoin

### LINKED ISSUE #295: Bitcoin currency repository should throw an exception when currency is unkown

frederikbosch:
Fixes #295

### DIFF
--- a/src/Currencies/BitcoinCurrencies.php
+++ b/src/Currencies/BitcoinCurrencies.php
@@ -4,6 +4,7 @@ namespace Money\Currencies;
 
 use Money\Currencies;
 use Money\Currency;
+use Money\Exception\UnknownCurrencyException;
 
 /**
  * @author Frederik Bosch <f.bosch@genkgo.nl>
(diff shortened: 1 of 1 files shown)


src/Currencies/BitcoinCurrencies.php
```diff
@@ -26,6 +27,10 @@ public function contains(Currency $currency)
      */
     public function subunitFor(Currency $currency)
     {
+        if ($currency->getCode() !== self::CODE) {
+            throw new UnknownCurrencyException($currency->getCode().' cannot be found within BitcoinCurrencies');
```
sagikazarmark: I would rather say "XXX is not bitcoin and is not supported by this currency repository"

frederikbosch:
Ok, will push that directly to master.

src/Currencies/BitcoinCurrencies.php
frederikbosch: Done.

PR #307, opened 2016-10-20
Update Swap

### LINKED ISSUE #298: Update Swap

sagikazarmark:
Closes #298

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,13 +27,15 @@
         "ext-bcmath": "*",
         "ext-gmp": "*",
         "ext-intl": "*",
-        "florianv/swap": "^2.3",
+        "florianv/swap": "^3.0",
         "psr/cache": "^1.0",
         "cache/taggable-cache": "^0.4.0",
         "phpspec/phpspec": "^2.5",
         "henrikbjorn/phpspec-code-coverage": "^2.0.2",
         "moneyphp/iso-currencies": "~1.0.0",
-        "sllh/php-cs-fixer-styleci-bridge": "^2.1"
+        "sllh/php-cs-fixer-styleci-bridge": "^2.1",
+        "php-http/message": "^1.4",
+        "php-http/mock-client": "^0.3.3"
     },
     "suggest": {
         "ext-bcmath": "Calculate without integer limits",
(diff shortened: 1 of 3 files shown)


sagikazarmark:
I wonder if we should rather have a static factory for ExchangeRateProvider instead of the current solution.

@florianv your feedback is very welcome.

florianv:
Thanks @sagikazarmark for taking initiative on that!

This looks ok to me, nothing to add 👍

sagikazarmark:
Thanks for the feedback @florianv

composer.json
```diff
@@ -42,6 +44,9 @@
         "florianv/swap": "Exchange rates library for PHP",
         "psr/cache-implementation": "Used for Currency caching"
     },
+    "conflict": {
+        "florianv/swap": "<3.0"
```
pamil: Is it really needed? It's impossible to install this package with `florianv/swap<3.0`.

src/Exchange/SwapExchange.php
```diff
@@ -18,28 +18,38 @@
...
 {
     /**
-     * @var SwapInterface
+     * @var Swap
      */
     private $swap;
 
     /**
-     * @param SwapInterface $swap
+     * @param Swap|ExchangeRateProvider $exchange
```
pamil: If it's `SwapExchange` that uses `Swap`, shouldn't we make `ExchangerExchange` that uses `ExchangeRateProvider`?

src/Exchange/SwapExchange.php
```diff
@@ -18,28 +18,38 @@
...
+        if (!$exchange instanceof Swap && !$exchange instanceof ExchangeRateProvider) {
+            throw new \InvalidArgumentException(sprintf(
+                'Exchange must either be %s or %s',
+                Swap::class,
+                ExchangeRateProvider::class
+            ));
+        }
+
+        if ($exchange instanceof ExchangeRateProvider) {
+            $exchange = new Swap($exchange);
```
pamil: Anyway, what about just not accepting `ExchangeRateProvider` here, it seems easy to make `Swap` instance out of it and will reduce complexity.

src/Exchange/SwapExchange.php
```diff
@@ -18,28 +18,38 @@
...
+        }
+
+        $this->swap = $exchange;
```
pamil: `$exchange` -> `$swap`?

composer.json
pamil: Haven't noticed it's in `require-dev`, so in fact it is possible to install it with lower Swap version, but as you said, it's optional dependency so there is no sense to restrict it.

PR #320, opened 2016-10-26
Update README.md

### LINKED ISSUE #318: Remove integration section from README
Since this will be a new version, very likely to be incompatible with those (need to check), also we will have our own money bundle.

@frederikbosch WDYT?

frederikbosch:
Fixes #318 and adds some words on testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,13 +47,11 @@ Features
 - Currency repositories (ISO currencies included)
 - Money exchange (including Swap implementation)
 
-Integration
------------
-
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Sorry, I am already doing this and a bit more which will conflict

sagikazarmark:
Will integrate the test part

frederikbosch:
Cool.

frederikbosch:
Maybe add the logo in the README too. Now there is only one location where it is visible.

sagikazarmark:
Good idea

PR #326, opened 2016-11-06
Add exchange for retrying with reversed currencies

### LINKED ISSUE #323: Reversible Conversion
Wouldn't it make sense if the currency conversion is reversible?

`$exchange = new FixedExchange([
    'EUR' => [
        'USD' => 1.25
    ]
]);`
`$converter = new Converter(new ISOCurrencies(), $exchange);`

`$eur100 = Money::EUR(100);`
`$usd125 = $converter->convert($eur100, new Currency('USD'));`

`$eur100Original = $converter->convert($usd125, new Currency('EUR'));`
`assert($eur100Original->equals($eur100));`

sagikazarmark:
Closes #323 

(TODO: Maybe we need a better name?)

### DIFF
--- a/doc/spelling_word_list.txt
+++ b/doc/spelling_word_list.txt
@@ -15,3 +15,5 @@ Hieatt
 Mee
 USD
 greaterThanOrEqual
+
+behaviour
--- a/spec/Exchange/ReversedCurrenciesExchangeSpec.php
+++ b/spec/Exchange/ReversedCurrenciesExchangeSpec.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace spec\Money\Exchange;
+
+use Money\Currency;
+use Money\CurrencyPair;
+use Money\Exception\UnresolvableCurrencyPairException;
+use Money\Exchange;
+use Money\Exchange\ReversedCurrenciesExchange;
+use PhpSpec\ObjectBehavior;
+
+final class ReversedCurrenciesExchangeSpec extends ObjectBehavior
+{
+    function let(Exchange $exchange)
+    {
+        $this->beConstructedWith($exchange);
+    }
+
+    function it_is_initializable()
+    {
+        $this->shouldHaveType(ReversedCurrenciesExchange::class);
+    }
+
+    function it_is_an_exchange()
+    {
+        $this->shouldImplement(Exchange::class);
+    }
+
+    function it_exchanges_currencies(Exchange $exchange)
+    {
+        $baseCurrency = new Currency('EUR');
+        $counterCurrency = new Currency('USD');
+        $currencyPair = new CurrencyPair($baseCurrency, $counterCurrency, 1.25);
+
+        $exchange->quote($baseCurrency, $counterCurrency)->willReturn($currencyPair);
+
+        $this->quote($baseCurrency, $counterCurrency)->shouldreturn($currencyPair);
+    }
+
+    function it_exchanges_reversed_currencies_when_the_original_pair_is_not_found(Exchange $exchange)
+    {
+        $baseCurrency = new Currency('USD');
+        $counterCurrency = new Currency('EUR');
+        $conversionRatio = 1.25;
+        $currencyPair = new CurrencyPair($counterCurrency, $baseCurrency, $conversionRatio);
+
+        $exchange->quote($baseCurrency, $counterCurrency)->willThrow(UnresolvableCurrencyPairException::class);
+        $exchange->quote($counterCurrency, $baseCurrency)->willReturn($currencyPair);
+
+        $currencyPair = $this->quote($baseCurrency, $counterCurrency);
+
+        $currencyPair->shouldHaveType(CurrencyPair::class);
+        $currencyPair->getBaseCurrency()->shouldReturn($baseCurrency);
+        $currencyPair->getCounterCurrency()->shouldReturn($counterCurrency);
+        $currencyPair->getConversionRatio()->shouldReturn(1 / $conversionRatio);
+    }
+
+    function it_throws_an_exception_when_neither_the_original_nor_the_reversed_currency_pair_can_be_resolved(Exchange $exchange)
+    {
+        $baseCurrency = new Currency('USD');
+        $counterCurrency = new Currency('EUR');
+
+        // Exceptions are not matched based on identity, but instance and properties
+        $exchange->quote($baseCurrency, $counterCurrency)->willThrow(UnresolvableCurrencyPairException::createFromCurrencies($baseCurrency, $counterCurrency));
+        $exchange->quote($counterCurrency, $baseCurrency)->willThrow(UnresolvableCurrencyPairException::createFromCurrencies($counterCurrency, $baseCurrency));
+
+        $this->shouldThrow(UnresolvableCurrencyPairException::createFromCurrencies($baseCurrency, $counterCurrency))->duringQuote($baseCurrency, $counterCurrency);
+    }
+}
--- a/src/Exchange/ReversedCurrenciesExchange.php
+++ b/src/Exchange/ReversedCurrenciesExchange.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace Money\Exchange;
+
+use Money\Currency;
+use Money\CurrencyPair;
+use Money\Exception\UnresolvableCurrencyPairException;
+use Money\Exchange;
+
+/**
+ * Tries the reverse of the currency pair if one is not available.
+ *
+ * Note: adding nested ReversedCurrenciesExchange could cause a huge performance hit.
+ *
+ * @author Márk Sági-Kazár <mark.sagikazar@gmail.com>
+ */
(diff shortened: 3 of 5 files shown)


frederikbosch:
Like it, merge it in! Regarding your note: I think there is no performance hit if you nest the ReversedCurrenciesExchange. That class is not doing anything special performance wise. When nesting the quote method returns immediately and the result will only pass more stations that return as well.

sagikazarmark:
> I think there is no performance hit if you nest the ReversedCurrenciesExchange

Well, I think there is indirectly. Let's say you use swap which sends an HTTP request for each quote call. If you nest two ReversedCurrenciesExchange and the neither the currency pair nor it's reverse cannot be found then it will send 4 HTTP requests: two for the first reverse, two for the second, and this increases exponentially.

But I would say this is a human error. Nevertheless, should be mentioned in the docs.

sagikazarmark:
@frederikbosch added some docs, can you please give some feedback?

spec/Exchange/ReversedCurrenciesExchangeSpec.php
```diff
@@ -0,0 +1,70 @@
...
+        $currencyPair->getCounterCurrency()->shouldReturn($counterCurrency);
+        $currencyPair->getConversionRatio()->shouldReturn(1 / $conversionRatio);
+    }
+
+    function it_throws_an_exception_when_neither_the_original_nor_the_reversed_currency_pair_can_be_resolved(Exchange $exchange)
+    {
+        $baseCurrency = new Currency('USD');
+        $counterCurrency = new Currency('EUR');
+
+        $e = new UnresolvableCurrencyPairException();
```
pamil: The exceptions are compared by class and then by their properties, we can use `UnresolvableCurrencyPairException::class` instead of `$e` with the same outcome.

spec/Exchange/ReversedCurrenciesExchangeSpec.php
```diff
@@ -0,0 +1,70 @@
...
+namespace spec\Money\Exchange;
+
+use Money\Currency;
+use Money\CurrencyPair;
+use Money\Exception\UnresolvableCurrencyPairException;
+use Money\Exchange;
+use Money\Exchange\ReversedCurrenciesExchange;
+use PhpSpec\ObjectBehavior;
+
+class ReversedCurrenciesExchangeSpec extends ObjectBehavior
```
pamil: All other specs are final.

src/Exchange/ReversedCurrenciesExchange.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function quote(Currency $baseCurrency, Currency $counterCurrency)
+    {
+        try {
+            return $this->exchange->quote($baseCurrency, $counterCurrency);
+        } catch (UnresolvableCurrencyPairException $e) {
```
pamil: `$exception`?

src/Exchange/ReversedCurrenciesExchange.php
```diff
@@ -0,0 +1,50 @@
...
+            try {
+                $currencyPair = $this->exchange->quote($counterCurrency, $baseCurrency);
+
+                return new CurrencyPair($baseCurrency, $counterCurrency, 1 / $currencyPair->getConversionRatio());
+            } catch (UnresolvableCurrencyPairException $ie) {
```
pamil: `$inversedException`?

doc/spelling_word_list.txt
```diff
@@ -15,3 +15,5 @@ Hieatt
 Mee
 USD
 greaterThanOrEqual
+
+behaviour
```
pamil: What is the purpose of this file? Shouldn't it be gitignored?

spec/Exchange/ReversedCurrenciesExchangeSpec.php
sagikazarmark: Nice catch, thanks.

spec/Exchange/ReversedCurrenciesExchangeSpec.php
sagikazarmark: Actually I wanted to test here that the first exception is rethrown. I am not 100% sure about it, but it seemed to be right to throw the exception of the original currency pair and as such, it should be an expected behaviour => tested.

src/Exchange/ReversedCurrenciesExchange.php
sagikazarmark: Using `$e` seems to be a common pattern. Do you suggest to use `$exception` in this specific case because of the next exception? I would like to be consistent about this.

doc/spelling_word_list.txt
sagikazarmark: No, it contains a list of words that should be ignored by the spell checker.

src/Exchange/ReversedCurrenciesExchange.php
sagikazarmark: All right, sounds ok to me.

spec/Exchange/ReversedCurrenciesExchangeSpec.php
pamil: I think that's the correct behaviour too, but this spec does not assert it.

The following code:
```php
$exchange->quote($baseCurrency, $counterCurrency)->willThrow(UnresolvableCurrencyPair::createFromCurrencies($baseCurrency, $counterCurrency));
$exchange->quote($counterCurrency, $baseCurrency)->willThrow(UnresolvableCurrencyPair::createFromCurrencies($counterCurrency, $baseCurrency));

$this->shouldThrow(UnresolvableCurrencyPair::createFromCurrencies($baseCurrency, $counterCurrency))->duringQuote($baseCurrency, $counterCurrency);
```

Should assert it (though I would double-check it by trying to make this spec failing by expecting the second exception).

sagikazarmark: I don't see how this is better TBH. The current version makes the stub throw a specific instance and we expect to get it back. Testing the exception context is out of scope here IMO, as it is not part of the behaviour of this class, but part of any other exchanges instead.

What you suggest would require to know exception's internal state which is - again - out of scope IMO.

Do you disagree?

pamil: It's pretty obvious that `UnresolvableCurrencyPairException` thrown during `quote($baseCurrency, $counterCurrency)` will have the same base / counter currencies. We can think of an exception as a Value Object.

This way we can assert that the first exception is rethrown, the current spec implementation does not assert it.

sagikazarmark: > This way we can assert that the first exception is rethrown, the current spec implementation does not assert it. 

Why? The same instance is checked, although VO equality is not, but I don't think that responsibility belongs here. We just rethrow an exception that we receive from a delegated object. The exception context should be tested in the tests of that delegated object IMO.

sagikazarmark: If it belonged here we would have to construct a new exception in this class, or test it's content and throw some other exception if the thrown one has improper currencies in it. Otherwise the fault of another component could cause fault in this class.

pamil: That's the implementation of `shouldThrow` matcher in PhpSpec. If you take a look at `ThrowMatcher::verifyPositive()` you can see that the exceptions aren't compared by identity, but by their classes (`$exceptionThrown instanceof $exception`) and then by their properites (excluding `file`, `line`, `string`, `trace` and `previous`).

The exception `$e` is `new UnresolvableCurrencyPairException()`.
The exception created by `willThrow(UnresolvableCurrencyPairException::class)` is handled by `ThrowPromise` and results in `new UnresolvableCurrencyPairException()` too.
They are equal in terms of `ThrowMatcher` as their `message`, `code`, `baseCurrency` and `counterCurrency` properties are equal.

To sum it up, the spec will still pass even if the second exception was rethrown.

pamil: You're right, the exception creation might not belong here, but without it, it's not possible to assert that the first exception is rethrown but not the second one.

Anyway, I think we got too deep in this problem and the real solution here is to create a new exception in that case, which will explicitly say that it couldn't resolve both (base currency / counter currency) pair and (counter currency / base currency) pair.

sagikazarmark: Ah, I see, I though they are matched by identity. Indeed, your solution would is the right one then.

> real solution here is to create a brand new exception

Hm, that wouldn't make the decoration completely transparent and actually from a public API point of view the only important thing is that there was no match for our original currency pair.

frederikbosch:
@sagikazarmark Good job! I think you have proposed a good solution like this. Good idea to create a decorator for it. Merge it in.

PR #335, opened 2016-11-29
Precise money

### LINKED ISSUE #7: Deal with decimals using PreciseMoney
idea from http://joda-money.sourceforge.net/userguide.html

> There are two main classes which are both based on BigDecimal:
> Money - an amount in a single currency where the decimal places is determined by the currency
> BigMoney - an amount in a single currency where there is no restriction on the scale
> 
> For example, the currencies of US dollars, Euros and British Pounds all use 2 decimal places, thus a Money instance for any of these currencies will have 2 decimal places. By contrast, the Japanese Yen has 0 decimal places, and thus any Money instance with that currency will have 0 decimal places.
> 
> Instances of BigMoney may have any scale that can be supported by BigDecimal.
> Conversion between a Money and a BigMoney can be performed using the methods toBigMoney() and toMoney(). The latter optionally takes a rounding mode to handle the case where information must be lost.

frederikbosch:
Adds `PreciseMoney`. Fixes #7.

### DIFF
--- a/spec/Calculator/CalculatorBehavior.php
+++ b/spec/Calculator/CalculatorBehavior.php
@@ -40,7 +40,7 @@ trait CalculatorBehavior
 
     function it_divides_a_value_by_another()
     {
-        $this->divide(rand(-100, 100), rand(-100, 100))->shouldBeString();
+        $this->divide(rand(-100, 100), rand(1, 100))->shouldBeString();
     }
 
     function it_ceils_a_value()
--- a/spec/PreciseMoneySpec.php
+++ b/spec/PreciseMoneySpec.php
@@ -1,0 +1,339 @@
+<?php
+
+namespace spec\Money;
+
+use Money\Calculator;
+use Money\Currency;
+use Money\PreciseMoney;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+
+final class PreciseMoneySpec extends ObjectBehavior
+{
+    const AMOUNT = 10;
+    const OTHER_AMOUNT = 5;
+    const CURRENCY = 'EUR';
+    const OTHER_CURRENCY = 'USD';
+
+    function let(Calculator $calculator)
+    {
+        // Override the calculator for testing
+        $reflection = new \ReflectionProperty(PreciseMoney::class, 'calculator');
+        $reflection->setAccessible(true);
+        $reflection->setValue(null, $calculator->getWrappedObject());
+
+        $this->beConstructedWith(self::AMOUNT, new Currency(self::CURRENCY));
+    }
+
+    function it_is_initializable()
+    {
+        $this->shouldHaveType(PreciseMoney::class);
+    }
+
+    function it_is_json_serializable()
+    {
+        $this->shouldImplement(\JsonSerializable::class);
+    }
+
+    function it_has_an_amount()
+    {
+        $this->getAmount()->shouldBeLike(self::AMOUNT);
+    }
+
+    function it_has_a_currency()
+    {
+        $currency = $this->getCurrency();
+
+        $currency->shouldHaveType(Currency::class);
+        $currency->equals(new Currency(self::CURRENCY))->shouldReturn(true);
+    }
+
+    function it_throws_an_exception_when_amount_is_not_numeric()
+    {
+        $this->beConstructedWith('ONE', new Currency(self::CURRENCY));
+
+        $this->shouldThrow(\InvalidArgumentException::class)->duringInstantiation();
+    }
+
+    function it_constructs_integer()
+    {
+        $this->beConstructedWith(5, new Currency(self::CURRENCY));
+    }
+
+    function it_constructs_string()
+    {
+        $this->beConstructedWith('5', new Currency(self::CURRENCY));
+    }
+
+    function it_constructs_integer_with_decimals_of_zero()
+    {
+        $this->beConstructedWith('5.00', new Currency(self::CURRENCY));
+    }
+
+    function it_tests_currency_equality()
+    {
+        $this->isSameCurrency(new PreciseMoney(self::AMOUNT, new Currency(self::CURRENCY)))->shouldReturn(true);
+        $this->isSameCurrency(new PreciseMoney(self::AMOUNT, new Currency(self::OTHER_CURRENCY)))->shouldReturn(false);
+    }
+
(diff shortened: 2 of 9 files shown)


frederikbosch:
Few considerations before continuing:

1. Is this the right path to start with?
2. What to do with the PHP Calculator? It will have strange results when multiplying and dividing, so I left it out of the `PreciseMoney` Calculators. However, I did implement some changes for tests to succeed (all floating number bug related changes).

What should be next?
1. What to do with formatting and parsing? Do we want to support it at all?
2. What to do with conversion? Do we want to support it at all?
3. ...

m4tthumphrey:
Is this still active? I presume I am right in thinking that this is to allow us to perform operations using more decimal places? Ie `GBP 1.1535 * 0.4251` etc?

frederikbosch:
@m4tthumphrey You can already do that.

```php
new Money(11535, new Currency('GBP'));
new Money(4251, new Currency('GBP'));
```

You should implement your own currency with subunits set to 4. See the [BitcoinCurrencies](https://github.com/moneyphp/money/blob/master/src/Currencies/BitcoinCurrencies.php) how to do this. If there are more currencies involved than only the pound, use the [Aggregate Currencies class](http://moneyphp.org/en/latest/features/currencies.html#aggregate-currencies) to turn multiple currencies it into a single repository.

`PreciseMoney` could be used if you do not want to round your amounts. But as you can see, this branch has not been merged yet. Use at your own risk.

m4tthumphrey:
I was more enquiring than actually looking to use it! 

I have just looked into the custom currencies and I see how it works now! Thanks!

gmponos:
Hello,

What's the status of this? Is it gonna be merged?

frederikbosch:
No progress. I expect to look at in the end of the year. If someone could give some feedback, that could - of course - help to have the feature earlier.

frederikbosch:
@sagikazarmark I am planning to update this PR. The considerations I mentioned earlier on are still valid.

Is this the right path? Meaning: should we create a *new* class `PreciseMoney` that supports unlimited decimals? Or should we embed the functionality in `Money`? The current solution is the former. However it has cons. First of all, we cannot reuse the formatters/parsers/currencies within PreciseMoney. But that is - of course - only a con if the domain requires the functionality. And there is a way of overcoming this by creating conversion methods: `Money` <-> `PreciseMoney`.

So the questions are?

1. Create a new class `PreciseMoney` or embed in `Money`?
2. If answer is create a new class `PreciseMoney`, how to support formatters/parsers/currencies?
3. Should we create a `toMoney(int $decimals)` in `PreciseMoney`?
4. Should we create a `toMoney(Currencies $currency)` method in `PreciseMoney`?
5. Or should there be something like `PreciseToMoneyConverter`?
6. Or should be add methods to `Money`, like `fromPreciseMoney()`?
7. What if `PreciseMoney` was not be currencies aware, so `PreciseMoney = f(amount)` instead of `PreciseMoney = f(amount, currency)`?

frederikbosch:
My initial answers are as follows.

1. Create a new class `PreciseMoney` or embed in `Money`?
> Yes
2. If answer is create a new class `PreciseMoney`, how to support formatters/parsers/currencies?
> Convert PreciseMoney to Money (one direction).
3. Should we create a `toMoney(int $decimals)` in `PreciseMoney`?
> Yes
4. Should we create a `toMoney(Currencies $currency)` method in `PreciseMoney`?
> No
5. Or should there be something like `PreciseToMoneyConverter`?
> No
6. Or should be add methods to `Money`, like `fromPreciseMoney()`?
> No, we should add `fromMoney(Money $money)` and `toMoney(int $decimals): Money` to `PreciseMoney` only.
7. What if `PreciseMoney` was not be currencies aware, so `PreciseMoney = f(amount)` instead of `PreciseMoney = f(amount, currency)`?
> Bad idea.

frederikbosch:
@sagikazarmark Could you emphasize your thoughts. Maybe just start with some initials thoughts. Then we will iterate towards the optimal solution.

willemstuursma:
Hi @frederikbosch. 

Here at Mollie we think adding a `PreciseMoney` type is a great idea, but you should not be able to convert from `PreciseMoney` to `Money`.

frederikbosch:
@willemstuursma Cool to see you drop in. Could you emphasize why we should not be able to convert to `Money`? How to support formatters/parsers/currencies for `PreciseMoney` without conversion possibility?

willemstuursma:
@frederikbosch Why don't you drop by our office next week and we can discuss it with out team - maybe a few beers and we can see how Mollie can help bring this to fruition.

frederikbosch:
@willemstuursma Great idea. However, I am on honeymoon leave from Tuesday. So, it will have to wait until November.

willemstuursma:
Oh congrats! Well, send me an email when you're back and enjoy your honeymoon together!

sagikazarmark:
@frederikbosch congrats! Let's talk about this later!

pcbulldozer:
Hi all. Not sure where to see if there is any activity on this (and the other release) items?
Looking for a Money library around which to wrap a series of financial calculators and models - this one works well, bar the precision aspects. I'd have thought the quickest (perhaps not the most elegant) way to get there would have been to allow for instance-specific registration of a calculator (since the calculators can already be created with higher precision) and then to expose a NO_ROUNDING mode on the multiply and divide methods. I see @frederikbosch comments above - I'd probably have gone the other way and expand the functionality of Money, rather than add PreciseMoney to the mix, but I think I can see his motivation. 
Either way, great library, great potential, so hoping these items are on the cards for the near future!

willemstuursma:
@pcbulldozer in the mean time, you can also create high precision currencies and use those.

pcbulldozer:
thanks @willemstuursma will be giving this some thought today! appreciate the response.

arubacao:
@willemstuursma Did you take a look and mind sharing your solution?

nwehrhan:
Just checking in on this thread. I would love to see this PR come to fruition. Not sure what the current status on this is? Did you guys ( @frederikbosch and @willemstuursma ) come to a solution/agreement on how to implement this? Is there something I could do to help? I see there are currently conflicts, but not sure if this idea is abandoned.

BonBonSlick:
So, what do other do with high fraction 8+ numbers?

nottobysnot (approved):

PR #336, opened 2016-12-08
Correct Allocation docs.

MichaelGooden:
Just a small correction to the allocation documentation.


frederikbosch:
Good catch, thanks

PR #337, opened 2016-12-21
Fix Formatter trailing decimal point when subunit is 0

elazer:


frederikbosch:
@elazer could you add tests for this?

elazer:
There is a test for this already, but for some reason the tests passed even though it should have failed.

$this->assertEquals(1, '1.'); does not fail.

Any ideas why?

frederikbosch:
@elazer assertEquals means ==, so I guess 1. is interpreted as a float which == 1 as an int. The test should be changed to assertSame, meaning ===. Then it will fail I guess. The end result should be assertSame('1', $formattedMoney).

frederikbosch:
@sagikazarmark Why is this failing? Dont get the sphinx problem.

frederikbosch:
@sagikazarmark Thanks for fixing!

frederikbosch:
@elazer Thanks for the PR

PR #339, opened 2016-12-27
Fix integer validation when chunk started with a dash

### LINKED ISSUE #338: [Bug] validating integers
Hi! I just looked to the source code https://github.com/moneyphp/money/blob/master/src/Number.php#L171 and there seems to be a bug:
```php
function validateNumberAsInteger($number)
{
    // Check if number is invalid because of integer overflow
    $invalid = array_filter(
        str_split($number, strlen((string) PHP_INT_MAX) - 1),
        function ($chunk) {
            // Leading zeros should not invalidate the chunk
            $chunk = ltrim($chunk, '0');
            // Allow chunks containing zeros only
            return '' !== $chunk && false === filter_var($chunk, FILTER_VALIDATE_INT);
        }
    );
    return count($invalid) === 0;
}

assert(!validateNumberAsInteger('123456789012345678-123456'));  // should throw something
```
Can anyone verify and/if fix? Thanks.

sagikazarmark:
Closes #338

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 
 - Reversed Currencies Exchange to try resolving reverse of a currency pair
 
+### Fixed
+
+- Passing integer validation when chunk started with a dash
+
 
 ## 3.0.0 - 2016-10-26
 
--- a/src/Number.php
+++ b/src/Number.php
@@ -170,6 +170,16 @@ final class Number
      */
     private static function validateNumberAsInteger($number)
     {
+        // Number allowed to be a negative one
+        if ('' !== $number && '-' === $number[0]) {
+            $number = substr($number, 1);
+        }
+
+        // Leading zeros are not allowed
+        if (strlen($number) > 1 && '0' === $number[0]) {
+            return false;
+        }
+
         // Check if number is invalid because of integer overflow
         $invalid = array_filter(
             str_split($number, strlen((string) PHP_INT_MAX) - 1),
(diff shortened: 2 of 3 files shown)


frederikbosch:
Good catch and fix.

wimby:
Thanks @sagikazarmark for fixing this so quick 👍  
one little thing pls: `validateNumberAsInteger('---123')` should return false too.

frederikbosch:
@wimby I think it does with his PR. Why wouldn't it?

wimby:
actually, there seems to be still bugs for numbers `-0`, `+123` and even `123456789012345678+13456`

@frederikbosch https://3v4l.org/j6QUv

wimby:
how about `preg_match('/^-?[1-9]\d*|0$/', $input)` IMHO its more readable and faster (no need for creating new arrays)

frederikbosch:
Regular expressions are generally very slow.

sagikazarmark:
Round two pushed. I would say -0 is kind of okayish, as the real number might be -0.5

I also added tests for multiple leading zeros before a number.

wimby:
I guess ctype_digit is LC_CTYPE locale specific, I don't think you want to support [all these](http://www.fileformat.info/info/unicode/category/Nd/list.htm) characters:
https://forums.phpfreaks.com/topic/157880-solved-you-guys-ever-use-ctype-for-validation/?p=832933

wimby:
why wouldn't you just use https://github.com/moneyphp/money/pull/341? And if you don't want to support plus sign, remove it from signs...

frederikbosch:
See #341.

PR #341, opened 2016-12-27
Fix integer validation when chunk started with a dash

frederikbosch:
@sagikazarmark @wimby Created this alternative solution. Positive and negative signs are allowed. but only when they are the first character.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 
 - Reversed Currencies Exchange to try resolving reverse of a currency pair
 
+### Fixed
+
+- Passing integer validation when chunk started with a dash
+
 
 ## 3.0.0 - 2016-10-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,13 @@
 ### Added
 
 - Reversed Currencies Exchange to try resolving reverse of a currency pair
+- Documentation on allowed integer(ish) values when constructing Money
 
 ### Fixed
 
 - Passing integer validation when chunk started with a dash
+- Passing integer validation when the fractional part started with a dash
+- Money constructed from a string with fractional zeroes equals to a Money constructed without the fractional part (eg. `'5.00'` and `'5'`)
 
 
 ## 3.0.0 - 2016-10-26
--- a/doc/getting-started.rst
+++ b/doc/getting-started.rst
@@ -19,7 +19,8 @@ See :doc:`features/parsing` for additional ways to instantiate a Money object fr
 
 Accepted integer values
 -----------------------
-The Money object only supports integer(ish) values on instantiation. The following is (not) supported.
+The Money object only supports integer(ish) values on instantiation. The following is (not) supported. When a
+non-supported value is passed a `\InvalidArgumentException` will be thrown.
 
 .. code-block:: php
 
@@ -35,13 +36,13 @@ The Money object only supports integer(ish) values on instantiation. The followi
     // string is accepted if fractional part is zero
     $fiver = new Money('500.00', new Currency('USD'));
 
-    // leading zero's are accepted
+    // leading zero's are not accepted
     $fiver = new Money('00500', new Currency('USD'));
 
-    // multiple zero's is not accepted, throws \InvalidArgumentException
+    // multiple zero's are not accepted
     $fiver = new Money('000', new Currency('USD'));
 
-    // plus sign is not accepted, throws \InvalidArgumentException
+    // plus sign is not accepted
     $fiver = new Money('+500', new Currency('USD'));
 
 
--- a/doc/getting-started.rst
+++ b/doc/getting-started.rst
@@ -38,7 +38,7 @@ The Money object only supports integer(ish) values on instantiation. The followi
     // leading zero's are accepted
     $fiver = new Money('00500', new Currency('USD'));
 
-    // multiple zero's is accepted
+    // multiple zero's is not accepted, throws \InvalidArgumentException
     $fiver = new Money('000', new Currency('USD'));
 
     // plus sign is not accepted, throws \InvalidArgumentException
--- a/doc/getting-started.rst
+++ b/doc/getting-started.rst
@@ -17,6 +17,35 @@ All amounts are represented in the smallest unit (eg. cents), so USD 5.00 is wri
 
 See :doc:`features/parsing` for additional ways to instantiate a Money object from strings.
 
+Accepted integer values
+-----------------------
+The Money object only supports integer(ish) values on instantiation. The following is (not) supported.
+
+.. code-block:: php
+
+    use Money\Currency;
+    use Money\Money;
+
+    // int is accepted
+    $fiver = new Money(500, new Currency('USD'));
+
+    // string is accepted if integer
+    $fiver = new Money('500', new Currency('USD'));
+
+    // string is accepted if fractional part is zero
+    $fiver = new Money('500.00', new Currency('USD'));
+
+    // leading zero's are accepted
+    $fiver = new Money('00500', new Currency('USD'));
+
+    // multiple zero's is accepted
+    $fiver = new Money('000', new Currency('USD'));
+
+    // plus sign is not accepted, throws \InvalidArgumentException
+    $fiver = new Money('+500', new Currency('USD'));
+
+
+
 Installation
 ------------
 
--- a/src/Money.php
+++ b/src/Money.php
@@ -56,8 +56,13 @@ final class Money implements \JsonSerializable
      */
     public function __construct($amount, Currency $currency)
     {
-        if (filter_var($amount, FILTER_VALIDATE_INT) === false && !Number::fromString($amount)->isInteger()) {
-            throw new \InvalidArgumentException('Amount must be an integer(ish) value');
+        if (filter_var($amount, FILTER_VALIDATE_INT) === false) {
+            $numberFromString = Number::fromString($amount);
+            if (!$numberFromString->isInteger()) {
+                throw new \InvalidArgumentException('Amount must be an integer(ish) value');
+            }
+
+            $amount = $numberFromString->getIntegerPart();
         }
 
         $this->amount = (string) $amount;
--- a/src/Number.php
+++ b/src/Number.php
@@ -191,7 +191,7 @@ final class Number
             $digit = $number[$position];
             if (!isset(static::$numbers[$digit]) && !($position === 0 && $digit === '-')) {
                 throw new \InvalidArgumentException(
-                    'Invalid integer part '.$number.'. Invalid digit '.$digit . ' found'
+                    'Invalid integer part '.$number.'. Invalid digit '.$digit.' found'
                 );
             }
 
@@ -217,7 +217,7 @@ final class Number
             $digit = $number[$position];
             if (!isset(static::$numbers[$digit])) {
                 throw new \InvalidArgumentException(
-                    'Invalid fractional part '.$number.'. Invalid digit '.$digit . ' found'
+                    'Invalid fractional part '.$number.'. Invalid digit '.$digit.' found'
                 );
             }
 
--- a/src/Number.php
+++ b/src/Number.php
@@ -37,13 +37,13 @@ final class Number
     {
         if ($integerPart !== '' && $this->validateNumberAsInteger($integerPart) === false) {
             throw new \InvalidArgumentException(
-                'Invalid number, integer part ' . $integerPart . ' is not an integer'
+                'Invalid number, integer part '.$integerPart.' is not an integer'
             );
         }
 
         if ($fractionalPart !== '' && $this->validateNumberAsInteger($fractionalPart) === false) {
             throw new \InvalidArgumentException(
-                'Invalid number, fractional part ' . $integerPart . ' is not an integer'
+                'Invalid number, fractional part '.$integerPart.' is not an integer'
             );
         }
 
--- a/src/Number.php
+++ b/src/Number.php
@@ -195,7 +195,7 @@ final class Number
                 return false;
             }
 
-            $position++;
+            ++$position;
         }
 
         return true;
--- a/src/Number.php
+++ b/src/Number.php
@@ -27,7 +27,7 @@ final class Number
     /**
      * @var array
      */
-    private static $signs = ['-' => 1, '+' => 1];
+    private static $signs = ['-' => 1];
 
     /**
      * @param string $integerPart
@@ -35,13 +35,13 @@ final class Number
      */
     public function __construct($integerPart, $fractionalPart = '')
     {
-        if ($integerPart !== '' && $this->validateNumberAsInteger($integerPart) === false) {
+        if ($integerPart !== '' && $this->validateAsIntegerPart($integerPart) === false) {
             throw new \InvalidArgumentException(
                 'Invalid number, integer part '.$integerPart.' is not an integer'
             );
         }
 
-        if ($fractionalPart !== '' && $this->validateNumberAsInteger($fractionalPart) === false) {
+        if ($fractionalPart !== '' && $this->validateAsFractionalPart($fractionalPart) === false) {
             throw new \InvalidArgumentException(
                 'Invalid number, fractional part '.$integerPart.' is not an integer'
             );
@@ -49,6 +49,8 @@ final class Number
 
         if ($integerPart === '-') {
             $integerPart = '-0';
+        } else {
+            $integerPart = ltrim($integerPart, '0');
         }
 
         $this->integerPart = $integerPart ? $integerPart : '0';
@@ -186,12 +188,32 @@ final class Number
      *
      * @return bool
      */
-    private static function validateNumberAsInteger($number)
+    private static function validateAsIntegerPart($number)
     {
         $position = 0;
         while (isset($number[$position])) {
             $digit = $number[$position];
             if (!isset(static::$numbers[$digit]) && !($position === 0 && isset(static::$signs[$digit]))) {
+                return false;
+            }
+
+            ++$position;
+        }
+
+        return true;
+    }
+
+    /**
+     * @param string $number
+     *
+     * @return bool
+     */
+    private static function validateAsFractionalPart($number)
+    {
+        $position = 0;
+        while (isset($number[$position])) {
+            $digit = $number[$position];
+            if (!isset(static::$numbers[$digit])) {
                 return false;
             }
 
--- a/src/Number.php
+++ b/src/Number.php
@@ -188,11 +188,14 @@ final class Number
      */
     private static function validateNumberAsInteger($number)
     {
-        $digits = str_split($number);
-        foreach ($digits as $position => $digit) {
+        $position = 0;
+        while (isset($number[$position])) {
+            $digit = $number[$position];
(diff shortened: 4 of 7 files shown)


frederikbosch:
Also tested '.5' and '-.5'.

frederikbosch:
@wimby I changed `str_split` into a while loop using string offsets. That must be faster.

wimby:
@frederikbosch Confirmed.

Interesting is that preg_match seems to be 2x slower in php7.0 than php5.6, creating arrays is about 2x faster... for preg_match to be faster than arrays would require numbers 2k+ digits.

Cycle is definitely a better way.

wimby:
@frederikbosch same questions here:
should `000`, `+0`, `-0` and `+123` be allowed?
https://3v4l.org/0J2ij

frederikbosch:
Yes, see my tests. PHP allows them too.

wimby:
I can't see `000` and `+0` in tests... `000` returns false for FILTER_VALIDATE_INT

frederikbosch:
https://3v4l.org/tqkZ1

According to Wikipedia "In common numeral notation (which is used in arithmetic and elsewhere), the sign of a number is often denoted by placing a plus sign or a minus sign before the number. For example, +3 denotes "positive three", and −3 denotes "negative three". When no plus or minus sign is given, the default interpretation is that a number is positive. Because of this notation, as well as the definition of negative numbers through subtraction, the minus sign is perceived to have a strong association with negative numbers (of the negative sign). Likewise, "+" associates with positivity."

So, I'd say that closes +0 as to be allowed. Don't have the time now to look into 000.

frederikbosch:
Again Wikipedia "A leading zero is any 0 digit that comes before the first nonzero digit in a number string in positional notation." suggests indeed that 000 should not be allowed because of the lacking first non-zero digit.

wimby:
I don't think you should look into what PHP accepts as a token T_LNUMBER:
- 0123 is read as octal representation of 83 in decimal
- 0x123 is read as hexadecimal (291 in decimal)
both are allowed as integers in PHP parser but not with FILTER_VALIDATE_INT.

I would hold with wikipedia and FILTER_VALIDATE_INT, which disallow anything starting with 0 except only 0. In Math usually does not make sense to have positive and negative zero (+0 and -0), but FILTER_VALIDATE_INT allows it, so what. Same with plus sign. 

If I understand it correctly, what is allowed here goes to `Calculator` which HAS to handle it (gmp and bcmath can handle plus sign but behave differently with leading zero).

frederikbosch:
Number is used in calculators too. Mostly to get consistent output. Especially when using scale to multiply and divide with BC math.

@wimby Agree, let's only support 0, -0 and +0. So 000 should not be supported. I will fix that. And I will document valid inputs.

@sagikazarmark Do you agree; or have any objection on this?

frederikbosch:
I also think we should use assertSame here instead of assertEquals. Will update that too (and fix some minor test-related bugs).

sagikazarmark:
I agree that -0 should be supported, but plus sign shouldn't IMO. Please see my updated solution, it seems to be simpler, not sure about the performance though, but if it's negligible, I would go with the simpler way.

frederikbosch:
@wimby @sagikazarmark I fixed some more bugs. '-100.-100' was accepted. This has been fixed. I also documented accepted values and changed `assertEquals` to `assertSame`.

@wimby I did accept '000'. Simply because it involves more code to not accept it. I do not find it a problem to accept it. If you have strong arguments against this, please let me know. I might reconsider.

@sagikazarmark Could you review this PR? I think it provides a better solution than we had before. It validates the integer part differently than the fractional part (fractional part cannot have a sign). It should also be faster than the previous solution. Finally, I took your advise and made the plus sign to be not accepted.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 
 - Reversed Currencies Exchange to try resolving reverse of a currency pair
 
+### Fixed
+
+- Passing integer validation when chunk started with a dash
```
sagikazarmark: As far as I can see you fixed quite a number of bugs with this PR, so maybe we should list them or write them in a more general way.

doc/getting-started.rst
```diff
@@ -17,6 +17,35 @@ All amounts are represented in the smallest unit (eg. cents), so USD 5.00 is wri
...
+    $fiver = new Money('500', new Currency('USD'));
+
+    // string is accepted if fractional part is zero
+    $fiver = new Money('500.00', new Currency('USD'));
+
+    // leading zero's are accepted
+    $fiver = new Money('00500', new Currency('USD'));
+
+    // multiple zero's is accepted
+    $fiver = new Money('000', new Currency('USD'));
```
sagikazarmark: Is this equal to 0? Should `(new Money('000', new Currency('USD')))->equals(new Money('0', new Currency('USD'))) === true`?

src/Number.php
```diff
@@ -20,17 +20,37 @@
...
+            throw new \InvalidArgumentException(
+                'Invalid number, integer part '.$integerPart.' is not an integer'
+            );
         }
 
-        if ($fractionalPart !== '' && $this->validateNumberAsInteger($fractionalPart) === false) {
-            throw new \InvalidArgumentException('Invalid number');
+        if ($fractionalPart !== '' && $this->validateAsFractionalPart($fractionalPart) === false) {
+            throw new \InvalidArgumentException(
+                'Invalid number, fractional part '.$integerPart.' is not an integer'
```
sagikazarmark: Guess it should be `$fractionalPart`

src/Number.php
```diff
@@ -20,17 +20,37 @@
...
+            );
+        }
+
+        if ($integerPart === '-') {
```
sagikazarmark: Some comment about what's happening here? Also, Yoda pls for variables.

src/Number.php
```diff
@@ -20,17 +20,37 @@
...
+            $integerPart = '-0';
+        } else {
+            $integerPart = ltrim($integerPart, '0');
```
sagikazarmark: Note that this removes all leading zeros. Can't recall why this was an issue for me.

src/Number.php
```diff
@@ -168,21 +188,39 @@ public function getIntegerRoundingMultiplier()
...
+    private static function validateAsIntegerPart($number)
     {
-        // Check if number is invalid because of integer overflow
-        $invalid = array_filter(
-            str_split($number, strlen((string) PHP_INT_MAX) - 1),
-            function ($chunk) {
-                // Leading zeros should not invalidate the chunk
-                $chunk = ltrim($chunk, '0');
+        $position = 0;
+        while (isset($number[$position])) {
```
sagikazarmark: Wouldn't this be better with a for/foreach loop, since we use the counter variable (and the value too)?

src/Number.php
```diff
@@ -168,21 +188,39 @@ public function getIntegerRoundingMultiplier()
...
+            $digit = $number[$position];
+            if (!isset(static::$numbers[$digit]) && !($position === 0 && isset(static::$signs[$digit]))) {
```
sagikazarmark: We really have one sign supported, do we need an array for that?

CHANGELOG.md
frederikbosch: Done.

doc/getting-started.rst
frederikbosch: Changed into not accepted. I think that is better,

src/Number.php
frederikbosch: Done.

src/Number.php
frederikbosch: We are supporting `.5` in the current version due to our current number validation. With `.5` the integer part is empty, str_split results in an empty array, so `count($invalid) === 0` becomes true. Hence, we support `.5`.

So, because I do not want BC break, I continue supporting this. But then we have to be consequent. This means `.5` and `-.5` are now supported. Hence, `-` is a valid integer part and should result in `-0`. We might want to undo this in release where we do BC again.

src/Number.php
frederikbosch: We should decide now what to do with it. Afterwards, it becomes a feature (or not). I have nothing against, but I also see no reason why to support it.

src/Number.php
frederikbosch: Nope, then we need a function call (`strlen`). This should make it faster.

src/Number.php
frederikbosch: Agreed, was left when I had support for the plugs sign. It is removed now.

frederikbosch:
Okay. I changed more. `000` is now *not* supported any more. Otherwise I moved the responsibility into a single method for integer part and a single method for the fractional part.

frederikbosch:
@sagikazarmark Any more change requests? I would love to merge this PR.

doc/getting-started.rst
```diff
@@ -17,6 +17,35 @@ All amounts are represented in the smallest unit (eg. cents), so USD 5.00 is wri
...
+    // int is accepted
+    $fiver = new Money(500, new Currency('USD'));
+
+    // string is accepted if integer
+    $fiver = new Money('500', new Currency('USD'));
+
+    // string is accepted if fractional part is zero
+    $fiver = new Money('500.00', new Currency('USD'));
+
+    // leading zero's are accepted
```
sagikazarmark: zeros

sagikazarmark: Why do we allow leading zeros?

doc/getting-started.rst
```diff
@@ -17,6 +17,35 @@ All amounts are represented in the smallest unit (eg. cents), so USD 5.00 is wri
...
+    $fiver = new Money('00500', new Currency('USD'));
+
+    // multiple zero's is not accepted, throws \InvalidArgumentException
```
sagikazarmark: zeros are

sagikazarmark: IMO not accepted is always the same as InvalidArgumentException, so it should be mentioned once somewhere.

doc/getting-started.rst
```diff
@@ -17,6 +17,35 @@ All amounts are represented in the smallest unit (eg. cents), so USD 5.00 is wri
 
 See :doc:`features/parsing` for additional ways to instantiate a Money object from strings.
 
+Accepted integer values
+-----------------------
```
sagikazarmark: Are you sure this should be in the getting started guide? I would say we cover edge cases with this documentation.

src/Money.php
```diff
@@ -56,8 +56,13 @@
...
     {
-        if (filter_var($amount, FILTER_VALIDATE_INT) === false && !Number::fromString($amount)->isInteger()) {
-            throw new \InvalidArgumentException('Amount must be an integer(ish) value');
+        if (filter_var($amount, FILTER_VALIDATE_INT) === false) {
+            $numberFromString = Number::fromString($amount);
+            if (!$numberFromString->isInteger()) {
+                throw new \InvalidArgumentException('Amount must be an integer(ish) value');
+            }
+
+            $amount = $numberFromString->getIntegerPart();
```
sagikazarmark: I am still failing to see how this is a solution. Can you please enlighten me, because I am missing something. isInteger checks if the fractional part is empty, and getIntegerPart returns the integer part. But if the fractional part is empty, $amount itself should be valid as well, shouldn't it?

src/Number.php
```diff
@@ -166,23 +167,64 @@ public function getIntegerRoundingMultiplier()
...
+
+        $number = ltrim($number, '0');
+        if ('' === $number) {
+            throw new \InvalidArgumentException(
+                'Invalid integer part '.$number
+            );
+        }
+
+        $position = 0;
+        while (isset($number[$position])) {
```
sagikazarmark: Still don't see why not a for loop here

tests/NumberTest.php
```diff
@@ -14,19 +14,20 @@ public function it_has_attributes($number, $decimal, $half, $currentEven, $negat
...
         $number = Number::fromString($number);
 
-        $this->assertEquals($decimal, $number->isDecimal());
-        $this->assertEquals($half, $number->isHalf());
-        $this->assertEquals($currentEven, $number->isCurrentEven());
-        $this->assertEquals($negative, $number->isNegative());
-        $this->assertEquals($integerPart, $number->getIntegerPart());
-        $this->assertEquals($fractionalPart, $number->getFractionalPart());
-        $this->assertEquals($negative ? '-1' : '1', $number->getIntegerRoundingMultiplier());
+        $this->assertSame($decimal, $number->isDecimal());
```
sagikazarmark: I remember we had an issue with Same. Maybe it was related to gmp?

doc/getting-started.rst
frederikbosch: Because you requested that feature I believe, I am fine with not supporting it.

src/Money.php
frederikbosch: Before, if you pass 3.000, then the private property 'amount' would be set to '3.000'. If you then do equals on 'new Money (3, new Currency())' it would fail. Afterwards, it will not. It is a good solution.

src/Number.php
frederikbosch: And I dont see a problem with while.

tests/NumberTest.php
frederikbosch: Dont recall that.

doc/getting-started.rst
sagikazarmark: Did I? I think I meant something like the fractional part, and further chunks of the integer part can start with zeros.

src/Money.php
sagikazarmark: Yeah, but i case of 3.000 isInteger would be false, wouldn't it?

src/Number.php
sagikazarmark: IMO it's cleaner to use for when using a counter.

src/Money.php
frederikbosch: Nope. BC math returns like this, depending on the scale variable.

tests/NumberTest.php
frederikbosch: Actually I do. It's because of the scale variable with BC math. Then trailing zero's are added. So we are not using `assertSame` in calculator tests.

src/Number.php
frederikbosch: Updated.

doc/getting-started.rst
frederikbosch: Updated

doc/getting-started.rst
frederikbosch: Dropped support for leading zero's

doc/getting-started.rst
frederikbosch: Let's not worry too much on this. Getting started should help how to construct a Money object. I agree - it is the edge cases. But let's accept it now. If we want to move it to another section later, we can always do that.

wimby:
@frederikbosch sorry for delayed response.
> I did accept '000'... If you have strong arguments against this, please let me know.

It is probably irrelevant now, but if you asked...

I don't have any arguments against accepting 000 or +123 as inputs. I usually try to go with some standard (like common arithmetics which is intuitive to user or common subset of PHP, BCMath and GMP which can help interoperability).
Its true that it involves more code not to accept it, but another code would be required for e.g. comparison.
On the other hand *be liberal in what you accept and conservative in what you send*, so probably accepting 000 and +123 and then internally normalize it to 0 and 123 is more robust.

But I think that the most important is _specification_ what is accepted than _what_ is accepted, because it's constraint and it's part of the interface.

frederikbosch:
@wimby Since documentation is part of this part we do have 'specification' after merging.

@sagikazarmark I want to finish this. It's open for too long.

At this moment:
- trailing zero's are (must be due to BC math)
- leading zero's are not accepted
- only zero's with length more than 1 is not accepted
- plus sign is not accepted

doc/getting-started.rst
sagikazarmark: I still think this does not belong to a Getting started guide, but let's not block the PR with it.

src/Number.php
```diff
@@ -166,23 +167,61 @@ public function getIntegerRoundingMultiplier()
...
-                return '' !== $chunk && false === filter_var($chunk, FILTER_VALIDATE_INT);
+        if ('-' === $number) {
+            return '-0';
+        }
+
+        $nonZero = false;
+
+        for ($position = 0, $characters = strlen($number); $position < $characters; ++$position) {
+            $digit = $number[$position];
+            if (!isset(static::$numbers[$digit]) && !($position === 0 && $digit === '-')) {
```
sagikazarmark: For variables we use Yoda condition (constant value, operator, variable)

src/Number.php
```diff
@@ -166,23 +167,61 @@ public function getIntegerRoundingMultiplier()
...
+                throw new \InvalidArgumentException(
+                    'Invalid integer part '.$number.'. Invalid digit '.$digit.' found'
```
sagikazarmark: I would use sprintf here. For one we do it elswhere, secondly it's more readable IMO.

src/Number.php
```diff
@@ -166,23 +167,61 @@ public function getIntegerRoundingMultiplier()
...
+                );
             }
-        );
 
-        return count($invalid) === 0;
+            if ($nonZero === false && $digit === '0') {
+                throw new \InvalidArgumentException(
+                    'Leading zero\'s are not allowed'
```
sagikazarmark: Leading zeros, not zero's

src/Number.php
```diff
@@ -166,23 +167,61 @@ public function getIntegerRoundingMultiplier()
...
+            $digit = $number[$position];
+            if (!isset(static::$numbers[$digit]) && !($position === 0 && $digit === '-')) {
+                throw new \InvalidArgumentException(
+                    'Invalid integer part '.$number.'. Invalid digit '.$digit.' found'
+                );
             }
-        );
 
-        return count($invalid) === 0;
+            if ($nonZero === false && $digit === '0') {
```
sagikazarmark: Yoda

src/Number.php
```diff
@@ -166,23 +167,61 @@ public function getIntegerRoundingMultiplier()
...
+    }
+
+    /**
+     * @param string $number
+     *
+     * @return string
+     */
+    private static function parseFractionalPart($number)
+    {
+        if ($number === '') {
```
sagikazarmark: Yoda

src/Number.php
```diff
@@ -166,23 +167,61 @@ public function getIntegerRoundingMultiplier()
...
+            return $number;
+        }
+
+        for ($position = 0, $characters = strlen($number); $position < $characters; ++$position) {
+            $digit = $number[$position];
+            if (!isset(static::$numbers[$digit])) {
```
sagikazarmark: New line before control structures

src/Number.php
```diff
@@ -166,23 +167,61 @@ public function getIntegerRoundingMultiplier()
...
+                throw new \InvalidArgumentException(
+                    'Invalid fractional part '.$number.'. Invalid digit '.$digit.' found'
```
sagikazarmark: Sprintf

src/Number.php
wimby: Just for curiosity, what is the reason for using Yoda style? I would guess that typos are hardly imaginable with tests, code reviews and triple equal operators... isn't readability more valuable?

sagikazarmark: We started using this style, so it's rather about being consistent.

PR #342, opened 2017-01-02
Ignore the fractional part after validating it is full of zeroes

VaclavSir:
Code example:

```php
$amountA = new Money(5, new Currency('CZK'));
$amountB = new Money('5.00', new Currency('CZK'));

var_dump($amountA->compare($amountB));
var_dump($amountA->equals($amountB));
```

Expected output:

```
int(0)
bool(true)
```

Actual output:

```
int(0)
bool(false)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 
 - Reversed Currencies Exchange to try resolving reverse of a currency pair
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


sagikazarmark:
Thanks for the contribution. To be completely honest, I don't really see how this works. Being an integer means there is no fractional part, in which case the string casting should be the same as the integer part. Not saying this is wrong, just trying to understand what's going on.

@frederikbosch can you please take a look?

frederikbosch:
@sagikazarmark @VaclavSir This is fixed in #341 too. Thanks for reporting and the fix.

PR #344, opened 2017-01-14
Fix values in comparison examples

miclf:
In the [documentation page about comparison](http://moneyphp.org/en/latest/features/comparison.html) of Money objects, two examples have an inconsistency between the real value and the value that is specified in the accompanying comment. This PR simply fixes that.

### DIFF
--- a/doc/features/comparison.rst
+++ b/doc/features/comparison.rst
@@ -45,7 +45,7 @@ Greater Than
 .. code-block:: php
 
     $value1 = Money::USD(800);                  // $8.00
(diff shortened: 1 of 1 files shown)


frederikbosch:
Thanks!

sagikazarmark:
Thanks for the PR

miclf:
@frederikbosch & @sagikazarmark it’s nothing :)

Thanks for making submitting PRs to the docs easy in the first place because, for tiny fixes like this, too many projects underestimate how perceived easiness is important in one’s decision to contribute or not.

Issue #347, opened 2017-01-31
Support for historical exchange rates


stefandoorn:
The search didn't show an early PR / discussion to me regarding this: support for converting using historical exchange rates. There is of course the option to use the `FixedExchange`, but this requires loading the exchange rates separately.

Money already supports the `SwapExchange`, using Swap's `latest` method to retrieve the exchange rate. But, Swap also offers an `historical` method to get the exchange rate of an earlier moment in time. 

I'm in need of it for a current project and if there is interest for including it, I could submit a PR. 

If so, there a some options of implementation:

* Add a method to `SwapExchange` & and a method to `Converter`
* Add an optional parameter to `SwapExchange::quote` & `Converter::quote`
* Add an extra Exchange class, e.g. `HistoricalSwapExchange` & and add method to `Converter`

Etc.

What would have your preference?

sagikazarmark:
Thanks for contributing to the project.

I think the third option sounds like the cleanest, but modifying the converter sounds a little bit weird to me. It has a nice, single responsibility, I wouldn't add more to that.

Probably adding a historical exchange interface, an implementation for swap and a HistoricalConverter could be the way to go?

WDYT @frederikbosch

stefandoorn:
I had to move on already a bit to get the support in my current project, so I took your suggestions and forked the repo and implemented support.

Progress here: stefandoorn/moneyphp@7b267d1b625798dceb2f4ab5861fdb00d42b4a79.

If this makes sense, I can finalise and submit the PR.

PR #350, opened 2017-02-03
Add support for Historical Exchange Rates

stefandoorn:
Discussion in issue #347. This is the first work, if adjustments are required, let me know.

### DIFF
--- a/doc/features/currency-conversion.rst
+++ b/doc/features/currency-conversion.rst
@@ -91,6 +91,29 @@ Then conversion is quite simple:
 .. _Composer: https://getcomposer.org
 
 
+Third Party Exchange At Historical Date
+---------------------------------------
+
+Next to the ``Money\Exchange`` mentioned before, we also provide a way to integrate external
+sources of conversion rates at historical dates by implementing the ``Money\HistoricalExchange``
+interface. There is a default one as mentioned above.
+
+Then conversion is quite simple:
+
+.. code:: php
+
+    use Money\Money;
+    use Money\HistoricalConverter;
+
+    // $swap = Implementation of \Swap\SwapInterface
+    $exchange = new HistoricalSwapExchange($swap);
+
+    $converter = new HistoricalConverter(new ISOCurrencies(), $exchange);
+    $date = new \DateTime();
+    $eur100 = Money::EUR(100);
+    $usd125 = $converter->convert($eur100, new Currency('USD'), $date);
+
+
 CurrencyPair
 ------------
 
--- a/spec/Exchange/SwapExchangeSpec.php
+++ b/spec/Exchange/SwapExchangeSpec.php
@@ -9,6 +9,7 @@ use Money\CurrencyPair;
 use Money\Exception\UnresolvableCurrencyPairException;
 use Money\Exchange;
 use Money\Exchange\SwapExchange;
+use Money\HistoricalExchange;
 use Swap\Swap;
 use PhpSpec\ObjectBehavior;
 
@@ -27,6 +28,7 @@ final class SwapExchangeSpec extends ObjectBehavior
     function it_is_an_exchange()
     {
         $this->shouldImplement(Exchange::class);
+        $this->shouldImplement(HistoricalExchange::class);
     }
 
     function it_exchanges_currencies(Swap $swap, ExchangeRate $exchangeRate)
@@ -49,5 +51,27 @@ final class SwapExchangeSpec extends ObjectBehavior
 
         $this->shouldThrow(UnresolvableCurrencyPairException::class)
             ->duringQuote(new Currency('EUR'), new Currency('XYZ'));
+    }
+
+    function it_exchanges_currencies_for_a_historical_date(Swap $swap, ExchangeRate $exchangeRate, \DateTime $date)
+    {
+        $exchangeRate->getValue()->willReturn(1.0);
+
+        $swap->historical('EUR/USD', $date)->willReturn($exchangeRate);
+
+        $currencyPair = $this->historical($base = new Currency('EUR'), $counter = new Currency('USD'), $date);
+
+        $currencyPair->shouldHaveType(CurrencyPair::class);
+        $currencyPair->getBaseCurrency()->shouldReturn($base);
+        $currencyPair->getCounterCurrency()->shouldReturn($counter);
+        $currencyPair->getConversionRatio()->shouldReturn(1.0);
+    }
+
+    function it_throws_an_exception_when_cannot_exchange_historical_currencies(Swap $swap, \DateTime $date)
+    {
+        $swap->historical('EUR/XYZ', $date)->willThrow(Exception::class);
+
+        $this->shouldThrow(UnresolvableCurrencyPairException::class)
+            ->duringHistorical(new Currency('EUR'), new Currency('XYZ'), $date);
     }
 }
--- a/spec/Exchange/SwapExchangeSpec.php
+++ b/spec/Exchange/SwapExchangeSpec.php
@@ -53,10 +53,11 @@ final class SwapExchangeSpec extends ObjectBehavior
             ->duringQuote(new Currency('EUR'), new Currency('XYZ'));
     }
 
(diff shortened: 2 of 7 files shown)


stefandoorn:
Hm, HHVM segmentation fault..

sagikazarmark:
Thanks for the contribution.

I am thinking about making the two exchange interfaces somehow compatible, so the same underlying implementations can be implemented in one class.

Maybe rename HistoricalExchange::quote to quoteHistorical or just historical and implement this interface on the existing SwapExchange?

I feel that having the same method name can be confusing, also using the same object as kind of a facade could make using it a bit easier.

WDYT?

stefandoorn:
I'll take a look on that this week. I'm ok with it :) I think the main criteria should be that we are not introducing BC.

stefandoorn:
@sagikazarmark I've been thinking about this, but I'm afraid extending the current interface is going to incur BC, as it will require implementations (which are not always just this library) will need to add this additional method also. So then this would require releasing a new major release. Or, am I missing something?

We might be able to have 2 interfaces and implement both interfaces on the implementing class. That could be of something. Then it's kind of optional to the implementing party. Shall I take a look into that?

sagikazarmark:
> I've been thinking about this, but I'm afraid extending the current interface is going to incur BC

Of course it is, never said anything else. What I meant that it should be possible to implement the two interfaces in ONE CLASS.

stefandoorn:
Alright. 

Additional: what do you think of combining Converter & HistoricalConverter, either by adding a 'convertHistorical' method or adding an optional variable to 'convert' and checking inside whether Exchange implemented the HistoricalExchange and throwing an exception if not?

frederikbosch:
Thought about this problem. The best thing we could to do is to add a DateTime argument to `Exchange` and `Converter` because an exchange and a conversion is always on a certain moment. It is up to the implementation to do something with it or not. But because that leads to BC breaks, we cannot do that right now.

Is there any alternative? Yes, and that involves no change to the library code at all. If I had to solve this without arguments in the exchange and converter classes, I'd go for something like this.

```
<?php
namespace External\Controller;

class Controller {
  private $converterFactory;
  
  public function __construct(External\Service\HistoricalConverterFactory $converterFactory) {
    $this->converterFactory = $converterFactory;
  }

  public function index ($date, $amount, $currency, $counterCurrency) {
     $converter = $this->converterFactory->factory(DateTimeImmutable::createFromFormat($date));
     $counterAmount = $converter->convert(new Money($amount, new Currency($currency)), new Currency($counterCurrency));
  }
}

namespace External\Service;

class HistoricalConverterFactory {
  private $database;
  private $currencies;
  
  public function __construct(Database $database, Currencies $currencies) {
    $this->database = $database;
    $this->currencies = $currencies;
  }

  public function factory(DateTimeImmutable $date) {
    $rates = $this->database->select()->where('date', '=', $date->format('Y-m-d'))->get();
    // modify $rates as such that it becomes a currency to currency (array) map
    $exchange = new FixedExchange($rates);
    return new Converter($this->currencies, $exchange);
  }

}
```

Actually, I think this is maybe an even better approach than to add the argument. At least it is a perfect separation of concerns. But, a date (moment) is part of the exchange/conversion domain. So it should eventually become part of the library, in a version where BC would be allowed.

frederikbosch:
@sagikazarmark How do you think about it? My suggestion would be to close the issue, leave the PR open and tag it as version 4.

sagikazarmark:
Whoa, not so fast. 😄 

Honestly I don't really like the idea of making DateTime a mandatory argument. Although strictly speaking what you are saying is true, I hardly think that in most of the use cases we need that kind of precision. On the other hand it adds a considerable amount of complexity to not just the code, the interface itself.

I think payment is hardly ever handled by applications using this library, so if someone REALLY need realtime/historical currency rates, then he/she can use our more sophisticated implementation of Historical rates.

If we targeted PHP 5.6, we could use something like this probably for arbitrary options passed to the convert method: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

Until then @stefandoorn's solution for a separate convertHistorical method sounds like a viable way for me.

frederikbosch:
@sagikazarmark Alright. But why do you prefer this method over the factory pattern - using existing classes - that I am suggesting? I mean: why should we extend the library with functionality that is not required? What do you we gain from adding two interfaces and consequent implementations?

jarobe42:
Apologies for jumping in, but I had a go at extending @stefandoorn's changes as per the discussion: https://github.com/moneyphp/money/compare/master...jarobe42:historical-exchange

Basically, I've added an extra `HIstoricalExchange` interface which extends `Exchange`. Along with this, Converter has a `convertHistorical` which throws an exception if called when the Converter's exchange does not implement `HIstoricalExchange`

Before I go further, is my understanding of the suggested changes correct?

frederikbosch:
@sagikazarmark When reading this I see that there was no reply to the factory pattern in this case. I still think that is the best solution in this case. See my code above.

jarobe42:
@frederikbosch The issue I see with the Factory Pattern is that a new Converter instance is going to be created for each and every historical date. This could potentially grow quite large if you were to be iterating over a list of values of different dates that needed to be converted.

frederikbosch:
Why would the list be large? If you iterate it, the extra converter class would be gone when you move to the next item.

sagikazarmark:
Sorry, I'm on a holiday now, this has to wait a little. 🌴

frederikbosch:
@sagikazarmark Enjoy! Well deserved I guess!

jarobe42:
> Why would the list be large? If you iterate it, the extra converter class would be gone when you move to the next item.

In which case we'd lose the ability to reuse a converter for a repeated date - we'd be instantiating a new Converter each time. It's not a huge issue, but I just think it fails to really solve the problem.

Perhaps my interpretation is wrong, but to me, the Converter object should not be pinned to a date - rather it's just extra information to be passed on to the Exchange.

frederikbosch:
> In which case we'd lose the ability to reuse a converter for a repeated date - we'd be instantiating a new Converter each time. It's not a huge issue, but I just think it fails to really solve the problem.

That's not true. Your factory can return the same object, e.g. internally cached in an array, right? Or you can increase the design and create a `CachedHistoricalConverterFactory` and `SqlHistoricalConverterFactory` which comes along with the `HistoricalConverterFactory` interface and acts as a decorator for the `SqlHistoricalConverterFactory`.

And then you inject the interface in your controller, which means you lean on an interface instead of a concretion. And that's what we always want, right?

frederikbosch:
@sagikazarmark I'd love to make a final decision on the issue. People have made their points, no need to extend the discussion. My viewpoint is clear: our current solution is sufficient for this use case, I see no reason to extend this library as suggested. @sagikazarmark Your viewpoint was different. I leave the final decision up to you. Shall we go for it or not?

sagikazarmark:
Okay, so as far as I can see we are looking at two proposed changes:

- add historical rate support to exchanges
- add historical conversion support

As for the exchange part, I don't see any problem with it. Some exchange implementations indeed support historical rates. I could even imagine that the abstraction you suggested @frederikbosch be implemented as an exchange as well, not a converter factory (pass the date to the exchange, which wraps a historical exchange).

On the other hand, I have to agree with @frederikbosch that relying on historical rates from whatever source doesn't seem to be a good idea. So keeping some kind of local cache/list for rates you are interested in sounds like a better choice.

As for the converter part: I'm really not sure about it. I have to think about it a bit.

WDYT @frederikbosch ?

frederikbosch:
@sagikazarmark I think we should leave the functionality as is and not extend it to historical rates. The current interfaces allow developer to take any path to extend conversion their own way.

1. They can add an interfaces in their own code for historical conversion and then implement that interface in their code.
2. They can extend our interfaces in their own code (would not be my preference though) and then implement those interfaces in their code.
3. They can go for the factory solution I proposed above.

I think that this package must not provide a 100% technical solution to the problem. Let the developers decide for themselves.

frederikbosch:
Closed until someone decides it should be implemented.

PR #353, opened 2017-02-14
Fixes #352, adds scale when comparing decimal numbers for bc math

### LINKED ISSUE #352: Divide method and division by zero checking
The divide function uses the current calculator to check for a zero divisor:
```
if ($this->getCalculator()->compare((string) $divisor, '0') === 0) {
        throw new \InvalidArgumentException('Division by zero');
}
```

As the BCMathCalculator doesn't use any precision value, values < 1 are treated as 0 (which shouldn't be a problem for values that are not even close to 0).

For example, if I have a value which has received a 10% discount (original * 0.9), dividing by 0.9 would give me the original value before the discount.

frederikbosch:


sagikazarmark:
Looks good to me. Can you add the exact same example provided by the issue reporter, to make sure his use case is covered?

sagikazarmark:
I added the examples to the Calculator test case and it broke Gmp. 😢

sagikazarmark:
@frederikbosch can you please review my changes?

frederikbosch:
@sagikazarmark Thanks for the help here. Merged the master into this branch because I recently released 3.0.1. There were quite a number of fixes. Bugfixes should not wait too long with a release. I know how much I appreciate it when a maintainer follows a bugfix directly with a tag/release. So I released a minor version in February.

frederikbosch:
@sagikazarmark I modified your solution. I think this one is easier to understand. Thanks for your noticing that my initial solution lacked scope.

frederikbosch:
@sagikazarmark If Travis thinks is alright, I merging it in.

sagikazarmark:
@frederikbosch careful!!! Why did you merge master into this branch? Please revert!

frederikbosch:
Why revert? Tests are OK.

sagikazarmark:
Because of the history. I will fix it :\

frederikbosch:
What history?

sagikazarmark:
Ah, okay, you squashed the commits, it's fine. Please do not merge the master branch into feature branches. Use rebase if you have to update the history.

frederikbosch:
@sagikazarmark Okay, I am still not familiar how rebasing (really) works.

sagikazarmark:
> Merge branch 'master' into bc-compare-zero

This one is not really good.

Rebase is simple: it rewinds the HEAD to the latest common point with master, then replays the commits on that tree. This way commits will look like they were added after the latest master state. But it requires some care too: it rewrites the history, so you have to force push, which will conflict for others.

Squashing before merging is also okay, but in that case we don't have a preview of the state before merging.

PR #354, opened 2017-02-17
Refactors ISOCurrencies, removes duplication of code.

Sharom:


frederikbosch:
I dont have a problem with this. However, I do prefer isset over array_key_exists. But I dont have a specific reason why.

sagikazarmark:
I will tell you the reason: the key differences are that isset is a language construct, array_key_exists is not, also the first checks if an array has a value for a specific key, while the latter checks if a specific key exists in the array (without accessing the value).

frederikbosch:
Will be merged when Travis feels OK with #359.

frederikbosch:
Thanks, merged and tagged.

PR #359, opened 2017-03-11
Refactoring of iso currencies

frederikbosch:


frederikbosch:
Same as #354 but now with `isset` instead of `array_key_exists`.

frederikbosch:
@sagikazarmark Let's tag after this is merged...

frederikbosch:
@sagikazarmark I updated the changelog, so a new tag will follow after travis has approved the latest merges with the master.

frederikbosch:
@sagikazarmark Thanks for style fix. I know I am also not great at that. You are the guard regarding code style.

PR #362, opened 2017-03-20
Parsing empty strings and number starting or ending with a decimal point

### LINKED ISSUE #360: DecimalMoneyParser() not parsing some common numeric values
We have had to put in extra checks to modify some values before passing it on to the parser. The parse throws an exception if it is given any of these:

* An empty string (arguably correct)
* \.99 (No digits before the decimal point)
* 99\. (No digits after the decimal point)

I would argue that the second two at least are unambiguous in what value they represent and should be parsed. The previous `mathiasverraes/money` package we used parsed all these, including an empty string to zero.

Our workaround is to append or prepend a "0" if the value to parse starts or ends with a decimal point.

frederikbosch:
Fixes #360.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Parsing empty strings and number starting or ending with a decimal point for DecimalMoneyParser
 - Multiplying and dividing with a locale that use commas as separator
 
 ## 3.0.2 - 2017-03-11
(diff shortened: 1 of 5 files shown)


frederikbosch:
Added tests for all parsers. Only the `DecimalMoneyParser` can parse empty strings because the other parsers require a currency sign to be present.

frederikbosch:
@sagikazarmark If this is alright, I will tag a new version.

PR #367, opened 2017-03-28
Correct two old mentions of integer amounts to strings

jaikdean:
I found a couple of places which weren't updated when internal amounts were changed from ints to strings.


frederikbosch:
Thanks!

PR #371, opened 2017-04-08
Add to string method

jycr753:
This is really use full when using package on blade templates

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -517,4 +517,14 @@ final class Money implements \JsonSerializable
 
         return self::$calculator;
     }
+
+    /**
+     * Handle to string formatting.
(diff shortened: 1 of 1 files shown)


sagikazarmark:
I think we specifically rejected these PRs and requests in the past, because:

- not everyone expects the same format when casting money to string
- we have formatters

That said, I would consider merging this as this contains the less formatting logic possible. However, I think it can still be confusing, not to mention it's not useful at all when you use subunits.

frederikbosch:
@sagikazarmark I'd say we do not do this. Maybe we should create .github where specifically mention that we refuse related to `__toString` and removing `final`.

PR #372, opened 2017-04-19
Remove blank currency from ISOCurrencies

MichaelGooden:
Ready to merge.

Currently, due to an issue in the `moneyphp/iso-currencies` repository, a 'blank' currency is present in the ISOCurrencies repository.

There is a fix for that issue pending https://github.com/moneyphp/iso-currencies/pull/4.

Initial tests have been added for proof of claim and regression testing. Once the issue in iso-currencies has been resolved, this PR will be updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Change Log
 
+## 3.0.4 - 2017-04-20
+
+### Fixed
+
+- ISOCurrencies will no longer have a blank currency
+
+### Changed
+
+- Updated ISO Currencies
+- Removed old Belarusian ruble from ISOCurrencies (BYR)
+
 ## 3.0.3 - 2017-03-22
 
 ### Fixed
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "cache/taggable-cache": "^0.4.0",
         "phpspec/phpspec": "^2.5",
         "henrikbjorn/phpspec-code-coverage": "^2.0.2",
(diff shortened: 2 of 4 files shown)


frederikbosch:
@MichaelGooden You can upgrade the currencies dependency to 3.0 now.

MichaelGooden:
Maybe a bit extreme to bump the iso-currencies version from 2.0 to 3.0 but it works! This should be ready to merge.

Side note: With this update the old Belarusian ruble (BYR) was removed as it has been discontinued in favour of the new Belarusian ruble (BYN). [https://en.wikipedia.org/wiki/Belarusian_ruble#Third_ruble.2C_2016_-_present]

frederikbosch:
@MichaelGooden Thanks for the update. I moved from 2 to 3 because it is essentially a BC breaking update. Could you also update the CHANGELOG? Maybe nice to add the remark of the Belarusian currency as well.

MichaelGooden:
Changelog done.

The only effective change in iso-currencies was removing essentially broken behavior that no one should have been relying on in the first place. Not BC breaking in my humble opinion. :D

frederikbosch:
Hehe, I guess you are right. Sometimes I take decisions too quickly. @sagikazarmark is probably already aware of that ;)

MichaelGooden:
Rather safe than sorry @frederikbosch. Thanks for merging this so quickly. Any chance we could get this tagged with a version?

frederikbosch:
Will tag today, have to update some more CHANGELOG.

frederikbosch:
Done.

Issue #373, opened 2017-04-26
Expose numericCode for ISOCurrencies


barryvdh:
The ISOCurrencies have a numeric ISO code, can we expose that, similar to `subunitFor`?
For usage in Omnipay for example, see https://github.com/thephpleague/omnipay-common/pull/151

sagikazarmark:
Do you mean exposing it as part of the interface or as part of that specific repository?

If it's the latter, I am not against it.

barryvdh:
Not the interface (I assume only ISOCurrencies has it), but as part of the ISOCurrencies class itself.

sagikazarmark:
I don't have a reason why we couldn't, but in that case you need to directly rely on that repository. It depends on your use case if it's good or bad.

@frederikbosch anything to add?

@barryvdh would you mind opening a PR?

frederikbosch:
Hmmm... the ISOCurrencies states the following.

> Please be aware that this library solely exists to provide iso currencies for moneyphp/money. Using this library for other purposes is at your own risk. Numeric codes are not use, because we didn't need it.

Therefore, if numeric codes will serve a purpose for *functionality / domain logic* within the Money library, I will obviously agree. If, however, it is being done purely to extend our currencies class as *repository / persistence logic*, I am against it.

If this is for the latter case, I would recommend to create your own currencies repository. You are allowed to grab my code from the [ISOCurrencies](https://github.com/moneyphp/iso-currencies) package and turn it into your own needs. Then you still have the opportunity to implement the `Currencies` interface, if you want to serve your own class as repository for `Money`. It will serve as perfect boundary.

But why not simply extend the class/repository? `Money` is created for domain logic, not for currency repositories. We should focus on that, and do that well. Moreover, the amount of code you need to create your own repo [is very small](https://github.com/moneyphp/iso-currencies/tree/master/src).  

@barryvdh If I were you, I'd have the currencies in your package, even if we would extend our repository. Having that responsibility in your package, gives you the power to change. It is not worth it to have it as an abstraction that is handled by someone else.

frederikbosch:
@barryvdh Why do you need it any how?

barryvdh:
I have them in my package currently, even copied from that package: https://github.com/thephpleague/omnipay-common/blob/master/update_currencies.php

[...]

PR #374, opened 2017-04-26
Add numericCode method to ISOCurrencies

barryvdh:
This adds a method to retrieve the NumericCode, as discussed in #373


frederikbosch:
I do not fully agree, but what the heck. I think you would want to have a tag too? :)

frederikbosch:
Tagged! Good luck with Omnipay!

barryvdh:
Thanks :)

PR #377, opened 2017-06-07
Update bitcoin.rst

gerardroche:
BitcoinMoneyParser accepts only number of fraction digits: `BitcoinMoneyParser::__construct($fractionDigits)`

### DIFF
--- a/doc/features/bitcoin.rst
+++ b/doc/features/bitcoin.rst
@@ -27,7 +27,7 @@ Please see the example below how to use the Bitcoin currency:
     echo $formatter->format($money); // prints Ƀ1000.00
 
     // parse bitcoin
-    $parser = new BitcoinMoneyParser($intlParser, 2);
+    $parser = new BitcoinMoneyParser(2);
     $money = $parser->parse("\0xC9\0x831000.00", 'XBT');
     echo $money->getAmount(); // outputs 100000000000
 


frederikbosch:
@sagikazarmark Continue supporting hhvm and update travis? Or drop? Maybe drop next major version.

sagikazarmark:
HHVM support was never my top priority. As a fellow PHP developer expressed what HHVM was good for:

> hhvm seems to have done its job by pushing the php core devs to make php 7 about as fast as hhvm

frederikbosch:
Totally agree. Drop on Travis now then? Or wait for 4.x?

frederikbosch:
It fails due to misconfiguration (something on ubuntu version). We should update the travis.yml. Adding it to allowed_failures would be the same as removing it. Because we are not testing hhvm now/then.

frederikbosch:
So we have three options.

1. Make sure HHVM tests work, but are allowed to fail
2. Make sure HHVM tests work, but are not allowed to fail
3. Drop HHVM tests

My preference is number three.

sagikazarmark:
I would say as a quick fix we should just allow it to fail then invest some time and check if we can fix the tests and properly support HHVM. If we can do that with a reasonable effort then we should, if not we can drop it entirely. I don't recall exactly what HHVM specific code we have, so I can't make a commitment to drop it right now.

frederikbosch:
Bueno

frederikbosch:
Thanks!

PR #378, opened 2017-06-21
Correct scale for add/subtract

angekouf:
In add and subtract functions you run bcmath with 0 scale which floored the result.

### DIFF
--- a/src/Calculator/BcMathCalculator.php
+++ b/src/Calculator/BcMathCalculator.php
@@ -45,7 +45,7 @@ final class BcMathCalculator implements Calculator
      */
     public function add($amount, $addend)
     {
-        return bcadd($amount, $addend, 0);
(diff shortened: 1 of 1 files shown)


sagikazarmark:
@frederikbosch can you please review this?

frederikbosch:
@sagikazarmark Will do, it seems OK. But I am missing tests that prove the point.

@angekouf Can you add a test that shows that there the calculator is not correct at some point?

sagikazarmark:
Ping @angekouf can you please take a look at the tests?

sagikazarmark:
Ping @angekouf can you please take a look at the tests?

frederikbosch:
Tests fail, no clear reason why to do it. It can only make sense when using PreciseMoney. Closing until better explanation given.

PR #384, opened 2017-07-18
Add IndirectExchange

mcordingley:
Adds a decorator for `Exchange` that will query the underlying exchange to find a shortest chain of conversions to go from the base currency to the counter currency using breadth-first search.

### DIFF
--- a/src/Exchange/IndirectExchange.php
+++ b/src/Exchange/IndirectExchange.php
@@ -1,0 +1,142 @@
+<?php
+
+namespace Money\Exchange;
+
+use Money\Currencies;
+use Money\Currency;
+use Money\CurrencyPair;
+use Money\Exception\UnresolvableCurrencyPairException;
+use Money\Exchange;
+use SplQueue;
+use stdClass;
+
+/**
+ * Provides a way to get an exchange rate through a minimal set of intermediate conversions.
+ *
+ * @author Michael Cordingley <Michael.Cordingley@gmail.com>
+ */
+final class IndirectExchange implements Exchange
+{
+    /**
+     * @var Currencies
+     */
+    private $currencies;
+
+    /**
+     * @var Exchange
+     */
+    private $exchange;
+
+    /**
+     * @param Exchange $exchange
+     * @param Currencies $currencies
+     */
+    public function __construct(Exchange $exchange, Currencies $currencies)
+    {
+        $this->exchange = $exchange;
+        $this->currencies = $currencies;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function quote(Currency $baseCurrency, Currency $counterCurrency)
+    {
+        $rate = array_reduce($this->getConversions($baseCurrency, $counterCurrency), function ($carry, CurrencyPair $pair) {
+            return $carry * $pair->getConversionRatio();
+        }, 1.0);
+
+        return new CurrencyPair($baseCurrency, $counterCurrency, $rate);
+    }
+
+    /**
+     * @param Currency $baseCurrency
+     * @param Currency $counterCurrency
+     * @return CurrencyPair[]
+     * @throws UnresolvableCurrencyPairException
+     */
+    private function getConversions(Currency $baseCurrency, Currency $counterCurrency)
+    {
+        $nodes = $this->initializeNodes();
+
+        $startNode = $nodes[$baseCurrency->getCode()];
+        $startNode->discovered = true;
+
+        $frontier = new SplQueue();
+        $frontier->enqueue($startNode);
+
+        while ($frontier->count()) {
+            /** @var stdClass $currentNode */
+            $currentNode = $frontier->dequeue();
+
+            /** @var Currency $currentCurrency */
+            $currentCurrency = $currentNode->currency;
+
+            if ($currentCurrency->equals($counterCurrency)) {
+                return $this->reconstructConversionChain($nodes, $currentNode);
+            }
+
+            /** @var Currency $candidateCurrency */
+            foreach ($this->currencies as $candidateCurrency) {
+                /** @var stdClass $node */
+                $node = $nodes[$candidateCurrency->getCode()];
+
+                if (!$node->discovered) {
+                    try {
+                        // Check if the candidate is a neighbor. This will thrown an exception if it isn't.
+                        $this->exchange->quote($currentCurrency, $candidateCurrency);
+
+                        $node->discovered = true;
+                        $node->parent = $currentNode;
+
+                        $frontier->enqueue($node);
+                    } catch (UnresolvableCurrencyPairException $exception) {
+                        // Not a neighbor. Move on.
+                    }
+                }
+            }
+        }
+
+        throw UnresolvableCurrencyPairException::createFromCurrencies($baseCurrency, $counterCurrency);
+    }
+
+    /**
+     * @return stdClass[]
+     */
+    private function initializeNodes()
+    {
+        $currencies = [];
+
+        /** @var Currency $currency */
+        foreach ($this->currencies as $currency) {
+            $node = new stdClass;
+
+            $node->currency = $currency;
+            $node->discovered = false;
+            $node->parent = null;
+
+            $currencies[$currency->getCode()] = $node;
+        }
+
+        return $currencies;
+    }
+
+    /**
+     * @param array $currencies
+     * @param stdClass $goalNode
+     * @return CurrencyPair[]
+     */
+    private function reconstructConversionChain(array $currencies, stdClass $goalNode)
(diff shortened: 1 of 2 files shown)


mcordingley:
The `getConversions` method is intended to be able to be taken public if the user is interested in finding the exact chain of conversions needed to get from the base currency to the counter currency. I left it private, as increasing visibility is easier than decreasing it.

frederikbosch:
@mcordingley Sorry, but I do not understand the problem that is trying to be solved here. Could you elaborate?

mcordingley:
Sure. It's all about being able to resolve exchange rates between currencies that don't have a direct mapping in the base exchange. If such a direct mapping doesn't exist, then it's impossible with the current tools to resolve an exchange rate. The existence of `ReversedCurrenciesExchange` already suggests that there are gaps in the set of currencies exchanged. This complements that by finding a minimal set of conversions using intermediate currencies to make the desired exchange.

It can also enhance a fixed set of exchanges. Stacking this on top of `ReversedCurrenciesExchange` means that a fixed exchange with only mappings from USD to other currencies could now map between any two currencies found in the exchange.

frederikbosch:
@mcordingley Got it. Will review accordingly!

frederikbosch:
@sagikazarmark I am in favor of this feature. How about you?

I have got some style and code remarks on the PR, but before going into detail I rather have an agreement on the feature first.

bendavies:
Nice PR!

src/Exchange/IndirectExchange.php
```diff
@@ -0,0 +1,144 @@
...
+        $this->currencies = $currencies;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function quote(Currency $baseCurrency, Currency $counterCurrency)
+    {
+        $rate = array_reduce($this->getConversions($baseCurrency, $counterCurrency), function ($carry, CurrencyPair $pair) {
+            return $carry * $pair->getConversionRatio();
```
bendavies: is this safe?
What about using one of the `Money\Calculator` classes?

frederikbosch: It *should* be safe. Why wouldn't it?

bendavies: well, php floating point precision!

frederikbosch: Sharp!

src/Exchange/IndirectExchange.php
```diff
@@ -0,0 +1,144 @@
...
+
+namespace Money\Exchange;
+
+use Money\Currencies;
+use Money\Currency;
+use Money\CurrencyPair;
+use Money\Exception\UnresolvableCurrencyPairException;
+use Money\Exchange;
+use SplQueue;
+use stdClass;
```
frederikbosch: Please do `use` global classes, rather use `\stdClass`.

src/Exchange/IndirectExchange.php
bendavies: Sorry, didn't mean to sound sharp if that's what you mean.
I'm just referring to the warning on this page
http://php.net/manual/en/language.types.float.php

tests/Exchange/IndirectExchangeTest.php
```diff
@@ -0,0 +1,79 @@
...
+     */
+    public function it_calculates_adjacent_nodes()
+    {
+        $exchange = $this->createExchange();
+
+        $pair = $exchange->quote(new Currency('USD'), new Currency('EUR'));
+
+        static::assertEquals('USD', $pair->getBaseCurrency()->getCode());
+        static::assertEquals('EUR', $pair->getCounterCurrency()->getCode());
+        static::assertEquals(4, $pair->getConversionRatio());
```
frederikbosch: use `$this->assertEquals`

tests/Exchange/IndirectExchangeTest.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * @test
+     */
+    public function it_throws_when_no_chain_is_found()
+    {
+        $exchange = $this->createExchange();
+
+        static::setExpectedException(UnresolvableCurrencyPairException::class);
```
frederikbosch: use `$this->setExpectedException`.

src/Exchange/IndirectExchange.php
frederikbosch: Not at all, `sharp` in the meaning of thank you for your good (sharp) remark.

bendavies: hah, good. I wasn't sure which definition you were using :)

tests/Exchange/IndirectExchangeTest.php
bendavies: is this for consistency? they are actually static....

src/Exchange/IndirectExchange.php
bendavies: so, shall we use a Calculator?

tests/Exchange/IndirectExchangeTest.php
frederikbosch: @bendavies Yes, if we start doing this, we should update all our tests. I rather have consistency indeed.

src/Exchange/IndirectExchange.php
mcordingley: I don't entirely follow. Did you mean "Please don't `use`"?

mcordingley: Just pushed a removal of those `use` statements.

tests/Exchange/IndirectExchangeTest.php
mcordingley: Updated.

tests/Exchange/IndirectExchangeTest.php
mcordingley: Updated.

src/Exchange/IndirectExchange.php
mcordingley: If precision is a concern here, then there _has_ to be a `Calculator`.

mcordingley: Added a `Calculator` argument to the constructor. What do you think about letting it be an optional argument that defaults to one of the built-in `Calculator` classes?

bendavies:
Unfortunately there is no registry to get a supported `Calculator`, so I can't see a nice way of inferring a calculator automatically apart from duplicating https://github.com/moneyphp/money/blob/cf363f54cb3b8fe9d2c1eb82de4e6e6bbb1e9fa8/src/Money.php#L486-L527

mcordingley:
Seems as if that code should get shifted out into a `CalculatorRegistry` that classes can use. I don't have the particulars fully worked out in my head and it's beside the point of this PR. I think I'd rather keep this code dumb with regards to resolving calculators.

sagikazarmark:
I also like the idea, going through the implementation now.

About Calculators: I would like to keep that an internal detail if possible. So for now, I would just copy the code from the Money class and rewrite Calculator handling in a separate PR. Since it's an internal detail, we are allowed to make changes there.

frederikbosch:
Good job! Will review again soon. Think we can merge it in soon.

frederikbosch:
@mcordingley Did you consider to 'try' the decorated exchange *directly* before actually going for the *indirect* conversion?

```php
    public function quote(Currency $baseCurrency, Currency $counterCurrency)
    {
        try {
                return $this->exchange->quote($baseCurrency, $counterCurrency);
        } catch (UnresolvableCurrencyPairException $e) {
                $rate = array_reduce($this->getConversions($baseCurrency, $counterCurrency), function ($carry, CurrencyPair $pair) {
                    return static::getCalculator()->multiply($carry, $pair->getConversionRatio());
                }, '1.0');
                return new CurrencyPair($baseCurrency, $counterCurrency, $rate);
        }
    }
```

mcordingley:
That _would_ speed things up in the common case of a direct mapping.

frederikbosch:
@mcordingley Exactly, so, I prefer to include it. Afterwards I think it is good to go.

mcordingley:
Good catch. That's in, and I think we're good to go. The only part I don't like is the `Calculator` handling code, but that's a whole separate issue. 👍

frederikbosch:
@mcordingley We could reconsider the calculator part in next major versions.

frederikbosch:
Tag will have to wait until tomorrow.

frederikbosch:
Thanks @mcordingley. Great feature!

mcordingley:
You're welcome!

frederikbosch:
Tag is there, happy converting!

sagikazarmark:
Thanks @mcordingley

Issue #328, opened 2016-11-07
Doctrine integration


sagikazarmark:
There are two possible options in my mind.

## Option 1

Same as in #327.

Use the money object itself as an embeddable.

``` yaml
Money\Money:
    type: embeddable
    fields:
        amount:
            type: string
    embedded:
        currency:
            class: Money\Currency

Money\Currency:
    type: embeddable
    fields:
        code:
            type: string
            length: 3
```

Pros:

- Simple
- Easy to integrate

Cons:

- Not really flexible

## Option 2

Use a separate embeddable.

``` php
<?php

namespace Money\Doctrine;

final class Money
{
    private $amount;
    private $currencyCode;

    /**
     * @return \Money\Money
     */
    public function getValue()
    {
        return new \Money\Money($this->amount, new \Money\Currency($this->currencyCode));
    }

    /**
     * @param \Money\Money $money
     */
    public function setValue(\Money\Money $money)
    {
        $this->amount = $money->getAmount();
        $this->currencyCode = $money->getCurrency()->getCode();
    }
}
```

And the mapping:

``` xml
<doctrine-mapping>
    <embeddable name="Money\Doctrine\Money">
        <field name="amount" type="string" />
        <field name="currencyCode" type="string" />
    </embeddable>
</doctrine-mapping>
```

Then in the entity:

``` php
class ObjectWithPrice
{
    /**
     * @var \Money\Doctrine\Money
     */
    private $price;

    public function getPrice()
    {
        return $this->price->getValue();
    }

    public function setPrice(\Money\Money $price)
    {
        $this->price->setValue($price);
    }
}
```

Since VO equality is not checked based on identity, but value this solution is valid. The advantage is the flexibility gained (less doctrine embed magic), but I am not sure it worth the extra effort.

@pamil I would love to hear your opinion as well.

frederikbosch:
The second option does not make any sense. If you ask me: 1. use the value object directly, 2. implement a doctrine interface/extend doctrine.

Example: http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/cookbook/custom-mapping-types.html.

sagikazarmark:
Doctrine type mapping is probably not an option since it allows you to have ONE column.

My concern is that option one includes too much Doctrine magic and reflection whereas in case of option too it's just scalars that it needs to take care of. I am not saying you are not right, just checking the options.

Padam87:
[...]

Issue #172, opened 2016-02-22
Different values in two servers with same php


rastydnb:
we have two server,the two servers have php 5.6.17 , one of server is mi local with nginx ,the other server is production with apache .

When use "add()" function over some moneys, get different result in two servers.

In local we get positive values and no get erros but,
in the producction server get negative values and get this error:

The first parameter of Money must be an integer. It's the amount, expressed in the smallest units of currency (eg cents)

We work with 6 decimals configured in config.yml
We need special php module, or any idea

Thanks

frederikbosch:
@rastydnb What version of Money are you using?Can you say which server is right, which is wrong? Which extensions are on the server that is right, which are on the server that is wrong?

When you work with 6 decimals, you could overflow PHP_MAX_INT. Unlimited integer support has been added in Money 3.0. You should realize that if you want to use Money with unlimited integer calculations, you need bc math or gmp to be installed as extension.

Probably one machine has one of these extensions while the other has not. Or you are Money 1.x. Final option: there is a bug ;)

rastydnb:
HI @frederikbosch thanks for the response. 
In the good server we have debian jessie 64bits nginx and php-fpm, and in the wrong server we have CentOS release 6.7 (Final) 32bits and apache without cgi ,the money version is 1.2.2.
The extension are the same in two servers

We think like you, and is PHP_MAX_INT where is the error

Is posible that the error was because the os is 32bits?

Thx

frederikbosch:
@rastydnb Yes that is possible. You can check PHP_INT_MAX from cli by running.

```sh
php -r "var_dump(PHP_INT_MAX);"
```

Compare that to your amount and you know if that is the cause. Upgrading to the 3.0 alpha release of Money should fix the problem, if you have either BC Math or GMP installed as extension.

rastydnb:
Thanks @frederikbosch

frederikbosch:
@rastydnb Yesterday we released a beta version. Please use that instead of the alpha version. We fixed one important bug in that version.

PR #389, opened 2017-08-03
Adding bigint disclaimer

programarivm:
Please refer to https://github.com/moneyphp/money/issues/328

### DIFF
--- a/doc/concept.rst
+++ b/doc/concept.rst
@@ -71,8 +71,10 @@ Because of PHP's integer limit, money values are stored as string internally and
 
 
 .. note::
-    Remember, because of the integer limit in PHP, you should inject a string that represents your huge amount.
-
+    Remember, because of the integer limit in PHP, you should inject a string that represents your
+    huge amount with a good reason behind. As a rule of thumb -- for security reasons -- it is recommended
+    to apply the least privilege principle when dealing with integers.
+    
 
 JSON
 ----


frederikbosch:
This is only acceptable with theoretical proof of the security limits. You did not provide any proof before. Only examples of vulnerabilities in libraries that provide unlimited integers usage. That is, however, not a proof of concept and therefore there is no ground to accept this PR.

programarivm:
I just posted this:

[The Least Privilege Principle Applied to PHP Bigints](https://programarivm.com/the-least-privilege-principle-applied-to-php-bigints/)

On the other hand, please refer to:

[When must arbitrary precision arithmetic functions be used in PHP?](https://softwareengineering.stackexchange.com/questions/127512/when-must-arbitrary-precision-arithmetic-functions-be-used-in-php)

IMO, a disclaimer should be added informing users that in order for Money to work OK, your server needs to rely on special libraries for running precision arithmetic functions, and the attack surface of your app may be affected because of that.

sagikazarmark:
@programarivm read your post and the discussion on the other thread.

I think there has been some misunderstanding, specifically about BIGINT != bigint and assuming that gmp/bcmath are mandatory.

The library uses string as the numeric representation, which allows, but does not enforce using BIGINT as the data type in your database.

The maximum numeric value which you can represent with Money is much bigger than PHP_MAX_INT thanks to this, but yeah, know your limits and use whichever data type makes sense in your persistence layer.

> in order for Money to work OK, your server needs to rely on special libraries for running precision arithmetic functions

this is not correct. those are completely optional, you don't have to use them if you don't want to.

generally I agree with the principle, but I don't agree with all your reasoning:

> What is a realistic max limit for money in my app?

> Does my app need to use currencies?

These questions should probably be asked in the context of an application. However, this is a generic purpose library which chose to use currencies and tries not to limit the maximum amount you can represent with it. We found that this is the way we can provide a library for the most common use cases, but it might not fit for everyone for whatever reasons and that's perfectly fine.

TBH I am not against adding a note to the documentation, but I think that it should already be obvious to choose the appropriate data type for your application.

frederikbosch:
Doctrine and database is not the responsibility of this library. And that BC math and GMP would open a window of opportunity for security vulnerabilities is just speculation. I did not read any proof for that statement.

Maybe - and I hope he does not care that I drag him into this thread - @paragonie-scott could give his opinion on this statement.

programarivm:
OK @frederikbosch  please don't merge yet the exact phrase originally suggested as it is written right now in this PR. Let's elaborate on it a bit further.

@sagikazarmark to my understanding the current Money library induce developers to infere that everything is OK with thinking in terms of BIGINT always -- because PHP is limited? -- and this could lead them to activate the BC and GMP functions.

Would you accept the above as an example of how violating the least privilage principle makes an app prone to human error? :)

Don't get me wrong, your idea is great, and it is OK to include BC and GMP functions if you know what you're doing, but I think the Money library is silently evangelizing a bad practice among the developers community by not raising awareness about how things work behind the scenes actually.

programarivm:
I don't agree with the following:

> However, this is a generic purpose library which chose to use currencies and tries not to limit the maximum amount you can represent with it.

IMO, Money should not be addressed to the general public including GMP and BC functions out of the box. But no problem. If it turns out that you want to manage huge numbers, knowing what you're doing, then extend the library with the GMP and BC features.

sagikazarmark:
> not raising awareness about how things work behind the scenes.

I can accept this point if you say that the documentation does not give enough explanation how the Calculator logic works under the hood. But I think it is:

http://moneyphp.org/en/latest/concept.html#integer-limit

> my understanding the current Money library induce developers to infere that everything is OK with thinking in terms of BIGINT always

Can you please point me to the docs or a single source which you think  encourages developers to ALWAYS think in terms of BIGINT? The possibility is given, the tools are given, that's it.

Or is it the problem that Money internally picks BC/GMP by default if it detects it?

programarivm:
Exactly, the magic shouldn't be like this:

1. BC Math (requires bcmath extension)
2. GMP (requires gmp extension)
3. Plain integer

According to the least privilege principle, we should simply have:

1. Plain integer

That's it.

Then, in order to run advanced, arbitrary precision arithmetic features, the API user would explicitly set the mode they wish to operate on -- GMP or BC Math. This way, if they decided to use GMP (explicitly) but it turns out that for whatever reason is off, the code will not run.

Does this makes any sense?

sagikazarmark:
Well, if you want to enforce that principle in the first place, you shouldn't install or enable those extensions.

Although there may be a slight performance penalty when having them enabled, but not using their capabilities, I don't think there is any more risk in having this convenience layer.

Last, but not least: changing this now would affect behaviour => BC break which would require a new major version.

Honestly, I am not really convinced by the points you raised. If you don't want to use GMP/BC then don't enable the extension. If you enabled the extension, but don't want to use it, you can still explicitly registering the integer calculator as the first one. This may change in the future, but I don't think it requires any actions from us now.

Thanks for raising these points anyway.

programarivm:
Thanks for letting me suggest how to make the library better.

> If you don't want to use GMP/BC then don't enable the extension. 

That was the point, it is you (Money's conceptual framework) actually encouraging to enable gmp and bcmath as shown in the following example.

- [Different values in two servers with same php](https://github.com/moneyphp/money/issues/172)

Can I address one question? I am really curious to know why we'd run arbitrary precision functions in common web apps. Where did this idea originate?

frederikbosch:
There was need a for unlimited integers in one of the issues. I decided to implement it using bc math and gmp, depending on the available extensions. And that is the final word to it. Case closed.

PR #390, opened 2017-08-04
Updated currencies

gmponos:


frederikbosch:
This cannot be accepted. We use this https://github.com/moneyphp/iso-currencies package as source for currencies, which in turn uses some official source. If you want to update iso currencies, update via that package.

gmponos:
Most probably I should have written that in the description but what I did was run the composer command that you have in the current package and commit the changes.

gmponos:
I am talking about [this command](https://github.com/moneyphp/money/blob/master/composer.json#L67) which uses the iso-currencies package.

frederikbosch:
Ah, then it is completely reasonable. I will check after weekend and merge it then.

PR #391, opened 2017-08-04
Add a class in order for the developer to instantiate his own Currencies.

gmponos:
Added a class SimpleCurrency in order for the developer to be able to add his own currencies.

**Edit:**

I don't mind if you believe that this does not belong in this package but I think that there should be a way for a developer to instantiate his own currencies. (He might need his own currencies for a Game App).


src/Currencies/SimpleCurrencies.php
```diff
@@ -0,0 +1,91 @@
...
+use Money\Currencies;
+use Money\Currency;
+use Money\Exception\UnknownCurrencyException;
+
+/**
+ * Cache the result of currency checking.
+ *
+ * @author George Mponos <gmponos@gmail.com>
+ */
+final class SimpleCurrencies implements Currencies
```
gmponos: Most probably you may disaggree with this name but I could not come with any better. Any thoughts

frederikbosch:
I am not necessarily against this. I'd rather call it ArrayCurrencies I guess. Not completely sure about the name.

gmponos:
I thought about the same name too. Let's leave it open until any better suggestion/feedback comes up.

> I am not necessarily against this.

Nice.. Then I will start writing some tests about it.

frederikbosch:
Please do :)

sagikazarmark:
CurrencyList works for me too

frederikbosch:
Agreed. Currency list would be even better!

gmponos:
> CurrencyList works for me too

> Agreed. Currency list would be even better!

OK. I am not renaming it yet. Maybe something else comes up. Rather than naming of the class do you believe you could review the rest?

gmponos:
> do you believe you could review the rest

Any news about this?

About the class naming I was thinking about it. All the other classes have a suffix `Currency` but the name `CurrencyList` has Currency as a prefix. I am a guy that likes consistency. So I would've suggest instead one of the following.

- ListCurrency
- ArrayCurrency
- SimpleCurrency
- CustomCurrency 

Let me know if there is any of the obove that you prefer or I should still go with `CurrencyList`.

sagikazarmark:
I think you are mixing things up a little: You implemented a currency repository.

https://github.com/moneyphp/money/tree/master/src/Currencies

We use the Currencies suffix there, but only to stay within the domain and avoiding naming, like `Repository`.

Even you originally called it `ArrayCurrencies`

So I would either find a name following that pattern or go with `CurrencyList` which is also somewhat okay within the domain.

gmponos:
🤦‍♂️ My mistake... What I wanted to say is to use the `Currencies` suffix and my suggestions are these:

- ListCurrencies
- ArrayCurrencies
- SimpleCurrencies
- CustomCurrencies

frederikbosch:
I'd say `ArrayCurrencies` or `CurrencyList`. @sagikazarmark Can you make the final decision? Then we can pull this in.

sagikazarmark:
I'm more in favor of `CurrencyList` at the moment.

frederikbosch:
Cool, @gmponos can you finish the PR? Then I will take care of the final review and merge it in.

gmponos:
> @gmponos can you finish the PR?

Done.

doc/features/currencies.rst
```diff
@@ -49,6 +49,36 @@ The Bitcoin currencies provides a single currency: the Bitcoin. It uses XBT as i
     $currencies->contains(new Currency('USD')); // returns boolean whether USD is available in this repository (false)
     $currencies->subunitFor(new Currency('XBT')); // returns the subunit for the Bitcoin (8)
 
+CurrencyList
+-----------------
```
sagikazarmark: Can you please fix length?

doc/features/currencies.rst
```diff
@@ -49,6 +49,36 @@ The Bitcoin currencies provides a single currency: the Bitcoin. It uses XBT as i
...
+
+The CurrencyList class provides a way for the developer to instantiate his own custom currencies. When initializing
```
sagikazarmark: I would rephrase this paragraph like this:

The CurrencyList class provides a way for a developer to build a custom currency repository.
The class accepts an associative array indexed by the currency code
and another associative array as the value containing `minorUnit` and `numericCode` keys and positive integer values.

spec/Currencies/CurrencyListSpec.php
```diff
@@ -0,0 +1,37 @@
...
+            'MY3' => ['minorUnit' => 1, 'numericCode' => 3],
+        ]);
+    }
+
+    function it_is_initializable()
+    {
+        $this->shouldHaveType(CurrencyList::class);
+    }
+
+    function it_contains_bitcoin()
```
sagikazarmark: Certainly not bitcoin

spec/Currencies/CurrencyListSpec.php
```diff
@@ -0,0 +1,37 @@
...
+    {
+        $this->contains(new Currency('MY1'))->shouldReturn(true);
+        $this->contains(new Currency('EUR'))->shouldReturn(false);
+    }
+
+    function it_is_iterable()
+    {
+        $this->getIterator()->shouldHaveCurrency('MY1');
+    }
+}
```
sagikazarmark: Missing new line

spec/Currencies/CurrencyListSpec.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace spec\Money\Currencies;
+
+use Money\Currencies\CurrencyList;
+use Money\Currency;
+use PhpSpec\ObjectBehavior;
+
+final class CurrencyListSpec extends ObjectBehavior
```
sagikazarmark: Looks like checking for the interface implementation is missing: https://github.com/moneyphp/money/blob/master/spec/Currencies/BitcoinCurrenciesSpec.php#L19

src/Currencies/CurrencyList.php
```diff
@@ -0,0 +1,100 @@
+<?php
+
+namespace Money\Currencies;
+
+use Money\Currencies;
+use Money\Currency;
+use Money\Exception\UnknownCurrencyException;
+
+/**
+ * An array currency list that can be initialized with custom currencies.
```
sagikazarmark: A list of custom currencies.

src/Currencies/CurrencyList.php
```diff
@@ -0,0 +1,100 @@
...
+    public function __construct(array $currencies)
+    {
+        foreach ($currencies as $currencyCode => $currency) {
+            if (empty($currencyCode) || !is_string($currencyCode)) {
+                throw new \InvalidArgumentException(
+                    sprintf('Currency code must be a string and not empty. "%" given', $currencyCode)
+                );
+            }
+
+            if (!isset($currency['minorUnit']) || !is_int($currency['minorUnit']) || $currency['minorUnit'] < 0) {
```
sagikazarmark: This could be a multiline if for readability.

src/Currencies/CurrencyList.php
```diff
@@ -0,0 +1,100 @@
...
+                throw new \InvalidArgumentException(
+                    sprintf('Currency %s does not have a valid minorUnit. Must be a positive integer.', $currencyCode)
+                );
+            }
+
+            if (!isset($currency['numericCode']) || !is_int($currency['numericCode']) || $currency['numericCode'] < 0) {
```
sagikazarmark: This could be a multiline if for readability.

doc/features/currencies.rst
gmponos: What do you mean? Does it need to be the same lenth as CurrencyList So it means 13 -?

sagikazarmark: > Does it need to be the same lenth as CurrencyList

Yup

sagikazarmark: > Does it need to be the same lenth as CurrencyList

Yup

spec/Currencies/CurrencyListSpec.php
gmponos: Done

spec/Currencies/CurrencyListSpec.php
gmponos: Done.

doc/features/currencies.rst
gmponos: Done

doc/features/currencies.rst
gmponos: Done

spec/Currencies/CurrencyListSpec.php
gmponos: Done

frederikbosch:
Thanks for reviewing @sagikazarmark. I think this is ready to go. Merge and tag?

sagikazarmark:
Will take a look at it tomorrow. I would like to slow down a bit on tags
and not release every time a new feature is added. We kinda abuse patch
releases.

frederikbosch:
@sagikazarmark Disagree there. I find it always extremely annoying to wait for release with other libraries, especially when it is a dependency of a dependency in a project. Then you will have to change your root composer.json to use `minimum-stability` and `prefer-stable`, which must be avoided at any cost. So I'd rather tag on every feature for my other libraries.

sagikazarmark:
> Then you will have to change your root composer.json to use minimum-stability and prefer-stable

Preferably you maintain a fork, until your feature is merged upstream and I see no problem with that, given that the project is actively maintained.

Currently we have a few patch releases containing mainly new features, not actual patches which is kind of abuse of semver.

Tagging a new release every time a new, mostly insignificant (no offense contributors) feature is added does not give us and users enough time to take a look at the changes, iterate whenever necessary, make them more mature. And TBH, in case of a feature like this I see no problem with just copying the code to the project and using it from there until the new version is released. If done right, it should not be more work than renaming the class in one or two places, which is what we keep the massive beasts called IDEs eating up all our RAM for.

I understand that keeping up the speed is cool, but that's not how stability and maturity are achieved, and in a project like this that's priority one.

gmponos:
I thought that you don't mind if I also state my opinion since I am on this discussion but I am not a maintainer.

> Currently we have a few patch releases containing mainly new features, not actual patches which is kind of abuse of semver.

Are you saying that you are going from 3.0.0 to 3.0.1 and that 3.0.1 version contains new features? Why not just go to 3.1.0?

> Tagging a new release every time a new, mostly insignificant (no offense contributors) feature is added does not give us and users enough time to take a look at the changes, iterate whenever necessary, make them more mature. 

I aggree on this one but most importantly I believe this should be handled differently per case.

But I have to note a few things more.

1. @sagikazarmark  since you are also a maintainer on Guzzle if I am not mistaken the new tag of guzzle took almost a year to be tagged. I hope this doesn't take so long.
2. Please tag before dropping any  PHP version.

frederikbosch:
> Tagging a new release every time a new, mostly insignificant (no offense contributors) feature is added does not give us and users enough time to take a look at the changes, iterate whenever necessary, make them more mature.

This is not a problem of tagging. That is a problem of pulling/merging to master. When merged to master, tagging must be always possible. If we want to have a longer review period, we must wait longer with merging to master. And in certain cases, that is justifiable, but in others it is not. We have unit and spec tests, and we have the human eye. That must be enough for small changes.

In other cases, when it involves many code lines, I agree that longer review periods are necessary. I guess I am too hasty on those reviews. My attitude is indeed to keep things going and not wait too long. This generates a window of opportunity for stability and maturity problems. I will try to improve here. Maybe we should make a distinction in PRs: (1) requires team review (2) requires single reviewer.

Nonetheless, I really think we should tag after merging to master. If we think changes are stable and merged to master, then we can directly open the code to the public. Then regarding the tag number. I think we should follow sem ver. Maybe I should have created `3.1.0` earlier on, but I also think that is debatable. Did we really add features? In my point of view we did not. We added implementations of features we already got. So in that light, I think the last number should be increased, as we/I did.

sagikazarmark:
> When merged to master, tagging must be always possible.

Not necessarily. You can branch off from the latest stable tag, do some changes, create a patch release, and merge back to master.

I tagged this PR with 3.1 release milestone. There are a few other new features proposed, so let's finish them as well and release them together as 3.1

src/Currencies/CurrencyList.php
```diff
@@ -0,0 +1,108 @@
...
+     * @var array
+     */
+    private $currencies;
+
+    public function __construct(array $currencies)
+    {
+        foreach ($currencies as $currencyCode => $currency) {
+            if (empty($currencyCode) || !is_string($currencyCode)) {
+                throw new \InvalidArgumentException(
+                    sprintf('Currency code must be a string and not empty. "%" given', $currencyCode)
```
gmponos: I need to fix the missing `s` at `%s` before merging this.

gmponos: Fixed

gmponos:
Inside the CurrencyList class I have added a function to get the CurrencyCode.

I've added this function because I thought it was required by the interface `Currencies` but it seems it is not.

I believe that it should be removed.

doc/features/currencies.rst
```diff
@@ -49,6 +49,35 @@ The Bitcoin currencies provides a single currency: the Bitcoin. It uses XBT as i
     $currencies->contains(new Currency('USD')); // returns boolean whether USD is available in this repository (false)
     $currencies->subunitFor(new Currency('XBT')); // returns the subunit for the Bitcoin (8)
 
+CurrencyList
+-------------
```
sagikazarmark: Looks like there is one extra dash here. ;)

doc/features/currencies.rst
```diff
@@ -49,6 +49,35 @@ The Bitcoin currencies provides a single currency: the Bitcoin. It uses XBT as i
...
+
+The CurrencyList class provides a way for a developer to build a custom currency repository.
```
sagikazarmark: provides a way for developers to create

doc/features/currencies.rst
```diff
@@ -49,6 +49,35 @@ The Bitcoin currencies provides a single currency: the Bitcoin. It uses XBT as i
...
+string. Also `minorUnit` keys must exists and have a positive integer value.
+
+.. code-block:: php
+
+    use Money\Currencies\CurrencyList;
+    use Money\Currency;
+
+    $currencies = new CurrencyList([
+        'MY1' => [
+            'minorUnit' => 2,
```
sagikazarmark: I wonder if we really need this inner associative array. I guess it's there for future options, but for now it's harder to document and understand IMO. Maybe just accept string => integer pairs? We can always add support in a BC way for other types.

WDYT?

doc/features/currencies.rst
```diff
@@ -49,6 +49,35 @@ The Bitcoin currencies provides a single currency: the Bitcoin. It uses XBT as i
...
     $currencies->subunitFor(new Currency('XBT')); // returns the subunit for the Bitcoin (8)
 
+CurrencyList
+-------------
+
+The CurrencyList class provides a way for a developer to build a custom currency repository.
+The class accepts an associative array indexed by the currency code
+and another associative array as the value containing `minorUnit` key and positive integer values.
+
+More specifically it must be an associative array. A currency code key must exists for all currencies and it must be a
```
sagikazarmark: I think this part is not necessary. Just mention in the previous section that an invalid array would end up in an exception being thrown.

doc/features/currencies.rst
```diff
@@ -49,6 +49,35 @@ The Bitcoin currencies provides a single currency: the Bitcoin. It uses XBT as i
...
+        'MY1' => [
+            'minorUnit' => 2,
+        ],
+    ]);
+
+    foreach ($currencies as $currency) {
+        echo $currency->getCode(); // prints MY1
+    }
+
+    $currencies->contains(new Currency('MY1')); // returns boolean whether XBT is available in this repository (true)
```
sagikazarmark: XBT?

spec/Currencies/CurrencyListSpec.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        $this->shouldHaveType(CurrencyList::class);
+    }
+
+    function it_contains_custom_currency()
+    {
+        $this->contains(new Currency('MY1'))->shouldReturn(true);
+    }
+
+    function it_does_not_contains_currency()
```
sagikazarmark: does_not_contain

src/Currencies/CurrencyList.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        foreach ($currencies as $currencyCode => $currency) {
+            if (empty($currencyCode) || !is_string($currencyCode)) {
+                throw new \InvalidArgumentException(
+                    sprintf('Currency code must be a string and not empty. "%s" given', $currencyCode)
+                );
+            }
+
+            if (
+                !isset($currency['minorUnit']) ||
```
sagikazarmark: See my eariler comment regarding this.

sagikazarmark:
@gmponos thanks for all the hard work. I did one, last, thorough review, so once those are resolved I don't think any more changes are necessary to get this merged.

There is one which might be subject of discussion, but I think we can sort it out quickly.

doc/features/currencies.rst
gmponos: Done.

doc/features/currencies.rst
gmponos: Done.

doc/features/currencies.rst
gmponos: I believe it's fine now. Check it out :)

doc/features/currencies.rst
gmponos: I agree... I believe it's fine now. Check it out :)

doc/features/currencies.rst
gmponos: Done

spec/Currencies/CurrencyListSpec.php
gmponos: Done

src/Currencies/CurrencyList.php
gmponos: I believe it's fine now. Check it out :)

gmponos:
I am not sure why styleci requires from me changes at classes that I have not changed. 😕

frederikbosch:
Thanks for everyone's hard work in here.

sagikazarmark:
Thanks a lot @gmponos for your work and your patience. :)

PR #396, opened 2017-08-22
Add tests for Aggregator constructors

gmponos:
According to the tests you can see here is this the expected behavior?

### DIFF
--- a/src/Formatter/AggregateMoneyFormatter.php
+++ b/src/Formatter/AggregateMoneyFormatter.php
@@ -23,7 +23,7 @@ final class AggregateMoneyFormatter implements MoneyFormatter
      */
     public function __construct(array $formatters)
     {
-        if (empty($parsers)) {
+        if (empty($formatters)) {
             throw new \InvalidArgumentException(sprintf('Initialize an empty %s is not possible', self::class));
         }
 
--- a/src/Formatter/AggregateMoneyFormatter.php
+++ b/src/Formatter/AggregateMoneyFormatter.php
@@ -23,6 +23,10 @@ final class AggregateMoneyFormatter implements MoneyFormatter
      */
     public function __construct(array $formatters)
     {
+        if (empty($parsers)) {
+            throw new \InvalidArgumentException(sprintf('Initialize an empty %s is not possible', self::class));
+        }
+
         foreach ($formatters as $currencyCode => $formatter) {
             if (false === $formatter instanceof MoneyFormatter) {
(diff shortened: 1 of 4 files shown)


frederikbosch:
@gmponos Obviously not. Passing in an empty array means 100% sure to have a failing parser/formatter object. Therefore I would not find it breaking to throw a `InvalidArgumentException` on such occasion.

gmponos:
OK... i'll fix this

src/Formatter/AggregateMoneyFormatter.php
```diff
@@ -23,6 +23,10 @@
      */
     public function __construct(array $formatters)
     {
+        if (empty($parsers)) {
```
slavcodev: Do you mean `empty($formatters)`?

gmponos: pfff.. copy-paste!

src/Formatter/AggregateMoneyFormatter.php
```diff
@@ -23,6 +23,10 @@
      */
     public function __construct(array $formatters)
     {
+        if (empty($formatters)) {
+            throw new \InvalidArgumentException(sprintf('Initialize an empty %s is not possible', self::class));
```
frederikbosch: Can you change `self::class` in `get_class($this)`? We are supporting PHP 5.5. Which I do not like, but until we are dropping it, we have to take care of it.

src/Parser/AggregateMoneyParser.php
```diff
@@ -22,6 +22,10 @@
      */
     public function __construct(array $parsers)
     {
+        if (empty($parsers)) {
+            throw new \InvalidArgumentException(sprintf('Initialize an empty %s is not possible', self::class));
```
frederikbosch: Can you change `self::class` in `get_class($this)`? We are supporting PHP 5.5. Which I do not like, but until we are dropping it, we have to take care of it.

src/Formatter/AggregateMoneyFormatter.php
gmponos: I am not sure I understand.

http://php.net/manual/en/language.oop5.constants.php

> The special ::class constant are available as of PHP 5.5.0, and allows for fully qualified class name resolution at compile, this is useful for namespaced classes: 

Also the tests are passing for PHP 5.5 also

frederikbosch: Oh right. Damn. My bad.

PR #397, opened 2017-08-26
Allow Parsers to take Currency object as the forceCurrency parameter

### LINKED ISSUE #386: Parser doesn't accept currency object
Are you willing to accept a pull request to make parser objects accept currency objects in addition to strings?

e.g. this currently throws an error

    $moneyParser = new DecimalMoneyParser($currencies);
    $c = new Currency('USD');
    $moneyParser->parse('1.00', $c);

It really should accept the currency object and not force me to write `$c->getCode()` in some places and just `$c` in others.

jarobe42:
Closes #386 

This adds support for $forceCurrency to either be a string or a Currency object.

If you'd rather wait until the next major version and just use the Currency object, let me know and I'll make changes.

### DIFF
--- a/src/MoneyParser.php
+++ b/src/MoneyParser.php
@@ -12,8 +12,8 @@ interface MoneyParser
     /**
      * Parses a string into a Money object (including currency).
      *
-     * @param string      $money
-     * @param string|null $forceCurrency
+     * @param string     $money
+     * @param mixed|null $forceCurrency
      *
      * @return Money
      *
--- a/src/MoneyParser.php
+++ b/src/MoneyParser.php
@@ -12,8 +12,8 @@ interface MoneyParser
     /**
      * Parses a string into a Money object (including currency).
      *
-     * @param string     $money
-     * @param mixed|null $forceCurrency
+     * @param string               $money
+     * @param Currency|string|null $forceCurrency
      *
      * @return Money
      *
--- a/src/Parser/DecimalMoneyParser.php
+++ b/src/Parser/DecimalMoneyParser.php
@@ -46,7 +46,15 @@ final class DecimalMoneyParser implements MoneyParser
             );
         }
 
-        $currency = new Currency($forceCurrency);
+        /*
+         * This conversion is only required whilst currency can be either a string or a
+         * Currency object.
+         */
+        $currency = $forceCurrency;
+        if (!$currency instanceof Currency) {
+            $currency = new Currency($currency);
+        }
+
         $decimal = trim($money);
 
         if ($decimal === '') {
--- a/src/Parser/DecimalMoneyParser.php
+++ b/src/Parser/DecimalMoneyParser.php
@@ -52,6 +52,7 @@ final class DecimalMoneyParser implements MoneyParser
          */
         $currency = $forceCurrency;
         if (!$currency instanceof Currency) {
+            @trigger_error('Passing a currency as string is deprecated since 3.1 and will be removed in 4.0. Please pass a '.Currency::class.' instance instead.', E_USER_DEPRECATED);
             $currency = new Currency($currency);
         }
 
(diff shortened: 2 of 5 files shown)


src/MoneyParser.php
```diff
@@ -12,8 +12,8 @@
     /**
      * Parses a string into a Money object (including currency).
      *
-     * @param string      $money
-     * @param string|null $forceCurrency
+     * @param string     $money
+     * @param mixed|null $forceCurrency
```
sagikazarmark: I would like to see Currency|string|null instead here. Better to be explicit about return types.

src/Parser/DecimalMoneyParser.php
```diff
@@ -46,7 +46,15 @@ public function parse($money, $forceCurrency = null)
             );
         }
 
-        $currency = new Currency($forceCurrency);
+        /*
+         * This conversion is only required whilst currency can be either a string or a
+         * Currency object.
+         */
+        $currency = $forceCurrency;
+        if (!$currency instanceof Currency) {
```
sagikazarmark: I believe we should deprecate passing a string here and remove that in 4.0

So I propose triggering a deprecation notice here:

``` php
@trigger_error('Passing a currency as string is deprecated since 3.1 and will be removed in 4.0. Please pass a '.Currency::class.' instance.', E_USER_DEPRECATED);
```

src/Parser/IntlMoneyParser.php
```diff
@@ -45,23 +45,30 @@ public function parse($money, $forceCurrency = null)
...
-        $decimal = (string) $decimal;
+        if (null !== $forceCurrency) {
+            $currency = $forceCurrency;
+        }
+
+        /*
+         * This conversion is only required whilst currency can be either a string or a
+         * Currency object.
+         */
+        if (!$currency instanceof Currency) {
```
sagikazarmark: Same deprecation could go here.

src/Parser/IntlMoneyParser.php
```diff
@@ -45,23 +45,30 @@ public function parse($money, $forceCurrency = null)
...
+            $currency = new Currency($currency);
+        }
+
         $subunit = $this->currencies->subunitFor($currency);
+
+        $decimal = (string) $decimal;
```
sagikazarmark: This doesn't seem to be related to the PR. Can you please explain?

src/MoneyParser.php
jarobe42: Good catch. I've updated this.

src/Parser/DecimalMoneyParser.php
jarobe42: Good idea. I'll add that in.

src/Parser/IntlMoneyParser.php
jarobe42: Nothing changed here - I swapped the order of these two following lines, as it made more sense to me that $decimal would be used immediately after being cast;
`$decimal = (string) $decimal;` and
`$subunit = $this->currencies->subunitFor($currency);`

There's no real benefit to doing this though, so I'll flip it back around so as not to add confusion in the git history.

jarobe42:
Thanks for the feedback, @sagikazarmark. I've made the discussed changes, let me know if there's anything else!

src/Parser/DecimalMoneyParser.php
frederikbosch: My instinct tells me that is a bad idea. I cannot recall the last time I used `trigger_error`. I rather use the docblock `@deprecated`, which could give visual feedback to developers if their IDE supports it.

jarobe42: `trigger_error` is something I haven't used much lately either, but unless I'm mistaken, there'd be no way to indicate that the use of `parse` is only deprecated when $currency is parsed as a string. I'd be concerned that showing a deprecation notice for all calls of `parse`, including those that pass the correct `Currency` parameter, would be more confusing.

sagikazarmark: This is widely used in the PHP ecosystem. Triggered errors will immediately be bypassed, so no BC break.

@jarobe42 you can and should trigger errors within the if block, only for those calls which received a string, not a currency.

sagikazarmark:
Thanks @jarobe42 a lot, this will be part of the 3.1 release which will happen in the next few weeks.

PR #400, opened 2017-09-06
Modulus support

jaikdean:
This PR adds `Money::mod()` to allow modulus operations.

We have a use-case for this whereby we need to transfer as large a value as possible from a user's balance in £5 increments. Currently this requires either looping, or arithmetic with raw amounts outside of the `Money` class.

With this PR, we can do:

```php
$balance = Money::GBP(1820); // £18.20
$incrementSize = Money::GBP(500); // £5.00

$newBalance = $balance->mod($incrementSize); // £3.20
$transferredBalance = $balance->subtract($newBalance); // £15.00
```

### DIFF
--- a/tests/Calculator/CalculatorTestCase.php
+++ b/tests/Calculator/CalculatorTestCase.php
@@ -215,6 +215,9 @@ abstract class CalculatorTestCase extends \PHPUnit_Framework_TestCase
             [9, 3, '0'],
             [1006, 10, '6'],
             [1007, 10, '7'],
+            [-13, -5, '-3'],
+            [-13, 5, '-3'],
+            [13, -5, '3'],
         ];
     }
 }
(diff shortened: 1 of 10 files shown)


tests/Calculator/CalculatorTestCase.php
```diff
@@ -198,4 +207,14 @@ public function compareExamples()
             ['1', '0.000000000000000000000000005', 1],
         ];
     }
+
+    public function modExamples()
+    {
+        return [
```
sagikazarmark: Can you please add a few tests for negative numbers as well?

sagikazarmark:
Thanks a lot for your PR @jaikdean , sorry for the late feedback. I like the idea, kudos for the really high quality PR.

I had one minor comment regarding the number of provided tests, I would like to see a bit more examples.

Apart from that, I have nothing to say, but thanks (and hit the merge button ;) )

tests/Calculator/CalculatorTestCase.php
jaikdean: Good spot, I've added some and they flagged up a bug in the `GmpCalculator::mod()` implementation, which I've now fixed.

jaikdean:
@sagikazarmark Does anything else need doing on this?

frederikbosch:
Thanks for the great effort here!

PR #402, opened 2017-09-14
Fix comparison operation

flajos:
The scale (precision) parameter was missing from the comparison operation, leading to wrong results.


sagikazarmark:
Thanks @flajos for your contribution, sorry for the late feedback.

As far as I can see the builds are breaking because of the documentation, but that has been fixed in https://github.com/moneyphp/money/commit/1fc94fcc55ae3b534932d8cdab33bc591e4c3e24, so I'm just going to merge your fix.

flajos:
Thanks @sagikazarmark!
Nice work by the way :+1:

Issue #403, opened 2017-09-15
Lost minus if rounded -9999


unglud:
Hi!
```
Number::roundMoneyValue("-9999", 2, 3); // 10009, expected '-10009'
Number::roundMoneyValue("-8999", 2, 3) // -9009, correct
```
In one point when we have $moneyValue '-0009' and we want to put '1' between '-' and '0', we remove '-' completely

frederikbosch:
See latest release.

PR #404, opened 2017-09-15
Add check for lost minus

unglud:
There is not most beautiful but workable fix for #403


frederikbosch:
@sagikazarmark I think we should merge this and tag. It is a bug that should be fixed asap.

sagikazarmark:
@frederikbosch can you please take care of changelog and the release?

PR #409, opened 2017-10-12
Add some typehints for static convenience methods.

willemstuursma:
If you like this, I can create some script to auto generate all static convenience methods.

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -10,6 +10,9 @@ use Money\Calculator\PhpCalculator;
  * Money Value Object.
  *
  * @author Mathias Verraes
+ *
+ * @method static Money EUR(int $amount) Create a new Money object of amount $amount in currency EUR.
+ * @method static Money USD(int $amount) Create a new Money object of amount $amount in currency USD.
  */
 final class Money implements \JsonSerializable
 {


sagikazarmark:
Does this thing work if you simply add them in a php file (not above the Money class)?

This request has been rejected previously, mainly because it pollutes the Money class and makes the file unreadable. If it works, I would even be happy to auto generate it just like we do with our currency list.

willemstuursma:
No sorry, they have to go on the class file for PhpStorm to understand it. They could also go on a parent class but than, logically, you would have to move the `__callStatic` method too.

I'm not sure what you think is unreadable? It would need one line in the PhpDoc per factory function.

An alternative would be to drop the `__callStatic` and get proper static methods.

sagikazarmark:
We had similar requests earlier with more than two proposed factory methods (aka. all of them). That's what I mean when I say unreadable.

I'm afraid if we start going down on this path it will never end.

I will think about it and look around for possible solutions.

Gummibeer:
But PhpDoc Tags don't make code unreadable!? They have a different styling, could be collapsed and are helpful to identify the start of a new class/method. And they also help the developer to find the right method and what he has to pass and what he get in return.
So I have to support a good code documentation. Yes, it will make the file longer but not unreadable.

willemstuursma:
I agree with @Gummibeer. 

As this is a library, the code won't be browsed a lot by most people using it but the syntax completion will be used a lot.

sagikazarmark:
Found it finally: #92 I would really like to avoid something like this.

I could accept saying have the top ten currencies added as typehint, but no more, because:

- Money is not limited to ISO currencies
- ISO currency list is generated (How do we maintain generated list within the class)
- number of currencies

I would really be happier with some other solution where we don't have to maintain a dynamic list within the source code. For example some generated file with doc blocks or whatever IDEs can read.

Alternatively we could generate an `ISO` or whatever class which contains a factory for all the currencies.

tl;dr This solution should be our last shot at the problem.

Gummibeer:
I agree - doing this by hand will end in chaos. If the DocBlock isn't this easy to generate a package that does something similar is https://github.com/barryvdh/laravel-ide-helper. PhpStorm also understands something like.

**src/Money.php**
```php
class Money {
 // __callstatic
}
```

**doc/Money.php** (autogenerated)
```php
class Money {
    public static function euro() {
        return 'EUR';
    }
}
```

It will have a small hint that this class exists two times but autocomplete and so will work. Cause of the fact that this file isn't loaded by composer there is no problem on runtime. The only "problem" is how to generate this file automaticly. And I think it would be enough if this repo ships a `generate_money_doc.php` file that will do the job on the local machine and don't have it in repo.

Gummibeer:
I have thought about it and have some pseudocode:

```php
$currenciesClasses = [
  \Money\Currencies\ISOCurrencies::class,
];

echo '<?php'.PHP_EOL;
echo 'namespace Money;'.PHP_EOL;
echo 'final class Money implements \JsonSerializable {'.PHP_EOL;
foreach($currenciesClasses as $currencies) {
  foreach(currencies as $currency) {
    $code = $currency->getCode();
    echo 'public static function '.$code.'($amount) { return new Currency(\''.$code.'\') }'.PHP_EOL;
  }
}
echo '}'.PHP_EOL;
```

willemstuursma:
Calling in @shirshir who had another solution.

shirshir:
You should have code completion in PhpStorm by creating a helper [file with meta data](https://confluence.jetbrains.com/display/PhpStorm/PhpStorm+Advanced+Metadata) in the root of your project:

```php
<?php

namespace PHPSTORM_META {

    use Money\Money;

    override(Money::EUR(0),
        map([
            '' => Money::class,
        ])
    );
}
```

Unfortunately this does not work, the new syntax (for example as generated by the Laravel ide helper that @Gummibeer suggested, like [this](https://gist.github.com/barryvdh/bb6ffc5d11e0a75dba67)) does not work for the magic static factory methods. 

I got it to work (for example with `EUR`) with the following helper file contents (the old syntax):

```php
<?php

namespace PHPSTORM_META {

    $STATIC_METHOD_TYPES = [
        \Money\Money::EUR('') => [
            '' == '\Money\Money',
        ],
    ];

}
```
Note that this still marks the `EUR` method as undefined, but at least code completion works 😄. You may need to restart PhpStorm to refresh the meta data.

image

Gummibeer:
@shirshir Which version of phpstorm do you use? The Code generated by by my snippet shows autocomplete and works except of the point that it shows that the class ist defined two times.
And the benefit is that it also works in other IDEs cause it creates the class with all methods in normal php style.

shirshir:
@Gummibeer I am using `2017.2.4`. Thanks for your suggestion. You could also include your generated file with a local [customised version of the PhpStorm stubs](https://stackoverflow.com/a/41555186). This has the advantage that you can re-use the generated stub in multiple projects. Still, PhpStorm will think that two versions of the class exist, one in the `vendor` directory, and one in the stubs directory. But that is a small tradeoff. 🙂

sagikazarmark:
> file that will do the job on the local machine and don't have it in repo.

Since this is just for development that could work, but I'm not against having this in the repo either if it works.

@shirshir thanks for your suggestion, I was hoping that there was some kind of official way to do this in PHPstorm, but it seems like we can't use it then. But I would love to revisit this if PHPStorm adds support for static methods. (Perhaps we could open an issue in Jetbrains' issue tracker).

@Gummibeer would you like to open a pull request?

(Hint: we usually put these stuff into composer scripts as well to make running them easier).

We can discuss in a PR whether we want to maintain the list in the repo or not.

Gummibeer:
I will do. But adding a phpstorm only version would be a nice addition but not as the only one. Cause it will disable the support for Atom, Sublime and whatever all the Editors are named.

So I will add a pollished script like the one I posted and make it available via composer.

sagikazarmark:
> But adding a phpstorm only version would be a nice addition but not as the only one.

Indeed, for now we can just go with this version and think about phpstorm later.

> So I will add a pollished script like the one I posted and make it available via composer.

Great, thanks a lot.

sagikazarmark:
Closing this then.

Gummibeer:
PR #415 is open - it works pretty good but I'm open for improvements.

Issue #408, opened 2017-10-11
Support for parsing scientific notation.


bendavies:
Hey there,

So i hit upon a bug in my API recently, where i'm accepting values of money using symfony's [MoneyType](https://symfony.com/doc/current/reference/forms/types/money.html).

Now, during some testing i managed to trigger an error in Money by submiting a large value to my API, specifically:
```
Uncaught PHP Exception InvalidArgumentException: 
"Invalid fractional part 0E+17. Invalid digit E found" at /.../vendor/moneyphp/money/src/Number.php line 219
```

What is happing here is:
1. I submit a value of `1000000000000000` to my API.
2. That is transformed to a float `1.0E+15` by [NumberToLocalizedStringTransformer](https://github.com/symfony/form/blob/7fdcf54a3a754f40dcea1d9851eac65adf6f8f97/Extension/Core/DataTransformer/NumberToLocalizedStringTransformer.php#L139) (and then `1.0E+17` by [MoneyToLocalizedStringTransformer](https://github.com/symfony/form/blob/7fdcf54a3a754f40dcea1d9851eac65adf6f8f97/Extension/Core/DataTransformer/MoneyToLocalizedStringTransformer.php#L80).
3. `1.0E+17` is given to the Money constructor, and fails in [Number](https://github.com/moneyphp/money/blob/149ba5943511e118bf68f9118e3c84bbe42bf94f/src/Number.php) as above.

So the questions are:
1. Does anyone know how I can get this number though symfony forms without being converted to float in scientific notation.  
2. Can `Number` add support for scientific notation?

Cheers

bendavies:
How about we convert the float back to a string with sprintf?
https://3v4l.org/H0FVj

unglud:
I guess you can try to parse value from symfony object before you put it to money. 
https://github.com/moneyphp/money/issues/406#issuecomment-332131006

bendavies:
I have fixed it in my forms MoneyTransformer at the moment:
`return new Money(\sprintf('%.0f', $value['amount']), new Currency($value['currency']));`

bendavies:
I've just noticed `Number::fromFloat`, so my `Money` is now constructed as
`return new Money(Number::fromFloat($value['amount']), new Currency($value['currency']));`

This can probably be closed.

PR #410, opened 2017-10-17
support instantiating money from float

bendavies:
This is mainly in response to https://github.com/moneyphp/money/issues/408

Any thoughts?

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -49,20 +49,24 @@ final class Money implements \JsonSerializable
     ];
 
     /**
-     * @param int|string $amount   Amount, expressed in the smallest units of $currency (eg cents)
-     * @param Currency   $currency
+     * @param float|int|string $amount   Amount, expressed in the smallest units of $currency (eg cents)
+     * @param Currency         $currency
      *
      * @throws \InvalidArgumentException If amount is not integer
      */
     public function __construct($amount, Currency $currency)
     {
         if (filter_var($amount, FILTER_VALIDATE_INT) === false) {
-            $numberFromString = Number::fromString($amount);
-            if (!$numberFromString->isInteger()) {
+            if (is_float($amount)) {
+                $number = Number::fromFloat($amount);
+            } else {
+                $number = Number::fromString($amount);
+            }
+            if (!$number->isInteger()) {
                 throw new \InvalidArgumentException('Amount must be an integer(ish) value');
             }
 
-            $amount = $numberFromString->getIntegerPart();
+            $amount = $number->getIntegerPart();
         }
 
         $this->amount = (string) $amount;
@@ -91,7 +95,7 @@ final class Money implements \JsonSerializable
     /**
      * Returns a new Money instance based on the current one using the Currency.
      *
-     * @param int|string $amount
+     * @param float|int|string $amount
      *
      * @return Money
      *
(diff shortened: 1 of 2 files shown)


src/Number.php
```diff
@@ -117,7 +117,7 @@ public static function fromString($number)
     /**
      * @param float $floatingPoint
      *
-     * @return Number
+     * @return self
```
sagikazarmark: I tend to prefer using the class name in return type hints to indicate a new object is going to be returned. At least for consistency, this should be reverted.

src/Money.php
```diff
@@ -49,20 +49,24 @@
...
+     * @param Currency         $currency
      *
      * @throws \InvalidArgumentException If amount is not integer
      */
     public function __construct($amount, Currency $currency)
     {
         if (filter_var($amount, FILTER_VALIDATE_INT) === false) {
-            $numberFromString = Number::fromString($amount);
-            if (!$numberFromString->isInteger()) {
+            if (is_float($amount)) {
```
sagikazarmark: Not sure this will work as intended. What if `$amount == (float)PHP_INT_MAX+1` ?

src/Number.php
bendavies: i was making consistent with `fromString` above. i'll revert this and you can handle in another PR if you wish.

src/Money.php
bendavies: this works fine.

bendavies: sorry, i misread. yes this doesn't work. what should be the behaviour/error here?

bendavies:
@frederikbosch there are two tests, one from a float, and one from scientific notation. (and i was just testing the maximum float representable by php fyi).

btw, why is there `MoneySpec` and `MoneyTest`? There seems to be some overlap.

frederikbosch:
Well I am simply against accepting one more type in the constructor. Use a named public static constructor it.

bendavies:
@frederikbosch sorry, i missed what you were objecting to, hence my non nonsense reply.

sagikazarmark:
I second @frederikbosch 's opinion. Would `Money::fromNumber(Number::fromFloat(3.5))` work for you @bendavies ?

bendavies:
@sagikazarmark hey! that doesn't add anything to me over my current solution which is:`new Money(Number::fromFloat(3.5), new Currency('USD');`

The point of this PR was to provide a constructor that can take a string, int or float, so I don't have to do any checking up front.

I'm not particularly attached to the idea though, so no problem if you guys don't like it!

sagikazarmark:
Hm, I see. I'll leave this decision to @frederikbosch

frederikbosch:
@bendavies I get your point, but I stick to my main point: I do not want to have a constructor that becomes a bag of mixed types.

What could be the solution that fits both viewpoints.

```php
Money::fromFloat(3.5);

public static function fromFloat($amount)
{
  if (!is_float($amount)) {
    throw new \InvalidArgumentException();
  }

  return new self((string)Number::fromFloat($amount));
}
```

frederikbosch:
This issue has too many problems. If there is a float, say `3.5`, should that become 3 or 4? Or should we use a rounding mode? But since the method will be static, we have to change more. And since I am totally against making the constructor a bag mixed types, I am going to close it. This is possible, but requires a more comprehensive solution than the current proposition.

PR #411, opened 2017-10-24
Escaping of bitcoin symbol is wrong

willemstuursma:
`\` escapes the `0` character, injecting a null byte (`0x00`). It should escape the `x` so that the right UTF-8 multibyte character is injected.


sagikazarmark:
LOL, how the hell did we miss this?

frederikbosch:
@sagikazarmark There was a request to add implementation for Bitcoin. I just created it and fulfilled the request. The person that needed it never replied on the code. I think it _looked_ fine from our point of view. However, if you are not using code that you have written, I think there is a higher chance that it contains bugs. Which is the case here. I think we can merge it in.

sagikazarmark:
@frederikbosch I didn't mean to blame, no need for excuses.

sagikazarmark:
Patch released, thanks @willemstuursma

PR #416, opened 2017-10-26
Add a comparator for use in PHPUnit tests

willemstuursma:
Adds a comparator for use in PHPUnit so you can use assertEquals() on Money objects.


frederikbosch:
@willemstuursma Never knew this PHPUnit function. Looks good. Could you also add one test to use this comparator? I would like to see what a new test would look like. And we could rewrite tests to this methodology. My final question would be if there is support for this in the latest PHPUnit (still 6?).

willemstuursma:
What's up with the overzealous StyleCI job @frederikbosch?

frederikbosch:
@willemstuursma Dunno, I believe that @sagikazarmark is already looking at it. Others were also complaining.

frederikbosch:
@willemstuursma I see you added a test to the comparator. That is perfect, but not what I meant. I'd like to see a test that is using it within the current test framework, e.g. [here](https://github.com/moneyphp/money/blob/master/tests/MoneyTest.php#L22). You told me that you will have to add it to the PHPUnit bootstrap or to the XML file. I prefer the XML file since we already have that one. Finally I think this should live within an own namespace. Since it is PHPUnit specific, I would suggest something like `Money\PhpUnit\Comparator`.

So, in order to merge.
- [ ] add a test to the current framework, using the XML file if possible
- [ ] move to namespace

frederikbosch:
@willemstuursma Looks good. Question, why is the `Comparator` class itself not final? Is there a reason for it?

willemstuursma:
Thanks @frederikbosch. Not used to `final` classes, added.

frederikbosch:
@sagikazarmark In my opinion this can be merged in. This could - over time - give a huge cleanup to our unit tests.

frederikbosch:
Without any complaints, this will be merged Thursday.

frederikbosch:
Wonderful, thank you!

PR #418, opened 2017-10-31
Update getting-started.rst

driade:
When dealing with a multi currency site, FE, it can drive you nuts to understand Money initialization if you don't realice you've to create the object represented in the smallest unit. The examples in the documentation use EUR and USD as example currencies, and both hace 2 minor units. I think it'd be very beneficial to state that other currencies may have other minor units, or none!

It took me 5 hours to notice it :)


frederikbosch:
Sorry, I dont see this as a beneficial. I simply don't get what has been added to docs.

Issue #422, opened 2017-11-14
Unexpected result when converting small bitcoin amounts


dmitry027:
I ran into an issue with XBT conversion. When source amount is small enough, the result (i think) must be 0, but it is not.
```php
use Money\Converter;
use Money\Currencies\AggregateCurrencies;
use Money\Currencies\BitcoinCurrencies;
use Money\Currencies\ISOCurrencies;
use Money\Currency;
use Money\Exchange\FixedExchange;
use Money\Money;

$currencies = new AggregateCurrencies([
    new BitcoinCurrencies(),
    new ISOCurrencies(),
]);

$exchange = new FixedExchange([
    'XBT' => [
        'USD' => 6597,
    ],
]);

$converter = new Converter($currencies, $exchange);

$xbt = Money::XBT('1');
$usd = $converter->convert($xbt, new Currency('USD'));
echo $usd->getAmount();         // expected: 0, actual: 66

$xbt = Money::XBT('10');
$usd = $converter->convert($xbt, new Currency('USD'));
echo $usd->getAmount();         // expected: 0, actual: 6597

$xbt = Money::XBT('100');
$usd = $converter->convert($xbt, new Currency('USD'));
echo $usd->getAmount();         // expected: 1, actual: 1

$xbt = Money::XBT('1000');
$usd = $converter->convert($xbt, new Currency('USD'));
echo $usd->getAmount();         // expected: 7, actual: 7

$xbt = Money::XBT('10000');
$usd = $converter->convert($xbt, new Currency('USD'));
echo $usd->getAmount();         // expected: 66, actual: 66
```

sagikazarmark:
Thanks for the report @dmitry027 

Indeed it seems to be a bug, and I think it's related to the subunit difference between these currencies.

Special thanks for the test cases!

sagikazarmark:
Hm, looks like I reacted too quickly. The above script produces the expected result. I also added a few test cases and they seem to pass: #423

Which version of Money do you use? We fixed a few rounding and formatting issues since 3.0

dmitry027:
Correction: bcmath extension must not be installed.

frederikbosch:
@dmitry027 #422 fixes your issue.

sagikazarmark:
You mean #427

sagikazarmark:
Maybe we could run all test suites with all calculators? That would uncover these issues much earlier.

PR #423, opened 2017-11-14
Add a few more test cases to currency conversion

sagikazarmark:
Larger number of subunit difference is also a real life use case.


frederikbosch:
Is this ready to merge?

sagikazarmark:
I think so, I was trying to provide more tests to reproduce #422, but turned out this is not the cause. Still, large subunit difference tests are nice to have.

PR #428, opened 2017-11-20
Prep for implementation of PHP Operator Overloading.

InquisitiveDevelopment:
This is just to show an option for implementing the PHP operator overloading that will hopefully be added in PHP7. This could also be added now by installing the RFC specs binary.

The RFC for this implementation can be found [here](https://wiki.php.net/rfc/operator-overloading).

NOTE: I haven't done any testing on this. I first wanted to get the opinions of the projects maintainers. There are a few operators that didn't have any obvious implementation for a Money class so for the functions that are currently throwing an exception since the aren't implemented...should they be? Does there need to be functionality added to the library before that would work?

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -5,6 +5,7 @@ namespace Money;
 use Money\Calculator\BcMathCalculator;
 use Money\Calculator\GmpCalculator;
 use Money\Calculator\PhpCalculator;
+use Money\Traits\OperatorOverloads;
 
 /**
  * Money Value Object.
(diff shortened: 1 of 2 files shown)


Gummibeer:
This will bring the `Money` class closer to a real type that is similar to a native php one.

frederikbosch:
For the moment this has no sense. The RFC is - so far as I can tell - pretty much dead. In case the RFC, or a similar one, gets merged this might be reevaluated. Closed for now.

InquisitiveDevelopment:
I threw that together simply because there is the library that could be installed already (linked in the RFC) if someone wants to use it. Since they are magic methods the fact that they exist wouldn't have any effect until the PHP extension was added. Also, since it's added as a trait it doesn't actually have to be added by default and could instead be an opt-in option.

Gummibeer:
Opt-in wouldn't work cause the class is final - the only way would be to have a second class like `MagicMoney` (no real idea).

PR #430, opened 2017-12-08
Undefined Index "digits"

rodnaph:
When trying to parse invalid input I'm seeing an undefined index error for `digits` (php strict mode).

```
Undefined index: digits

/Users/rod/Code/misc/money/src/Parser/DecimalMoneyParser.php:76
```

I've added a test for the handling of invalid inputs, but not sure what the behaviour here should be (though I assume it should result in a ParserException somehow so it can be gracefully handled by the caller).


frederikbosch:
Good one, thanks

PR #439, opened 2018-01-02
add ratioOf method

frederikbosch:


tests/MoneyTest.php
```diff
@@ -233,6 +233,14 @@ public function test_it_supports_max_int()
         $this->assertInstanceOf(Money::class, (new Money(PHP_INT_MAX, new Currency('EUR')))->subtract($one));
     }
 
+    public function test_it_returns_ratio_of()
+    {
+        $three = new Money(3, new Currency('EUR'));
+        $six = new Money(6, new Currency('EUR'));
+
+        $this->assertEquals(0.5, $three->ratioOf($six));
```
Gummibeer: I think it would be nice to also have a `$this->assertEquals(2, $six->ratioOf($three));`. And I'm also interested in the case of `$zero->ratioOf($six)` and vice versa.

frederikbosch: The zero case, good thinking, I will apply a test case and implementation for that one.

Regarding `$six->ratioOf($three)`, that is already there since `ratioOf` is a method of `Money`, right? It does not matter whether we call `ratioOf` on `$six` or `$three`.

Gummibeer: That's in general right but my thought was to check lower divided by bigger number and bigger number divided by lower one. It's not a real other case but I know that sometimes math functions do stupid things if one of the numbers is bigger/lower than the other one. It's no real request but I think that this single line won't hurt and is just another check to be sure that it works in every case.

The zero cases are real math special ones. And sometimes one is also a bit special. For ratios, in my understanding we have following cases:
```
x / y = r
{x=0} {y>0} | r=0
{0<x<y} | 0<r<1
{x=y} | r=1
{0<y<x) | 1<r<INFINITY
{x>0} {y=0} | r=INFINITY/null/Exception
```
To be sure that it works in all of them the test should be:

```php
public function test_it_returns_ratio_of()
{
    $currency = new Currency('EUR');
    $zero = new Money(0, $currency);
    $three = new Money(3, $currency);
    $six = new Money(6, $currency);

    $this->assertEquals(0, $zero->ratioOf($six));
    $this->assertEquals(0.5, $three->ratioOf($six));
    $this->assertEquals(1, $three->ratioOf($three));
    $this->assertEquals(2, $six->ratioOf($three));
    $six->ratioOf($zero); // will result in an exception
}
```

frederikbosch: Thanks for your feedback. Updated the PR.

tests/MoneyTest.php
```diff
@@ -235,10 +235,27 @@ public function test_it_supports_max_int()
...
+    /**
+     * @expectedException \InvalidArgumentException
+     */
+    public function test_it_throws_when_calculating_ratio_of_zero()
+    {
+        $currency = new Currency('EUR');
+        $zero = new Money(0, $currency);
+        $six = new Money(6, $currency);
+
+        $this->assertEquals(2, $six->ratioOf($zero));
```
Gummibeer: You can remove the assert, the test assertion is the exception in the doc Tag.

src/Money.php
```diff
@@ -426,6 +426,20 @@ public function allocateTo($n)
...
     }
 
+    /**
+     * @param Money $money
+     *
+     * @return string
+     */
+    public function ratioOf(Money $money)
+    {
+        if (!$money->amount) {
```
Gummibeer: Shouldn't this be done in the `divide` methods itself cause it's a Division problem and no ratio one!?

sagikazarmark:
@frederikbosch Looks like there were some comments made after merging. Sorry for not reviewing it earlier, but I kinda agree with those points.

frederikbosch:
@sagikazarmark I improved the test and removed the assert as suggested.

However I did not move the 0 check to the calculator. See how we are implementing the divide method: the division by 0 check is in there, and not in the calculator. For the sake of consistency I choose to go for the same implementation in the ratioOf method.

PR #432, opened 2017-12-15
Fix deprecation notice for IntlMoneyParser

sstok:
The IntlMoneyParser use the Intl NumberFormatter to parse a currency, and assigns the currency to a variable. But later-on the if statement throws an deprecation warning because the currency is not a `Currency` object.

The actual deprecation notice was never shown as it's silenced and the Symfony PHPUnit bridge was not installed.

PR #440, opened 2018-01-02
update phpunit and phpspec + drop php 5.5

frederikbosch:
Version 6 is not possible because it is blocked by PHPSpec, so for the moment an update to PHPUnit 5 is the maximum.


frederikbosch:
@sagikazarmark Let's drop PHP 5.5 from version 3.1. I have trouble getting a green light on Travis because of PHPUnit 4. That is why I am upgrading to PHPUnit 5.

Upgrading to PHPUnit 5 means dropping PHP 5.5, which I think is not a problem at all. It has [probably dropped below 5% of usage](https://seld.be/notes/php-versions-stats-2017-2-edition). It is not getting support from PHP anymore. Finally, those users can still use Money 3.0.x which is all they need as legacy projects.

frederikbosch:
The alternative is allowing failures for PHP 7.2 tests, which I think is far more worse.

frederikbosch:
I see that the originating problem is #432. But to be honest, I have no time and energy to fix that. PHP 5.5 is EOL / done.

frederikbosch:
Also updated the readme to clarify our standing point towards supporting PHP versions.

frederikbosch:
Rebased it to latest master.

PR #443, opened 2018-01-03
add intl localized decimal parser/formatter

frederikbosch:
Parsing and formatting for localized decimal amounts


frederikbosch:
All tests and functionality is almost exactly the same as for the `IntlMoneyFormatter` and `IntlMoneyParser`. I would love to have these features too in 3.1.

frederikbosch:
Use case: parse values from webforms with localized decimal inputs into `Money` objects.

PR #447, opened 2018-01-16
Fix float cast to string isssue on certain locales

### LINKED ISSUE #446: Float conversion ratio raises exception on some locales (v3.1.0 only)
```
$exchange = new FixedExchange([
    'EUR' => [
        'USD' => 1.25
    ]
]);

$converter = new Converter(new ISOCurrencies(), $exchange);

$eur100 = Money::EUR(100);
$usd125 = $converter->convert($eur100, new Currency('USD'));
```
In some cases this code (from documentation) raises error 
`Invalid integer part 1,25. Invalid digit , found` at `src/Number.php:234`
It depends on active locale. For example on ru_RU.UTF-8 float `1.25` ratio becomes a `"1,25"` string. Meanwhile Number::fromString($value) can't deal with comma separated number string.

sagikazarmark:
Fixes #446

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM php:7.2-cli
+
+RUN set -xe \
+	&& apt-get update \
+	&& apt-get install -qqy libicu-dev locales libgmp-dev \
+	&& docker-php-ext-install -j$(nproc) intl bcmath gmp
+
+RUN set -xe \
+	&& echo en_US ISO-8859-1 >> /etc/locale.gen \
+	&& echo en_US.UTF-8 UTF-8 >> /etc/locale.gen \
+	&& echo en_CA ISO-8859-1 >> /etc/locale.gen \
+	&& echo en_CA.UTF-8 UTF-8 >> /etc/locale.gen \
+	&& echo ru_RU ISO-8859-5 >> /etc/locale.gen \
+	&& echo ru_RU.UTF-8 UTF-8 >> /etc/locale.gen \
+	&& locale-gen \
+    && update-locale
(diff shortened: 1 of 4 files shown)


sagikazarmark:
Looks like when the language pack is not installed, the decimal separator still ends up being a comma.

https://travis-ci.org/moneyphp/money/jobs/329765887#L993

sagikazarmark:
Once travis gets back from where it is now the tests should pass.

frederikbosch:
Nice. I know we have `setlocale` call in `MoneyTest`. We should use the PHPUnit one there too. It gave me problems before. This can be fixed with this PR.

frederikbosch:
Is it possible that `quote` returns a non-float. If so, it will be a problem too. Maybe we should add `fromScalar` now.

```php
public static function fromScalar($numeric)
{
    if (is_float($numeric)) return self::fromFloat($numeric);
    if (is_string($numeric)) return self::fromString($numeric);
    if (is_integer($numeric)) return new self($numeric);
    throw new \InvalidArgumentException();
}
```

frederikbosch:
Or use a switch to reduce the overhead.

sagikazarmark:
I think conversion ratios are hardly ever integer values. The contract currently says float, so if something returns an integer it's a bug and needs to be converted to float. I can live with that for now, instead of adding some logic for guessing the format.

frederikbosch:
Aight, sounds good.

sagikazarmark:
@frederikbosch can you review this before merging?

frederikbosch:
Seems good. I changed the `MoneyTest` using `setlocale()` to also use `$this->setLocale()`.

frederikbosch:
Merged it. Also merged the PR of @bendavies. Will tag 3.1.1 if the Travis lights turn green on the master commits.

PR #448, opened 2018-01-17
Add Converter interface

### LINKED ISSUE #435: Converter is also final
There are many issues where discussed why Money class is final, it's value object and so on.
But why is Converter is final? It also doesn't implement any interface, so no way to mock it in tests. Converter is not value object, right?

ossinkine:
Closes #435

### DIFF
--- a/src/Converter.php
+++ b/src/Converter.php
@@ -7,7 +7,7 @@ namespace Money;
  *
  * @author Frederik Bosch <f.bosch@genkgo.nl>
  */
-final class Converter
+final class Converter implements CurrencyConverter
 {
     /**
      * @var Currencies
--- a/src/CurrencyConverter.php
+++ b/src/CurrencyConverter.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Money;
+
+/**
+ * Provides a way to convert Money to Money in another Currency.
+ *
+ * @author Gocha Ossinkine <ossinkine@ya.ru>
+ */
+interface CurrencyConverter
+{
+    /**
+     * @param Money    $money
+     * @param Currency $counterCurrency
+     * @param int      $roundingMode
+     *
+     * @return Money
+     */
+    public function convert(Money $money, Currency $counterCurrency, $roundingMode = Money::ROUND_HALF_UP);
+}


frederikbosch:
Hmm. This would be the ideal situation, and we should could consider going for this in a new major version. But this is never going to land in Money 3.

If you want this right now, and I believe you do, we should create an interface that is compatible with the current state of the library. Why not call the interface `CurrencyConverter`? Then we would not have to change the name of `Converter`?

frederikbosch:
In a new major version we then might rename `Converter` to `ExchangeConverter`.

ossinkine:
I tried to keep it consistent with already existing classes.
I'll do as you suggested.

frederikbosch:
Cool.

ossinkine:
Updated

frederikbosch:
@sagikazarmark What do you think? I don't think there is any harm in having the converter implementing an interface.

sagikazarmark:
Are there any use cases for alternative implementations? And even then, do they actually do the same thing?

In my opinion the Converter is a service, or a facade if you wish. It depends on abstractions, which makes it kinda extensible. It's also relatively small, so for any custom use cases it can just be copied and modified.

Personally I think this is a bit overengineering. If you can give me a few examples where this might actually be useful within the domain I might reconsider.

ossinkine:
@sagikazarmark It has been discussed a lot of times.

> for any custom use cases it can just be copied and modified

No, you cannot just copies because original converter and new one don't have a common interface (see [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle)). Class which use original converter does not accept new one.

Second reason is there is no way to test class which use converter because you cannot create a mock for converter.

frederikbosch:
> Second reason is there is no way to test class which use converter because you cannot create a mock for converter.

That is a bad reason. You don't mock the `Converter`, you mock what's injected in them. 

```php
$converter = new Converter($this->createMock(Currencies::class), $this->createMock(Exchange::class));
```

Furthermore, my advice would be to use a mock as a final resort, not as a common test technique.

ossinkine:
@frederikbosch Unit test should test my class but not the converter.

frederikbosch:
> Unit test should test my class but not the converter.

Yes, but that does not mean you are not allowed to instantiate other objects.

ossinkine:
I should not instantiate and use other objects if they can affect my code.
The goal of unit testing is to isolate each part of the program and show that the individual parts are correct. ([Wikipedia](https://en.wikipedia.org/wiki/Unit_testing#Advantages)). That mean if the converter will be broken my test should still be working.

sagikazarmark:
When I said copying code, I didn't mean to use it as an *alternative* to the Converter in this library. 

Dependencies between code does not always require an abstraction. Also, as I said I consider the Converter as a service which rarely require an abstraction, so I don't see LSP necessarily related here.

You are right about the unit testing part: it's not a unit test anymore, it's called an integration test. You shouldn't mock what you don't own anyway.

https://adamwathan.me/2017/01/02/dont-mock-what-you-dont-own/

ossinkine:
Your position is understandable, although it is very subjective and difficult to accept.

Integration test is good, but it is not a replacement for a unit test, it can be a complement.

frederikbosch:
@sagikazarmark Your position is the same as mine, but why should we object to an interface for the converter?

sagikazarmark:
> but why should we object to an interface for the converter?

I'm not objecting anything given valid reasons provided. I asked specifically for reasons to provide alternative implementations...none has been given.

Think about it this way: there is a Converter interface......then there is a DefaultConverter as we probably can't define any other use cases where a converter implementation makes sense. (The only difference is that they have meaningful names, but ultimately everyone will just think of it as the default implementation)

Please, prove that I'm wrong!

As far as I can see the only reason for this interface right now is to be able to mock it....which is just simply wrong. Even if there *was* an interface, mocking it would still be wrong.

> Integration test is good, but it is not a replacement for a unit test, it can be a complement.

Sure thing. But there are cases when you can't write unit tests, because on the unit level you don't own the whole code. That's the definition of writing integration tests.

Sadly you can find some examples even in Money where we mock interfaces not owned by us. Not good, but sometimes it's acceptable. But adding an interface purely for being able to mock it is (again) wrong IMHO.

If I were you, I would probably mock the exchange interface if I had to.
But I don't see why using the fixed exchange and calling it integration test is bad.
Ultimately, you can also provide your own abstraction over currency conversion (as pointed out in a comment in the linked issue), that is how you can own the entire code.

@frederikbosch I'm leaving this decision to you. It won't hurt if this interface makes it's way into the package, but currently I can only see bad practice out of it and that's not the effect an OS library should have or support.

frederikbosch:
> But I don't see why using the fixed exchange and calling it integration test is bad.

It definitely is not, see it as a [fake object](http://www.yegor256.com/2014/09/23/built-in-fake-objects.html).

> Ultimately, you can also provide your own abstraction over currency conversion (as pointed out in a comment in the linked issue), that is how you can own the entire code.

I second that. Nobody tells you that you are required to use _every_ class in this project.

> I asked specifically for reasons to provide alternative implementations...none has been given
> Please, prove that I'm wrong!

@ossinkine If you can prove him wrong, I will merge, otherwise I close it.

PR #456, opened 2018-02-09
Sum

### LINKED ISSUE #387: Money sum
Hello everyone

I want to make a sum for the amount of more than 2 Money objects. 
I thought I can make this using the `add` function, until I read the documentation and I've seen that `add` function returns a string not a Money object.

Do you have any idea how I could make a sum for more than 2 Money objects?

sagikazarmark:
Fixes #387

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
-- `Moneay::add` and `Money::subtract` now accept variadic arguments
+- `Money::add` and `Money::subtract` now accept variadic arguments
 
 
 ## [3.1.3] - 2018-02-16
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+
+- `Moneay::add` and `Money::subtract` now accept variadic arguments
+
 
 ## [3.1.3] - 2018-02-16
 
--- a/doc/features/operation.rst
+++ b/doc/features/operation.rst
@@ -17,6 +17,16 @@ Additions can be performed using ``add()``.
 
     $result = $value1->add($value2);   // €13.00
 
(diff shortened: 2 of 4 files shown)


frederikbosch:
Looks great. What do you think of a static named constructor sum in addition? That would look similar to the code of Ben. It would help my own codebase a lot.

frederikbosch:
Also, you could reduce the number of getCalculator calls.

sagikazarmark:
> What do you think of a static named constructor sum in addition?
> It would help my own codebase a lot.

Do you have an example use case? I'm not against it per se, but I would like to see how it would be used. My only concern is that having sum for add would be a little inconsistent. What about subtract then?

> Also, you could reduce the number of getCalculator calls.

Nice catch, thanks.

sagikazarmark:
> You should be able to supply the currency as second argument to create the zero amount.

I'm not sure I like that idea. Let's stick to the shopping cart example: normally your shopping cart contains items with prices. The cart itself doesn't know anything about currencies or payment methods.

With that sum solution you would have to pass a currency object to the cart to use it when there are no items in the cart. It's not only an unnecessary knowledge for the cart, but also makes it extremely easy to get into an inconsistent state: what if you change the currency? You naturally iterate through the cart and change the money values, but you also have to think about passing the new currency to the cart as well.

I understand the zero money problem, but I don't know the solution and I think the sum function this way is not really a good idea.

Without the currency fallback and requiring at least one element in the array the static sum function could work.

frederikbosch:
@sagikazarmark You are right. Have been thinking what would be best practise after this PR. That would be the following.

```php
$sum = (new Money(0, new Currency('EUR')))->add(...$collection);
```

So create a zero amount and then add the collection.

frederikbosch:
However, how would one create the sum of an iterable? This PR solves sum only for arrays. So the only way to solve it for `iterable`s is to convert the `iterable` to an array and then add.

sagikazarmark:
Can't really think of any other solution right now

frederikbosch:
Me neither, maybe another solution will be proposed in the future. I think the current solution is alright.

frederikbosch:
Perfect, great work!

sagikazarmark:
Thanks

PR #458, opened 2018-02-11
[WIP] Create calculator registry implementation

sagikazarmark:


frederikbosch:
Looks good!

frederikbosch:
@sagikazarmark I was thinking. We could decide to not make the `Registry` a static class. And set it as a static property of Money, like this. You would agree that we must avoid static classes. However, I also see that it involves some additional code. What is the best solution here, and why?

```php
// create a register
$registry = new Registry([
  new BcMathCalculator(),
  new GmpCalculator(),
  new PhpCalculator(),
]);

// Money class
final class Money {
  private function getCalculatorRegister() {
    if (!self::$calculatorRegister) {
      self::$calculatorRegister = Register::fromDefaultExtensions();
    }
    return self::$calculatorRegister;
  }
  private function getCalculator() {
    return self::getCalculatorRegister()->getCalculator();
  }
  /** @deprecated **/
  public static function registerCalculator($calculator) {
    self::getCalculatorRegister()->register($calculator);
  }
  public static function setCalculatorRegistry($registry) {
    self::$calculatorRegister = $register;
  }
}
```

sagikazarmark:
Hm, yeah, great idea, would definitely make testing easier. However, there has to be some sort of static initialization, preferably outside of money, because the whole point of this separation is that other components can access the same calculator context.

Will work on this, let's see what I can come up with.

sagikazarmark:
@frederikbosch okay, so I made the registry a singleton, added a few features and reorganized internals.

But somehow it still doesn't feel right this way. Could use some fresh eyes. 👁

src/Calculator/Registry.php
```diff
@@ -0,0 +1,148 @@
...
+
+        $this->calculators[$calculator] = false;
+
+        $this->resetCalculator();
+    }
+
+    /**
+     * Disables arbitrary precision calculators.
+     */
+    public function disableArbitraryPrecisionCalculators()
```
frederikbosch: Rather would say you want something like `Register::phpOnly()`.

src/Calculator/Registry.php
```diff
@@ -0,0 +1,148 @@
...
+
+        $this->resetCalculator();
+    }
+
+    /**
+     * Enables a calculator.
+     *
+     * @param string $calculator
+     */
+    public function enableCalculator($calculator)
```
frederikbosch: I would remove the suffix `Calculator`.

src/Calculator/Registry.php
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * @var Registry
+     */
+    private static $instance;
+
+    /**
+     * @param string $calculator
+     */
+    public function registerCalculator($calculator)
```
frederikbosch: I would remove the suffix `Calculator`.

src/Calculator/Registry.php
```diff
@@ -0,0 +1,148 @@
...
+
+        $this->resetCalculator();
+    }
+
+    /**
+     * Disables a calculator.
+     *
+     * @param string $calculator
+     */
+    public function disableCalculator($calculator)
```
frederikbosch: I would remove the suffix `Calculator`.

src/Calculator/Registry.php
```diff
@@ -0,0 +1,148 @@
...
+    {
+        $this->disableCalculator(BcMathCalculator::class);
+        $this->disableCalculator(GmpCalculator::class);
+    }
+
+    /**
+     * Resets the already resolved calculator instance.
+     * Used when the registry is modified in any way (eg. registering a new calculator).
+     */
+    private function resetCalculator()
```
frederikbosch: I would remove the suffix `Calculator`.

src/Calculator/Registry.php
```diff
@@ -0,0 +1,148 @@
...
+    {
+        $this->calculator = null;
+    }
+
+    /**
+     * @return Calculator
+     *
+     * @throws \RuntimeException If cannot find calculator for money calculations
+     */
+    private function initializeCalculator()
```
frederikbosch: I would remove the suffix `Calculator`.

src/Calculator/Registry.php
```diff
@@ -0,0 +1,148 @@
...
+            }
+        }
+
+        throw new \RuntimeException('Cannot find calculator for money calculations');
+    }
+
+    /**
+     * @return Calculator
+     */
+    public function getCalculator()
```
frederikbosch: I would remove the suffix `Calculator`.

frederikbosch:
Superseded by #646

PR #459, opened 2018-02-11
Update PHPUnit test case class

### LINKED ISSUE #457: Use new PHPUnit test class

sagikazarmark:
Fixes #457

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "php": ">=5.6"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5",
+        "phpunit/phpunit": "^5.7 || ^6.4 || ^7.0",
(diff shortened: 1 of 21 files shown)


composer.json
```diff
@@ -23,7 +23,7 @@
         "php": ">=5.6"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5",
+        "phpunit/phpunit": "^5.7 || ^6.4 || ^7.0",
```
frederikbosch: What's the point of this and not only `^7.0`?

sagikazarmark: We still support PHP 5.6

sagikazarmark: But we could drop 5.x from the requirements.

PR #460, opened 2018-02-11
Fix floating point casting to integer in PHP calculator

### LINKED ISSUE #455: Float to string casting provides errors on some locales
```php
$full_summ = new Money(1000000, new Currency('RUB'));
$vat_summ = $full_summ->divide(1.18);

echo 'Should be: ' . round(10000/1.18, 2);
echo "<br>\n";
echo 'We have: '. $vat_summ->getAmount()/100;
```

Gives us:
```
Should be: 8474,58
We have: 8474,57
```

After debugging I found that it's because of `divide($amount, $divisor)` in `PhpCalculator.php` — there we have `return (string) $result` — on some locales (e.g. ru_RU.UTF8) it creates 847457,627 instead of 847457.627 and after that that part after the comma are gone in further operations.

sagikazarmark:
Fixes #455

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [Unreleased]
+
+### Fixed
+
+- Float cast to string issue on certain locales
+
+
 ## [3.1.1] - 2018-01-19
 
 ### Fixed
--- a/src/Calculator/PhpCalculator.php
+++ b/src/Calculator/PhpCalculator.php
@@ -109,23 +109,23 @@ final class PhpCalculator implements Calculator
     public function round($number, $roundingMode)
     {
         if (Money::ROUND_HALF_POSITIVE_INFINITY === $roundingMode) {
-            $number = Number::fromString((string) $number);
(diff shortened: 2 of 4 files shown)


sagikazarmark:
@chekalskiy can you please check if this solves the issue for you?

sagikazarmark:
Yup, it did.

chekalskiy:
I'll check a bit later. Thank you

chekalskiy:
I've tested and now my test case looks fine. 👍

But shouldn't we add `castString()` to `add()`, `subtract()`, `absolute()` and `mod()` methods?

sagikazarmark:
No, as they don't involve casting float to string:

- add: adding two integers will result in an integer anyway
- subtract: subtracting one integer from another will result in an integer
- absolute: there is no calculation related to precision
- mod: there is no calculation related to precision

sagikazarmark:
But I've just found other places where the tests break. I added a new test case for running all tests with russian UTF8 locale for the php calculator.

sagikazarmark:
@frederikbosch see the separate commit with the breaking tests. Solution pushed in a separate commit.

sagikazarmark:
Merging this as it fixes the problem, but I'm going to do some refactoring around the number class. I feel that there is too many string casting here and there.

frederikbosch:
Totally agree. In my perception, in version 4 we should use Number everwhere internally. That means only accepting Number or directly do the conversion on every numeric parameter.

PR #461, opened 2018-02-12
Fix KRW minorUnit

lorenzomar:
minorUnit in case of South Korea Won must be 2 as specified by xe.com (http://www.xe.com/it/currency/krw-south-korean-won)


frederikbosch:
This must be done through [our iso-currencies package](https://github.com/moneyphp/iso-currencies), which uses [this xml file](https://www.currency-iso.org/dam/downloads/lists/list_one.xml) as repository for the currencies.

PR #464, opened 2018-02-14
Number

### LINKED ISSUE #462: Improve Number class numeric value handling

sagikazarmark:
Did a massive refactor on Number usage. For now, calculators do the conversion internally where necessary, but in the future we could just require a number in the calculator itself, but this is up to a future debate.

Fixes #462

### DIFF
--- a/src/Number.php
+++ b/src/Number.php
@@ -64,7 +64,7 @@ final class Number
     public static function fromFloat($number)
     {
         if (is_float($number) === false) {
-            throw new \InvalidArgumentException('Floating point expected');
+            throw new \InvalidArgumentException('Floating point value expected');
         }
 
         return self::fromString(sprintf('%.14F', $number));
--- a/src/Number.php
+++ b/src/Number.php
@@ -71,6 +71,24 @@ final class Number
     }
 
     /**
+     * @param float|int|string $number
+     *
+     * @return self
+     */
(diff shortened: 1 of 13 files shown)


sagikazarmark:
Also, this PR fixes a number of local aware decimal separator issues with bcmath and gmp.

frederikbosch:
Great idea!

src/Number.php
```diff
@@ -38,6 +38,56 @@ public function __construct($integerPart, $fractionalPart = '')
...
+    }
+
+    /**
+     * @param float|int|string $number
+     *
+     * @return self
+     */
+    public static function fromNumber($number)
+    {
+        if (is_float($number)) {
```
frederikbosch: I would not use `elseif` here. But rather only a new if statement, or maybe, even a `switch` statement.

```php
switch (gettype($number)) {
  case 'integer':
   ...
}
```

The above saves a one or more function calls.

sagikazarmark: Using gettype for type checking is generally not advised, but using separate ifs is probably cleaner.

sagikazarmark:
(Note in advance: this PR became a bit bigger, so I don't want to squash commits, there are a few meaningful ones)

sagikazarmark:
Thanks for the review!

PR #475, opened 2018-02-23
Add money denormalization

sagikazarmark:


src/Money.php
```diff
@@ -78,6 +78,24 @@ public function __construct($amount, Currency $currency)
...
+     *
+     * @return Money
+     *
+     * @throws \InvalidArgumentException when the normalized version cannot be denormalized
+     */
+    public static function denormalize($normalized)
+    {
+        if (is_object($normalized) && isset($normalized->amount) && isset($normalized->currency)) {
+            return new Money($normalized->amount, new Currency($normalized->currency));
+        } elseif (is_array($normalized) && isset($normalized['amount']) && isset($normalized['currency'])) {
```
frederikbosch: If on a new line.

willemstuursma: I guess you could also just cast the array to stdClass using `(object)` or the other way around and get rid of half the lines.

spec/MoneySpec.php
```diff
@@ -385,6 +385,36 @@ function it_throws_an_exception_when_currency_is_different_during_modulus(Calcul
         $this->shouldThrow(\InvalidArgumentException::class)->duringMod(new Money(self::AMOUNT, new Currency(self::OTHER_CURRENCY)));
     }
 
+    function it_denormalizes_money_from_a_stdclass_object()
+    {
+        $normalized = new \stdClass();
```
willemstuursma: I guess that it would be nice in the test to show its opposite usage:

`$normalized = json_decode(json_encode($money));`

sagikazarmark: Although I (finally) understand your use case, I believe this is fundamentally different from json serialization.

JSON is the serialization part, this is denormalization (converting an intermediate representation to the actual form). You could use it for YAML or XML serialization as well.

That's why I tried to keep this thing separated from JSON from the very beginning.

It could be mentioned in documentation though.

sagikazarmark:
Maybe we could implement it as a separate function?

gmponos:
Sorry for saying my opinion although not a member of this library but I really believe that serialization/deserialization has nothing to do with the library.

sagikazarmark:
You are somewhat right: I don't like having normalization logic in `Money` either. Actually it breaks SRP, so if this function ever gets inside this package it will probably be a separate function or placed in a separate class.

I don't think however, that this functionality would be totally out of scope for the package. Providing some utilities for serialization can help getting started, while you can still write your own when using Money in a more complex application.

frederikbosch:
Since this needs more work, and @gmponos provided valid arguments for not being a necessity, I will close this.

PR #478, opened 2018-03-22
add aggregate functions, fixes #473

### LINKED ISSUE #473: Function to get least / highest from iterable of Money's
We need a method or function to get the lowest or highest Money of an iterable of `Money`s. 

I was wondering if you have any suggestions on how to put this in the library and then I can provide a PR. 

Looking for something like:

`$money = Lowest(...$monies);`.

frederikbosch:
Add aggregate functions `minimum`, `maximum`, `sum` and `average`.

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -517,11 +517,11 @@ final class Money implements \JsonSerializable
 
     /**
      * @param Money   $first
-     * @param Money[] $collection
+     * @param Money   ...$collection
      *
      * @return Money
      */
-    public static function minimum(self $first, self ...$collection)
+    public static function min(self $first, self ...$collection)
     {
         $min = $first;
 
@@ -536,11 +536,11 @@ final class Money implements \JsonSerializable
 
     /**
      * @param Money   $first
-     * @param Money[] $collection
+     * @param Money   ...$collection
      *
      * @return Money
      */
-    public static function maximum(self $first, self ...$collection)
+    public static function max(self $first, self ...$collection)
     {
         $max = $first;
 
@@ -555,7 +555,7 @@ final class Money implements \JsonSerializable
 
     /**
      * @param Money   $first
-     * @param Money[] $collection
+     * @param Money   ...$collection
      *
      * @return Money
      */
@@ -566,11 +566,11 @@ final class Money implements \JsonSerializable
 
     /**
      * @param Money   $first
-     * @param Money[] $collection
+     * @param Money   ...$collection
      *
      * @return Money
      */
-    public static function average(self $first, self ...$collection)
+    public static function avg(self $first, self ...$collection)
     {
         return $first->add(...$collection)->divide(func_num_args());
     }
--- a/src/Money.php
+++ b/src/Money.php
@@ -516,12 +516,12 @@ final class Money implements \JsonSerializable
     }
 
     /**
-     * @param Money $first
+     * @param Money   $first
      * @param Money[] $collection
      *
      * @return Money
      */
-    public static function minimum(self $first, self...$collection)
+    public static function minimum(self $first, self ...$collection)
     {
         $min = $first;
 
@@ -535,12 +535,12 @@ final class Money implements \JsonSerializable
     }
 
     /**
-     * @param Money $first
+     * @param Money   $first
      * @param Money[] $collection
      *
      * @return Money
      */
-    public static function maximum(self $first, self...$collection)
+    public static function maximum(self $first, self ...$collection)
     {
         $max = $first;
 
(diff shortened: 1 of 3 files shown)


frederikbosch:
>how about return self::sum($first, ...$collection)->divide(func_num_args());

I tend to reduce the number of method calls, if possible and not decreasing readability. Regarding the test and the exceptions, I updated it. That's much better indeed.

@sagikazarmark What do you think?

src/Money.php
```diff
@@ -515,6 +515,66 @@ public function jsonSerialize()
         ];
     }
 
+    /**
+     * @param Money   $first
+     * @param Money[] $collection
+     *
+     * @return Money
+     */
+    public static function minimum(self $first, self ...$collection)
```
sagikazarmark: I believe max/min/avg are well-known forms aggregation functions. Sum is already in that shorter form.

src/Money.php
```diff
@@ -515,6 +515,66 @@ public function jsonSerialize()
...
+        return $min;
+    }
+
+    /**
+     * @param Money   $first
+     * @param Money[] $collection
+     *
+     * @return Money
+     */
+    public static function maximum(self $first, self ...$collection)
```
sagikazarmark: I believe max/min/avg are well-known forms aggregation functions. Sum is already in that shorter form.

src/Money.php
```diff
@@ -515,6 +515,66 @@ public function jsonSerialize()
         ];
     }
 
+    /**
+     * @param Money   $first
+     * @param Money[] $collection
```
sagikazarmark: This isn't correct. The second argument is not an array. I believe I saw `...Money` notation somewhere, but I'm not sure.

src/Money.php
```diff
@@ -515,6 +515,66 @@ public function jsonSerialize()
...
+        return $first->add(...$collection);
+    }
+
+    /**
+     * @param Money   $first
+     * @param Money[] $collection
+     *
+     * @return Money
+     */
+    public static function average(self $first, self ...$collection)
```
sagikazarmark: I believe max/min/avg are well-known forms aggregation functions. Sum is already in that shorter form.

src/Money.php
```diff
@@ -515,6 +515,66 @@ public function jsonSerialize()
...
+                $min = $money;
+            }
+        }
+
+        return $min;
+    }
+
+    /**
+     * @param Money   $first
+     * @param Money[] $collection
```
sagikazarmark: This isn't correct. The second argument is not an array. I believe I saw ...Money notation somewhere, but I'm not sure.

src/Money.php
```diff
@@ -515,6 +515,66 @@ public function jsonSerialize()
...
+                $max = $money;
+            }
+        }
+
+        return $max;
+    }
+
+    /**
+     * @param Money   $first
+     * @param Money[] $collection
```
sagikazarmark: This isn't correct. The second argument is not an array. I believe I saw ...Money notation somewhere, but I'm not sure.

src/Money.php
```diff
@@ -515,6 +515,66 @@ public function jsonSerialize()
...
+     * @return Money
+     */
+    public static function sum(self $first, self ...$collection)
+    {
+        return $first->add(...$collection);
+    }
+
+    /**
+     * @param Money   $first
+     * @param Money[] $collection
```
sagikazarmark: This isn't correct. The second argument is not an array. I believe I saw ...Money notation somewhere, but I'm not sure.

tests/MoneyTest.php
```diff
@@ -259,6 +260,22 @@ public function it_throws_when_calculating_ratio_of_zero()
         $six->ratioOf($zero);
     }
 
+    /**
+     * @dataProvider aggregateExamples
+     * @test
+     */
+    public function it_calculates_aggregates($values, $throws, $min, $max, $sum, $avg)
```
sagikazarmark: I really would love to see this test split per method.

tests/MoneyTest.php
```diff
@@ -365,4 +382,50 @@ public function modExamples()
...
+                [Money::USD(-5), Money::EUR(-10), Money::EUR(-15)],
+                \InvalidArgumentException::class,
+                null,
+                null,
+                null,
+                null,
+            ],
+            [
+                [],
+                version_compare(PHP_VERSION, '7.0.0') >= 0 ? \ArgumentCountError::class : Error::class,
```
sagikazarmark: This should rather be a separate test with a phpunit version requirement

https://phpunit.de/manual/current/en/incomplete-and-skipped-tests.html#incomplete-and-skipped-tests.skipping-tests-using-requires

sagikazarmark:
Well, I'm not going to use it much, but yeah, I can think of a few cases when it can come handy. So sure

frederikbosch:
Shall we disable those bot posts? It obscures the conversation.

frederikbosch:
@sagikazarmark I improved the PR by implementing your suggested changes.

frederikbosch:
@sagikazarmark Could you have another look at this PR? I would like to see this land in master and be tagged in a version.

sagikazarmark:
I disabled the bot.

Looks good to me. Added it to the 3.2 milestone.

PR #490, opened 2018-06-28
the calculator is optionally set by the programmer in the instantiation

programarivm:
Hi there,

This is the continuation of 

- https://github.com/moneyphp/money/pull/389
- https://github.com/programarivm/money-pattern

Sorry for bringing this topic up again. Just wondering if `moneyphp/money` could take this issue into consideration.

Safe examples:
```
$fiveEur = Money::EUR(500, new PhpCalculator());
$tenEur = new Money(1000, new Currency('EUR'), new PhpCalculator());
```

This way, the programmer can optionally set the calculator of their choice when instantiating `Money` objects -- regardless of the server settings.

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -2,6 +2,7 @@
 
 namespace Money;
 
+use Money\Calculator;
 use Money\Calculator\BcMathCalculator;
 use Money\Calculator\GmpCalculator;
 use Money\Calculator\PhpCalculator;
@@ -63,7 +64,7 @@ final class Money implements \JsonSerializable
      *
      * @throws \InvalidArgumentException If amount is not integer
      */
-    public function __construct($amount, Currency $currency)
+    public function __construct($amount, Currency $currency, Calculator $calculator = null)
     {
         if (filter_var($amount, FILTER_VALIDATE_INT) === false) {
             $numberFromString = Number::fromString($amount);
@@ -76,6 +77,7 @@ final class Money implements \JsonSerializable
 
         $this->amount = (string) $amount;
         $this->currency = $currency;
+        self::$calculator = $calculator;
     }
 
     /**
--- a/src/MoneyFactory.php
+++ b/src/MoneyFactory.php
@@ -202,6 +202,6 @@ trait MoneyFactory
      */
     public static function __callStatic($method, $arguments)
     {
-        return new Money($arguments[0], new Currency($method));
+        return new Money($arguments[0], new Currency($method), $arguments[1]);
     }
 }
--- a/tests/MoneyFactoryTest.php
+++ b/tests/MoneyFactoryTest.php
@@ -2,6 +2,9 @@
 
 namespace Tests\Money;
 
+use Money\Calculator\BcMathCalculator;
+use Money\Calculator\GmpCalculator;
+use Money\Calculator\PhpCalculator;
 use Money\Currencies\AggregateCurrencies;
 use Money\Currencies\BitcoinCurrencies;
 use Money\Currencies\ISOCurrencies;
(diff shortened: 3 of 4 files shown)


sagikazarmark:
I'm not sure it would make sense to choose a calculator on a per instantiation basis, it's rather a one time choice. Do you have any use cases that would require it?

As for choosing a calculator: I'm not against making it part of this PR: #458 
One could instruct a global registry to use a chosen calculator

programarivm:
Yep, the previous examples might not look very practical.

At this moment the idea is to make sure that the app won't run arbitrary precision arithmetic functions if a system administrator activates the extensions unintentionally -- this is an issue IMO, because `moneyphp/money` encourages to install the extensions.

Also, a malicious attacker could suggest somebody to just install `moneyphp/money` with the objective of enabling BC Math and GMP in order to use the extensions for other purposes. 

So, who's behind the decision that the server runs arbitrary precision arithmetic?

- moneyphp/money
- Programmer
- Sysadmin
- Attacker

I believe this is an issue -- at least theoretically!

The API interface could also use a second parameter to enable the so-called safe mode, which is optional, like this:

```
$fiveEur = Money::EUR(500, true);
$tenEur = new Money(1000, new Currency('EUR'), true);
```
The global registry sounds really good. 

I believe this will bring a bit of awareness and give a bit more power to the programmer.

sagikazarmark:
> this is an issue IMO, because moneyphp/money encourages to install the extensions.

I don't think we ever encouraged to install anything, but let's not get into that conversation again. 😄 

If you take a look at the linked PR, there is a `disableArbitraryPrecisionCalculators` method in the registry which does exactly the thing you want.

> give a bit more power to the programmer.

Yeah, this might not be that good with the currently available API.

src/Money.php
```diff
@@ -76,6 +77,7 @@ public function __construct($amount, Currency $currency)
 
         $this->amount = (string) $amount;
         $this->currency = $currency;
+        self::$calculator = $calculator;
```
sagikazarmark: This wouldn't be a good idea anyway, because it sets a global instance.

programarivm: Thanks for the remark. If my memory serves me, I took that line of code from the following method:
```
     /**
      * @return Calculator
      */
     private function getCalculator()
     {
         if (null === self::$calculator) {
             **self::$calculator = self::initializeCalculator();**
         }
 
         return self::$calculator;
     }
```

programarivm: Yep, `disableArbitraryPrecisionCalculators` is really helpful in my opinion. Anyway I still wonder if it makes any sense to suggest that `BcMathCalculator` and `GmpCalculator` can be moved from `moneyphp/money` to a new repo (`moneyphp/big-money`).

programarivm:
Hi, just provided this insight into another repo: https://github.com/php-amqplib/php-amqplib/pull/573
Let me share this: https://www.cisecurity.org/advisory/multiple-vulnerabilities-in-php-could-allow-for-arbitrary-code-execution_2017-093/
I agree with this recommendation: _Apply the principle of Least Privilege to all systems and services_

sagikazarmark:
> Anyway I still wonder if BcMathCalculator and GmpCalculator could be moved from moneyphp/money to a new repo (moneyphp/big-money).

I don't think it's necessary, but it would make the implementations less accessible. Having that code in this repository doesn't create a hard dependency on bcmath or gmp. I understand that you want to see these things gone, but please understand that it won't happen.

>  I agree with this recommendation: Apply the principle of Least Privilege to all systems and services

@programarivm we got your message, repeating it over and over and over again won't move things forward, quite the contrary.

sagikazarmark:
Closing as #458 will provide solution to the original problem.

PR #492, opened 2018-07-27
Add FAILING test reproducing allocation bug

mnapoli:
It seems there is a bug in the allocation behavior: if I allocate 1 cent like this:

- 33% to person A
- 66% to person B

one would expect rounding to take precedence over the "remainder cent logic".

Expected: 

- person A: 0 cents
- person B: 1 cent

Actual:

- person A: 1 cent
- person B: 0 cents

---

A real scenario where this can be problematic: a sale of 99.99€:

- 75% commission to person A: 99.99€×75% = 74.9925€ -> should be 74.99€
- 25% commission to person B: 99.99€×25% = 24.9975€ -> should be 25.00€

There is no additional or missing cent to allocate here. However `Money::allocate()` will allocate:

- person A: 75.00€
- person B: 24.99€

Hard to explain to person B that 24.9975€ is rounded to 24.99€ when it is obvious it should be 25.00€.

---

This PR contains a failing test to show how to reproduce the problem. It is not intended to be merged as is, we need to categorize this as a bug (or not) and fix it (or not).


frederikbosch:
@mnapoli I believe we have seen this report earlier. I think it is domain specific to whom to allocate the remaining cent. The current behaviour is: allocate to the first one in the list with a ratio above 0. This also means you can sort your list of ratio's to change the one who gets the remainder.

mnapoli:
> This also means you can sort your list of ratio's to change the one who gets the remainder.

That is cleverly simple indeed! The library could do that though (and I'm not against implementing it myself if others confirm this is what we want).

I have been talking about this to a few people working with numbers and it seems the answer is not that obvious (I was convinced I found a bug, but I now have doubts).

Let's see if others have an opinion on this.

frederikbosch:
I am not sure we want this. But I am definitely not against it (now). It is something that the library could perfectly handle. My doubt is only how to keep it simple, without being to restrictive.

sagikazarmark:
I can think of an optional second argument:

`public function allocate(array $ratios, $allocation = self::ALLOCATE_FIRST)`

frederikbosch:
@sagikazarmark Yes, the alternative would be `public function allocate(array $ratios, \Closure $remainder = null)`, which would allow the user to implement it's own sorted algorithm.

```php
$money->allocate([33, 66], function ($a, $b) { return $b <=> $a });
```

Then, the caller has full control. But that might be overkill. If we use your method @sagikazarmark, then a user that wants full control must still sort the ratio list before calling `allocate`.

frederikbosch:
p.s. We have more than 2000 stars. Good work!

sagikazarmark:
I would rather create an additional function for that use case, internally one can call the other, but until there is actual use case for such behavior, implementing allocation to the first and the last targets might just be enough and definitely easier to use.

frederikbosch:
You're right. And I guess you mean first and largest, because I guess that's what @mnapoli 's intention is.

frederikbosch:
Replaced by #526.

PR #494, opened 2018-08-18
Parsing very very tiny float.

gmponos:
Hello there....

I have an API that returns executes a reports, does some calculations and returns a result. One of these results happened to be an amount that was a very very tiny one. `2.8865798640254e-15`

My code that consumes the API parses this value to a Money object automatically by doing:

```
$money = $parser->parse((string)$amount, $currency);
```

I know that there is no actual currency that could represent a so tiny amount and it is actually zero (maybe in some bitcoins it have some value).

So the code above throws exception because the value is converted to a string in this form: `2.8865798640254e-15` and this does not pass the expression inside the parser. For now in order to stop getting exception I had to do a quick solution to round the number to 7 decimals and then pass it inside the parser and it feels kinda hackish. 

As a better solution I could create something like a `ExponentialParser` and pass it inside an `AggregateParser` but before going into that I felt that I should ask if you believe that this should be handled by the DecimalParser and not by a different parser... 

If yes maybe I could implement the changes on this PR.

### DIFF
--- a/tests/Parser/DecimalMoneyParserTest.php
+++ b/tests/Parser/DecimalMoneyParserTest.php
@@ -5,6 +5,7 @@ namespace Tests\Money\Parser;
 use Money\Currencies;
 use Money\Currency;
 use Money\Exception\ParserException;
+use Money\Money;
 use Money\Parser\DecimalMoneyParser;
 use PHPUnit\Framework\TestCase;
 use Prophecy\Argument;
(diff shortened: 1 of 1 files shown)


frederikbosch:
@gmponos I have some questions before coming to a solution.
1. What does that API return: a string with lots of fractional numbers or a really exponential floating point?
2. How is that number represented in PHP: a float or a string? 
3. If it is a float, is it bigger than the constant `PHP_FLOAT_MAX ` (constant available since PHP 7.2)?

And maybe you can try this one in the meanwhile.
```php
use Money\Number;
$money = $parser->parse((string)Number::fromNumber($amount), $currency);
```

gmponos:
Hello, basically most of the questions can be answered through the test case that I have added.

On the test case I just simulated the same thing as my App is doing but let's take it from the top and answer your questions just to be specific here:

1. It returns it as exponential.
2. it is represented as a float. check the test.
3. I guess not because PHP_FLOAT_MIN is `2.2250738585072e-308` and I just have `e-15`

I will give a try to the `Number` but never the less should this be inside the DecimalParser?

gmponos:
Any idea how this should be handled after all?

frederikbosch:
Closed in favour of #520.

PR #495, opened 2018-08-24
Fix PhpCalculator

ricardogobbosouza:
Problem...

```php
$calculator = new PhpCalculator();
$result = $calculator->divide(30, 2);
```
$result = 15.00000000000000 instead of 15


mnapoli:
I think it is necessary to add new tests to cover the problem you are describing.

frederikbosch:
@mnapoli @sagikazarmark It does seems a better solution than what we had. I think this can be merged. How about you?

sagikazarmark:
LGTM, thanks.

PR #496, opened 2018-09-02
Remove setUp template method

samnela:


frederikbosch:
Thanks

gmponos:
Just out of curiosity.. is there any benefit by removing it other than getting rid of the extra function call?

frederikbosch:
Not that I am aware of. Usually I have the following standard for libraries with high code coverage: any code that can be removed and still passes tests, should be removed.

sagikazarmark:
I guess it can also be a readability thing: don't have to check what the parent does, because it does nothing.

Issue #498, opened 2018-09-05
Allow Converter to accept a CurrencyPair


jaikdean:
As per the current documentation, you can get a conversion rate and convert a `Money` object to another currency as follows:

```php
$exchangeRate = $exchange->quote(new Currency('GBP'), new Currency('USD'));
$tenPoundsInDollars = $converter->convert(Money::GBP(1000), new Currency('USD'));
```

We need to log the exchange rate that was used for a conversion alongside the converted value. We can do the above, but that relies on `$exchange->quote()` returning the same value when called by `Converter::convert()`, which in some cases may not be true.

I suggest we add a method to `Converter` to allow a value and a `CurrencyPair` to be passed in, rather than having `Converter` always call `Exchange::quote()`:

```php
$exchangeRate = $exchange->quote(new Currency('GBP'), new Currency('USD'));
$tenPoundsInDollars = $converter->convertFromCurrencyPair(Money::GBP(1000), $exchangeRate);
```

Padam87:
+1 Same here.

Our system is HUF based, but we allow customers to use other currencies for payment.

We store the exchange rate at the time of purchase.

We also log expenses for every order, and store the ratio at the time of logging those expenses.

1 to 4 currencies are usually involved for every order.

For our profit stats, we convert everything to HUF, using the stored ratios. This feature would make it a lot easier.

sagikazarmark:
This sounds like a nice addition to the library.

For now you can create a custom exchanger with fixed rates to ensure you always get the same result.

Where this method should go: I will have to think about it. It would be nice to keep it separated from the exchange stuff, so a separate class would probably be a better fit, but we will see.

In the meantime, feel free to submit a PR to see how it would work.

judgej:
So I understand, is it necessary to create a dummy `Exchange`, in order to create a converter, if you are only going to convert using `convertAgainstCurrencyPair()`? Does that method use the exchange? It doesn't look like it does. Creating a class that is necessary as a depenency, but is then not used, feels wrong. Should the exchange be an optional argument, or should `convertAgainstCurrencyPair()` be in a class of its own?

PR #502, opened 2018-09-17
Currency pair param for converter

Padam87:
https://github.com/moneyphp/money/issues/498

### DIFF
--- a/src/Converter.php
+++ b/src/Converter.php
@@ -30,16 +30,22 @@ final class Converter
     }
 
     /**
-     * @param Money    $money
-     * @param Currency $counterCurrency
-     * @param int      $roundingMode
+     * @param Money             $money
+     * @param Currency          $counterCurrency
+     * @param int               $roundingMode
+     * @param CurrencyPair|null $currencyPair
      *
      * @return Money
      */
-    public function convert(Money $money, Currency $counterCurrency, $roundingMode = Money::ROUND_HALF_UP)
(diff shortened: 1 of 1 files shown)


Padam87:
@sagikazarmark How about leaving the exchanges out of this, and just adding an optional 4th parameter?

src/Converter.php
```diff
@@ -30,16 +30,22 @@ public function __construct(Currencies $currencies, Exchange $exchange)
...
-     * @param int      $roundingMode
+     * @param Money             $money
+     * @param Currency          $counterCurrency
+     * @param int               $roundingMode
+     * @param CurrencyPair|null $currencyPair
      *
      * @return Money
      */
-    public function convert(Money $money, Currency $counterCurrency, $roundingMode = Money::ROUND_HALF_UP)
+    public function convert(Money $money, Currency $counterCurrency, $roundingMode = Money::ROUND_HALF_UP, CurrencyPair $currencyPair = null)
```
sagikazarmark: This is a bit problematic, because we would have to check that the currency pair has the same base and counter currencies as the first two arguments. What we really need here is the ratio.

I would probably extract this to a separate function that requires a ratio:

```
// Converts a money to another currency without exchanging a conversion ratio first.
staticConvert(Money $money, Currency $counterCurrency, $ratio, $roundingMode = Money::ROUND_HALF_UP)
```


Or something like that.

sagikazarmark:
@Padam87 I've commented on the PR, can you please take a look at it. Also, please rebase the PR from master.

frederikbosch:
Superseded by #644

PR #508, opened 2018-10-06
Add fallback locale for 'locale that uses comma' test

mishavantol:
- OSX uses this format for locales


mishavantol:
Or it could just be changed to 'UTF-8'. I see the other tests use that format too.

frederikbosch:
Thanks for the suggestion, I think we will leave it to current setup for now.

PR #510, opened 2018-10-18
add docs for aggragation function fixes #501

### LINKED ISSUE #501: Write docs for aggregation and sum
@frederikbosch Can you write some docs for these topics before we tag a release? Not sure if I'll have the time before next week, so if you get there before could you pick this one up?

andythedandy:
Is it ok like that?

### DIFF
--- a/doc/features/aggregation.rst
+++ b/doc/features/aggregation.rst
@@ -1,0 +1,44 @@
+.. _aggregation:
+
+Aggregation
+===========
+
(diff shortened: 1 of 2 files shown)


sagikazarmark:
Yeah, it looks great. Thanks!

frederikbosch:
@sagikazarmark @andythedandy Thanks for both your help.

PR #511, opened 2018-10-28
Update BitcoinMoneyFormatter.php

zek:


gmponos:
It seems that this PR was created by the fact that the BitcoinFormatter accepts a list of currencies. Checking the file of the formatter.. this raised also some questions to me..

1. Why is it needed to pass a currencies inside this formatter since it is specific for only one Currency which is Bitcoin and not in general about about crypto currency.
2, If this must apply `if (BitcoinCurrencies::CODE !== $money->getCurrency()->getCode())` then why is this needed:

```
        $subunit = $this->currencies->subunitFor($money->getCurrency());
        $valueBase = Number::roundMoneyValue($valueBase, $this->fractionDigits, $subunit);
```

We already know the subunit of the currency.. it's the subunit passed on the constructor..

frederikbosch:
@gmponos Reasonable explanation. I will close this PR in favour of #524.

PR #514, opened 2018-11-06
MoneyRange implementation

jaikdean:
We're using this library in a couple of projects, and have hit upon the need for a monetary range a number of times. While we can work around this, it's much cleaner using a `MoneyRange` value object.

I'm aware this is slightly extending the scope of this library, so it perhaps doesn't belong here and should be in a separate library, but it's been a useful addition for us.

### Examples

```php
$range = MoneyRange::GBP(100, 1000);

$range->contains(Money::GBP(200)); // returns true
$range->midPoint(); // returns Money::GBP(500)
$range->lessThan(Money::GBP(2000)); // returns true
$range->lessThan(MoneyRange::GBP(2000, 3000)); // returns true
$range->multiply(2); // returns MoneyRange::GBP(200, 2000)
```

I'd welcome feedback on this!


frederikbosch:
I don't know if this should be part of this library. For now I tagged it for milestone 4.1.0. We will decide then to merge this or not. Thank you for sharing this idea and code.

frederikbosch:
> have hit upon the need for a monetary range a number of times

Could you explain the use-case? It helps to accept the PR.

PR #516, opened 2018-11-14
Add php-decimal calculator

rtheunissen:
Hi everyone, this patch adds support for the new php decimal extension. 

It's a real shame how the calculators are set up, because it allows `float` to used (thereby opening the door to poor precision) and requires that the return value of each operation be `string`. The decimal extension uses an optimized internal value that is not a string, so the conversion to and from string for every operation adds unnecessary overhead. 

What I'd love to see is a PHP 7 update of this library where input is normalized (`string` or `int`, ideally) before being passed to the calculator, and to not require that the return result of the calculators be string.

This implementation passes the tests though, so it'll have to do for the time being.

Please see http://php-decimal.io for more information.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ sudo: false
 addons:
     apt_packages:
         - enchant
+        - libmpdec-dev
 
 cache:
     directories:
@@ -18,6 +19,7 @@ php:
     - 7.0
     - 7.1
     - 7.2
+    - 7.3
     - nightly
 
 env:
@@ -45,6 +47,7 @@ install:
     - if [[ "$COMPOSER_FLAGS" == *"--prefer-lowest"* ]]; then travis_retry composer update --prefer-dist --no-interaction --prefer-stable --quiet; fi
     - travis_retry composer update ${COMPOSER_FLAGS} --prefer-dist --no-interaction
     - pip install --user -r doc/requirements.txt
+    - pecl install decimal && echo "extension=decimal.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini || true
 
 script:
     - $TEST_COMMAND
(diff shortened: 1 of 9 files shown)


frederikbosch:
@rtheunissen Nice work over there with the extension, I like it. I was not aware of it.

@sagikazarmark I am in favour of merging this PR in the 3.x range. And I also think @rtheunissen is right regarding risks of float usage. We should take care of that in the 4.x releases.

rtheunissen:
There is a breaking change on the horizon for decimal so stand by for that before this is merged please. Likely to have that released next week sometime.

sagikazarmark:
Sorry for the late answer, time has passed quickly. I'm definitely in favor of moving to PHP 7.3+ and improving internal semantics.

One comment related to this particular PR: we have moved to GitHub Actions recently, so the CI part of the PR needs to be actualized. Other than that 👍 

Can you update the PR @rtheunissen ?

sagikazarmark:
> and to not require that the return result of the calculators be string.

What should the result be then?

rtheunissen:
> > and to not require that the return result of the calculators be string.
> 
> What should the result be then?

An abstract number type. The decimal extension does not use strings internally, so to convert to and from string for every operation is wasteful. Instead, we can create an interface like `Number` that supports whatever operations we need, and leave it up to the implementation to store the internal data however it wants. BC and any other string-based calculator could use a generic `StringNumber`, for example.

sagikazarmark:
Would this be an interface between `Money` and calculators? Or an internal representation that `Money` uses as well? (Currently `Money uses string internally)

frederikbosch:
Since we want to keep maintenance for this library not too high, I would recommend publishing this in a separate repo. I am very sorry for the delay.

PR #517, opened 2018-11-16
Remove symfony-phpunit-bridge

gmponos:


gmponos:
I tried to find a related PR but I couldn't and I thought it had no reason to be there. Now I see...

PR #519, opened 2018-12-02
Deprecate currency aggregator

gmponos:
Hello,

If someone is using his custom Parsers he will never know about the intention of Deprecating currency as a string.

### DIFF
--- a/src/Parser/BitcoinMoneyParser.php
+++ b/src/Parser/BitcoinMoneyParser.php
@@ -41,6 +41,10 @@ final class BitcoinMoneyParser implements MoneyParser
             throw new ParserException('Value cannot be parsed as Bitcoin');
         }
 
+        if ($forceCurrency !== null && !$forceCurrency instanceof Currency) {
+            @trigger_error('Passing a currency as string is deprecated since 3.1 and will be removed in 4.0. Please pass a '.Currency::class.' instance instead.', E_USER_DEPRECATED);
+        }
+
         $decimal = str_replace(BitcoinCurrencies::SYMBOL, '', $money);
         $decimalSeparator = strpos($decimal, '.');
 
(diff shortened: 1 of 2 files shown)


gmponos:
Now that I think of this... not sure if it is the best.. A user will get the deprecation twice if he used a Aggregator with the one you have inside here...

So that might be the reason of not having this feauture at the start.

frederikbosch:
@gmponos While you are correct, you are only partly correct. Someone with an own implementation of Currencies will never get the error. Rather twice than zero.

@sagikazarmark We also missed this for the Bitcoin parser.

@gmponos Can you update the PR to include Bitcoin too?

sagikazarmark:
Noone is going to get the error twice: the code in this change will convert the currency into a Currency object, so the underlying parsers won't trigger the error.

frederikbosch:
@sagikazarmark Ah yes, you are right, so this is a good solution.

gmponos:
Correct... will do the changes

src/Parser/BitcoinMoneyParser.php
```diff
@@ -41,6 +41,10 @@ public function parse($money, $forceCurrency = null)
             throw new ParserException('Value cannot be parsed as Bitcoin');
         }
 
+        if ($forceCurrency !== null && !$forceCurrency instanceof Currency) {
+            @trigger_error('Passing a currency as string is deprecated since 3.1 and will be removed in 4.0. Please pass a '.Currency::class.' instance instead.', E_USER_DEPRECATED);
+        }
+
```
gmponos: OK I have made the change to BitcoinMoneyParser as well.. just to be sure here,,, `$forceCurrency` is never used inside this parser... so not sure if it makes sense to have a deprecation in here...

cebe: see #540 , #539

frederikbosch:
Because of the merge of #540 this has some duplicate (not conflicting) code. Therefore I created #543 which will be merged when the lights are green. Closing this in favour of #543.

frederikbosch:
Thanks @gmponos for your work!

PR #520, opened 2018-12-02
Parse exponential float as Money

gmponos:
Created a new Parser that will be able to parse exponential values to Money object. This solves the issue that I have presented on this PR #494 

This parser is used only for exponential values and not for decimals.. It will be good to be used combined with an Aggregator parser and a decimal parser.


gmponos:
@money-team let me know if this going to the correct direction for you in order to finalize the PR and fix the CI failures.

frederikbosch:
@gmponos Why not inject a Decimal parser and then delegate the parsing process once you have handled the exponential part? Maybe I should have a closer look, but as it seems to me there is a lot of duplicate code?

```php
public function __construct(Currencies $currencies, DecimalMoneyParser $delegatedParser) {
   $this->currencies = $currencies;
   $this->delegatedParser = $delegatedParser;
}

public function parse($money, $forceCurrency = null) {
   // ....
  $number = number_format($expo, $subunit, '.', '');
  return $this->delegatedParser->parse($number, $forceCurrency);
}

**Comment by gmponos** [comment]
@frederikbosch I though about the same solution too. But I dislike it for the following reasons:

1. First of all one of the alternative solutions can be found [here](https://github.com/moneyphp/money/pull/529). The argument about duplication stands almost for every parser.
2. Secondly I will need to do some checks inside `exponential` parser and then the same checks will be repeated inside the `decimal` parser. For instance:

**ExponentialParser.php**
```
    public function parse($money, $forceCurrency = null)
    {
        if (!is_string($money)) {
            throw new ParserException('Formatted raw money should be string, e.g. 1.00');
        }

        if (null === $forceCurrency) {
            throw new ParserException(
                'ExponentialParser cannot parse currency symbols. Use forceCurrency argument'
            );
        }
       ....
      $this->delegatedParser->parse(...);
```

The same checks above will be checked again inside the `delegatedParser`

My current solution stands to the following facts:

1. Most of the parsers already have the same code.. written a little bit differently.. check the PR above.
2. I could implement the solution directly inside `DecimalParser` but my solution requires to use `number_format` function. if I did that most probably it would raise some questions like why the `DecimalParser` doesn't use the number_format function on the rest of the code. And also are values like these "1.234e-11" considered "Decimals".. Playing with words here...

**Comment by sagikazarmark** [comment]
A trait indeed seems to be an easier solution in this case as the delegated parser would always be a decimal parser.

File: src/Parser/ExponentialMoneyParser.php [hunk 1]
```diff
@@ -0,0 +1,120 @@
...
+use Money\MoneyParser;
+use Money\Number;
+
+/**
+ * Parses an exponential string into a Money object.
+ *
+ * @example `2.8865798640254e+15`
+ *
+ * @author George Mponos <gmponos@gmail.com>
+ * @author Teoh Han Hui <teohhanhui@gmail.com>
```
sagikazarmark:
```suggestion
```

src/Parser/ExponentialMoneyParser.php
```diff
@@ -0,0 +1,119 @@
...
+     */
+    public function parse($money, $forceCurrency = null)
+    {
+        if (!is_string($money)) {
+            throw new ParserException('Formatted raw money should be string, e.g. 1.00');
+        }
+
+        if (null === $forceCurrency) {
+            throw new ParserException(
+                'DecimalMoneyParser cannot parse currency symbols. Use forceCurrency argument'
```
mwgamble: Shouldn't this say "ExponentialMoneyParser", not "DecimalMoneyParser"?

gmponos: Done.. nice catch

frederikbosch:
@gmponos Could you update this PR against the latest master? Sorry for our very late action on this.

frederikbosch:
Superseded by #668.

Issue #465, opened 2018-02-14
Make Number and/or Calculators internal


sagikazarmark:
Currently we use Number internally for a lot of things, but as far as I can see it's rarely present in interfaces. Let's make it internal, so we can do whatever we want with it.

The calculator interface is a bit harder from this point of view. IMO we can leave it public, but in that case we cannot use Number in type hints.

bmeynell:
@sagikazarmark - I've found `Number` to be a nifty little utility for representing a [Decimal](https://en.wikipedia.org/wiki/Decimal) - somewhere between the "numeric" PHP primitives and a full-fledged `Money` object.

PR #521, opened 2018-12-03
Altered CurrencyPair to use a Money object rather than a currency+float

jonjonw:
This needs some fixes but I was hoping it could serve as a suggestion/starting-point.

I Changed CurrencyPair to use a Money object rather than a currency/ratio pair. This way the ratio is treated as a big integer rather than a float.

So for instance, a CurrencyPair would be created as

```
$baseCurrency = new Currency('USD');
$counterCurrency = new Currency('CAD');
$conversionRatio = new Money(12943, $counterCurrency);

$pair = new CurrencyPair($baseCurrency, $conversionRatio);
```
I think this makes sense since the ratio really is Money (the amount of a currency for 1.0 units of another currency) and the value is most likely going to be multiplied against Money objects.

To get it working, several of the static values I changed would have to be changed to ints/strings rather than floats. Some code would need to be added to handle conversion from iso X.XXXX values to integers/strings.

### DIFF
--- a/doc/features/currency-conversion.rst
+++ b/doc/features/currency-conversion.rst
@@ -3,7 +3,7 @@ Currency Conversion
 
 To convert a Money instance from one Currency to another, you need the Converter. This class depends on
 Currencies and Exchange. Exchange returns a `CurrencyPair`, which is the combination of the base
-currency, counter currency and the conversion ratio.
+currency and a Money instance indicating the counter currency and conversion ratio.
 
 
 Fixed Exchange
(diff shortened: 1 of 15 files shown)


UlrichEckhardt:
"the ratio really is Money" -- sorry, but that is IMHO not a good way to look at things: The conversion ratio from CUR1 to CUR2 is a number and its unit is CUR2/CUR1. Basically, a big aspect of this library is to couple a value with a unit, and the conversion ratio is nothing different, only that the unit is a ratio of currencies and not a single currency here.

Notes:
 * Try to look at it from the other side and switch the argument that combines currency and value. That way, you could as well say that it should take a base amount, which is the cost of a single unit of the target currency.
 * Combined, you could supply both currencies and the conversion ratio using two Money objects. That approach would at least preserve some symmetry.
 * And yes, you can construct a conversion ratio from different parameters. However, I'd leave the current way as it is and use a static factory function instead, like the `createFromIso()` function there.

sagikazarmark:
Thanks for the contribution, but I'm not really in favor of this change.

> the ratio really is Money

This is just incorrect. As @UlrichEckhardt pointed out the ratio is just a number.

frederikbosch:
This PR, while not correct in the solution, is pointing at a weak point of the library itself. This is also noticed in #516. We should use strings or an object like Number everywhere to prevent float usage. I am closing this. Discussion should take place in #465.

PR #522, opened 2018-12-03
Deprecate initializing currency with an empty string

gmponos:
Should it be allowed to initialize an empty currency?


frederikbosch:
I think this should throw an `InvalidArgumentException` because I don't feel this is BC.

gmponos:
Done.

frederikbosch:
Thanks!

PR #523, opened 2018-12-03
Avoid function call in toString function of currency

gmponos:


sagikazarmark:
I'm gonna merge it, because I don't have anything against it, but it's premature optimization.

sagikazarmark:
Thanks!

PR #527, opened 2018-12-05
Remove currencies from bitcoin formatter

### LINKED ISSUE #524: Remove Currencies from constructor BitcoinMoneyFormatter
It does not make sense to inject `Currencies` into since it is used for only 1 currency and all unknown factors are static.

frederikbosch:
Fixes #524.


gmponos:
Hello just thimking out loud here... why not requiring bitcoin currencies instead and not the interface and remove the fraction argument?

frederikbosch:
The fraction argument is how to format the Bitcoin, e.g. *display* cents or not. It is a different concept than the subunits of the currency.

sagikazarmark:
Well, technically it is a BC break. I would rather make it optional and trigger a deprecation error when something is passed.

PR #529, opened 2018-12-05
Proof of concept: Use a trait for parsers

gmponos:
Made all the parsers use the same code by using a trait..

At my initial commits I made all the parsers use the same code. And then later I replaced the code with a trait.

@frederikbosch made a comment [here](https://github.com/moneyphp/money/pull/520#issuecomment-444468991)

I had also thought about the same idea but later on my plan was to submit a PR just like this one.

Marking this trait as internal will also not be inside the BC promise.


Big-Shark:
@gmponos Any updates?

gmponos:
This could be my fault.. but I need some feedback in order to proceed and finalize this or not.

Neither on the current PR or [here](https://github.com/moneyphp/money/pull/520#issuecomment-444468991) I got any feedback.

Big-Shark:
@gmponos This is look very nice for me, but I am not scure, why we can not use `number_format` ?
Yasterday I created this class https://gist.github.com/Big-Shark/84434409e4f4cb6754645ea283e48c74
And all my tests passed.
Can you explorer this?

src/Parser/DecimalParserTrait.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Money\Parser;
+
+use Money\Currency;
+use Money\Exception\ParserException;
+use Money\Money;
+use Money\Number;
+
+trait DecimalParserTrait
```
sagikazarmark:
```suggestion
/**
 * Common decimal parsing logic extracted from DecimalMoneyParser for reuse.
 */
trait DecimalParser
```

src/Parser/DecimalParserTrait.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    public static $decimalPattern = '/^(?P<sign>-)?(?P<digits>0|[1-9]\d*)?\.?(?P<fraction>\d+)?$/';
```
sagikazarmark: This should be a const IMHO

src/Parser/DecimalMoneyParser.php
```diff
@@ -16,7 +14,7 @@
  */
 final class DecimalMoneyParser implements MoneyParser
 {
-    const DECIMAL_PATTERN = '/^(?P<sign>-)?(?P<digits>0|[1-9]\d*)?\.?(?P<fraction>\d+)?$/';
```
sagikazarmark: Note that removing this constant is probably a BC break.

sagikazarmark:
Like the idea! Can you please rebase the branch from master?

src/Parser/DecimalParserTrait.php
gmponos: Traits can not have constants :)

src/Parser/DecimalMoneyParser.php
gmponos: reverted

src/Parser/DecimalParserTrait.php
sagikazarmark: In that case, I would make it protected.

gmponos: Done.

gmponos:
Marked as WIP.. let's merge this first: https://github.com/moneyphp/money/pull/586

frederikbosch:
Sorry @gmponos, I think we are going to leave this as is. In my opinion traits are to be avoided.

PR #530, opened 2018-12-05
PHPStorm suggests to require ext-json

gmponos:


frederikbosch:
Can you add it to suggest? It is not required.

gmponos:
You have Currency and Money object implement `JsonSerializable`.

http://php.net/manual/en/class.jsonserializable.php

Now that I've checked it online I see that PHP says that json extension is compiled into PHP by default

see http://php.net/manual/en/json.installation.php

🤔 so why does PHPStorm suggests this? Is it possible for someone to compile PHP without this extensions?

browner12:
correct, while 99% of PHP installs are going to have the JSON extension compiled in by default, it is possible to omit it.

I would agree that this belongs in 'require' since a package class is implementing `JsonSerializable`.

sagikazarmark:
Yeah, it looks like it's technically required.

gmponos:
it just occurred to me... https://twitter.com/gmponos/status/1073611157792006144

PR #533, opened 2018-12-10
Update generate-money-factory.php

golubkovden:
Allows to use int (for IDE hints)


mnapoli:
Yes please :)

We have red errors everywhere in PhpStorm now with the latest release.

PR #537, opened 2018-12-26
Allow numeric code

kunicmarko20:
This is one way of doing it, an alternative would be providing new "Currencies" class that can handle both and deprecating this one, but since this is not a BC Break I don't think that is needed.

### DIFF
--- a/src/Currencies/ISOCurrencies.php
+++ b/src/Currencies/ISOCurrencies.php
@@ -30,6 +30,7 @@ final class ISOCurrencies implements Currencies
 
     /**
      * @param string $code
+     *
      * @return array|null
      */
     private function findCurrencyByCode($code)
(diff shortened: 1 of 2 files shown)


kunicmarko20:
If this gets accepted, the only other thing I would propose is adding `alphabeticCodeFor` since there is `numericCodeFor`

kunicmarko20:
@frederikbosch is there any chance to get some feedback on this and maybe include it into the next release?

frederikbosch:
It could be included in the next release feature release (3.3.x). 

@sagikazarmark What do you think of this feature. Basically the wish is to use numeric codes as `Currency`, like below.

```php
$money = new Money(100, new Currency(123));
$currencies->contains($money->getCurrency()); // also checks against numeric code
```

Right @kunicmarko20?

If we want to have this feature, should we change the code of `ISOCurrencies` or should there be a new `NumericISOCurrencies` class? My feeling is that the latter is a better solution.

Issue #541, opened 2019-01-08
Invalid value parsed in BitcoinMoneyParser


cebe:
Parsing `0.020000000` with 2 decimals results in an amount value of `20000000` instead of `2`.

cebe:
Is there any reason why the BitcoinMoneyParser does not use the DecimalMoneyParser internally?

cebe:
Added a failing test in #542, it could probably be solved by using DecimalMoneyParser internally, but I am unsure about the reasons why it is not used already.

PR #542, opened 2019-01-08
Added failing test for BitcoinMoneyParser

cebe:
see issue #541 

Also added tests for DecimalMoneyParser which does not seem to be affected by this issue.


cebe:
probably solved by #529

PR #545, opened 2019-01-09
prepare new release

frederikbosch:
Modifies changelog to contain changes since last release


frederikbosch:
@sagikazarmark Current changes to master since last release are only bug fixes. I suggest to release `3.2.1`. What do you think?

PR #546, opened 2019-01-09
fix gmp division of negative

### LINKED ISSUE #538: Divisions throws InvalidArgumentException "Invalid fractional part" when using GmpCalculator
```
$value = new Money('500', new Currency('USD'));
$p = $value->divide(110); //throws exception
```

frederikbosch:
Fixes #538

### DIFF
--- a/src/Calculator/GmpCalculator.php
+++ b/src/Calculator/GmpCalculator.php
@@ -136,6 +136,10 @@ final class GmpCalculator implements Calculator
             )
         );
 
+        if ($divisionOfRemainder[0] === '-') {
+            $divisionOfRemainder = substr($divisionOfRemainder, 1);
+        }
+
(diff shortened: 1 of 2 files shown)


frederikbosch:
@sagikazarmark When dividing the a negative amount the outcomes are different. The calculation `-500 / 110` results in `-4.54545454545455` according to PHP and GMP, whereas BC says it is `-4.54545454545454`. I think that BC Math is not rounding, but using the scale argument to determine the number of fractional digits.

frederikbosch:
The solution could be.

1. Only add the negative division test to the GMP calculator.
2. Another fix.

hellodave76:
> The solution could be.

Seeing as the problem doesn't originate inside moneyphp, maybe an acceptable solution would be to loosen the test condition, like trimming the values down to 12 decimal places?

frederikbosch:
@hellodave76 Good suggestion, merged.

PR #550, opened 2019-01-30
Update MoneyFactory annotations

alexdo:
Hi 👋 

the MoneyFactory has been updated to include `int` in its typehints in #533, but the updated generator hasn't been executed yet ;) 
I ran it once. This change fixes IDE hints (like this one from PHPStorm):

bildschirmfoto 2019-01-30 um 15 51 03

Would be great to see this in the next patch!

Cheers 🍻 ,
Alex


frederikbosch:
Cool thanks

PR #551, opened 2019-02-19
Pure json response from jsonSerialize

### LINKED ISSUE #547: Money::jsonSerialize() should return currency code instead of Currency instance
Hi,
Both `Money` and `Currency` classes implement the `\JsonSerializable`. 👍
When the `Money::jsonSerialize()` is used in a `(string)` context, `\json_encode` returns the proper JSON representations, since to `Currency::jsonSerialize()` is implicitly called.
But if the `Money::jsonSerialize()` is called _alone_, it returns an instance of `Currency` for the `currency` key, instead of its `code`.

IMHO opinion, [this](https://github.com/moneyphp/money/blob/0ef4768e3a62166522081bfcc0feba00679268ff/src/Money.php#L521) could be:
`'currency' => $this->currency->getCode(),`.

zek:
Fixes #547

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -518,7 +518,7 @@ final class Money implements \JsonSerializable
     {
         return [
             'amount' => $this->amount,
-            'currency' => $this->currency,
+            'currency' => $this->currency->jsonSerialize(),
         ];
     }
 
--- a/tests/MoneyTest.php
+++ b/tests/MoneyTest.php
@@ -215,6 +215,11 @@ final class MoneyTest extends TestCase
             '{"amount":"350","currency":"EUR"}',
             json_encode(Money::EUR(350))
         );
+
+        $this->assertEquals(
+            ["amount" => "350", "currency" => "EUR"],
+            Money::EUR(350)->jsonSerialize()
+        );
     }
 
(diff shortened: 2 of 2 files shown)


frederikbosch:
@zek Could you include a test for this? One that failed in the situation before your change, and now works with your change.

zek:
> One that failed in the situation before your change, and now works with your change.
I am not familiar with testing. I don't know how can I test "before your change" part but I added second part. 

This PR actually is not critical but before this PR jsonSerialize function returns currency object.

~~~
array(2) { ["amount"]=> string(3) "100" ["currency"]=> object(Money\Currency)#738 (1) { ["code":"Money\Currency":private]=> string(3) "EUR" } }
~~~

I believe we must return

~~~
array(2) { ["amount"]=> string(3) "300" ["currency"]=> string(3) "EUR" }
~~~

frederikbosch:
Thanks, look good to me. @sagikazarmark What do you think?

frederikbosch:
Thanks @zek!

alle:
@zek @frederikbosch Thank you very much, guys! ☺️

PR #553, opened 2019-02-28
Update Class Name in Sample Code

deminy:
This is to fix an incorrect class name in some sample code.


frederikbosch:
Good catch, thanks

PR #555, opened 2019-03-13
Fix documentation issue with AggregateMoneyFormatter

Firehed:
I tried to run the example as-is, and there were a couple issues around missing parameters and wrong names. This updates the docs so that the example can be copy-paste run, and indicates the correct output for the bitcoin example (which was off by several orders of magnitude)


frederikbosch:
Thanks!

Issue #558, opened 2019-05-26
Remove bcmath dependency


terrafrost:
See https://github.com/phpseclib/bcmath_compat .

It basically turns phpseclib's BigInteger class into a BigDecimal class.

frederikbosch:
It’s only require-dev. Right? Nice project btw.

terrafrost:
> It’s only require-dev. Right?

Yah - good point.

I went ahead and created a PR: #559 

> Nice project btw.

Thanks!!

PR #559, opened 2019-05-29
replace ext-bcmath with phpseclib/bcmath_compat

terrafrost:
See #558


frederikbosch:
I am not sure why we would want this? I rather depend on the extension than on a compat lib. We are only talking of `require-dev` for bcmath. It is not a dependency that is required by `require`.

PR #562, opened 2019-07-09
Deserialization from var_export()

Grundik:
It would be nice, if money objects could deserialize themselves from var_dump() export format.


UlrichEckhardt:
Just one comment: The name of the tests is misleading, since no `var_dump()` is used. I'm at a loss if you ask me how to use that function though, without hacking into the output streams and `eval()`ing the result. I'm not sure, but maybe `var_export()` works, although it seems to use a different format at first glance.

Grundik:
I'm sorry, of course it is var_export. I've changed comments and added some more tests.

Grundik:
And also I've squashed all commits in one.

Grundik:
Use case of this scenario are configuration caching by laminas-config-aggregator: it aggregates multiple config files into configuration object, and caches it by dumping var_export into php file. So subsequent requests just includes that file, which are stored by opcode cacher, making very fast loading of any complexity configurations. But if there are objects in such config, which are not implements __set_state method, they would only be loaded first time, but not from the such cache. So it would be nice to be able to use Money objects in configuration.

frederikbosch:
`Money` is a value object. You can use a provided parser and formatter to serialize it.  Or you can implement your own. I don't see why this package should be functionality for `var_export`.

PR #570, opened 2019-10-01
Standard PHP formatter overrides fractions

Deltachaos:
The Standard PHP formatter overrides the fractions when using a currency object that has different fractions that what is defined in the icu lib on the operating system.

Therefore the `IntlMoneyFormatter` should use the same fraction for the PHP number formatter call as it has detected before.


frederikbosch:
You are supposed to inject the correctly constructed `NumberFormatter` object.

PR #573, opened 2019-10-08
[DOCS] Updates parsing example in docs

eduarguz:
As #543 added deprecation warnings for passing currency as a `string` in the `parse` method of current Parsers, it makes sense that the docs use the recommended way of passing currency to the `parse` method.

Thanks for amazing package!


sagikazarmark:
Thanks @eduarguz !

PR #580, opened 2019-12-27
Improve CI

### LINKED ISSUE #579: Add more checks
- PHPStan
- PHP CS Fixer
- Roave BC Check

sagikazarmark:
Fixes #579


composer.json
```diff
@@ -24,30 +28,38 @@
...
         "florianv/exchanger": "Exchange rates library for PHP",
+        "florianv/swap": "Exchange rates library for PHP",
         "psr/cache-implementation": "Used for Currency caching"
     },
+    "config": {
+        "sort-packages": true
+    },
+    "extra": {
+        "branch-alias": {
+            "dev-master": "3.0-dev"
```
gmponos: maybe it is a good chance to change this as well..

.github/workflows/static.yml
```diff
@@ -0,0 +1,59 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v1
+
+      - name: PHP-CS-Fixer
+        uses: docker://oskarstark/php-cs-fixer-ga:2.14.0
+        with:
+          args: --dry-run --diff-format udiff
```
gmponos: :/ what are the rules that this is using? is it me or a .php_cs.dist file does not exist on the current PR?

sagikazarmark: Rules are already present in the repository

PR #585, opened 2019-12-28
Update PHP action

sagikazarmark:


gmponos:
btw @sagikazarmark I see that pretty ci is still running.. is it needed?

sagikazarmark:
I don't mind it being there for the moment.

PR #586, opened 2019-12-28
Improve tests of parsers

gmponos:
before merging #529 if we could merge this first..


gmponos:
In short description:

-  Added more cases to `IntlLocalizedDecimalParserTest`
https://github.com/moneyphp/money/pull/586/files#diff-380f41fd51dd5f28ab63197ddc29fe85R116
- Replaces with equals https://github.com/moneyphp/money/pull/586/files#diff-380f41fd51dd5f28ab63197ddc29fe85R57
- Use actual Currency object instead of a Prophecy one: https://github.com/moneyphp/money/pull/586/files#diff-380f41fd51dd5f28ab63197ddc29fe85R23

sagikazarmark:
Why replace mocks with actual implementations?

gmponos:
> Why replace mocks with actual implementations?

For the usual reasons.. hahaah

- I have seen cases in the past where there were many WTF moments because of mocks... For instance say that it throws an exception although it never does.. for the sake of good programming practices I switched here to actual implementation.
- Although speed is not a problem here.. `Prophesize` is way heavier.

Let's not make a big deal out of it.. if this is no good I will revert.. no problem.

sagikazarmark:
Even if we consider replacing mocks with actual implementations, that should be a separate PR, probably covering the whole library.

gmponos:
Revert the test changes :)

sagikazarmark (approved):
Thanks!

PR #594, opened 2020-02-05
Absolute difference

eekes:
I was working on a project where I needed the absolute difference of 2 Money objects so I thought it would be useful adding a method for it 🙂

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -514,6 +514,20 @@ final class Money implements \JsonSerializable
     }
 
     /**
+     * Returns a new Money object that represents
+     * the absolute difference between this and an other Money object
+     *
+     * @param \Money\Money $subtrahend
+     * @return \Money\Money
+     */
+    public function difference(Money $subtrahend)
+    {
+        $this->assertSameCurrency($subtrahend);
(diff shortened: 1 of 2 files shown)


UlrichEckhardt:
There's an `absolute()` function which I'd suggest you use for that. The code then looks like:
```
return $this->subtract($subtrahend)->absolute();
```
This has the added benefit that it builds on other, proven functions instead of treating a string (`$this->amount`) like an integer. That integer may be unable to accurately represent the value,  which is why there are arbitrary-precision calculator implementations (see static `$calculator` member). Lastly, `difference` has an existing meaning. I'd rather call this `absoluteDifference()` to make that clear.

Even with the above adjusted, I'd still rather not add this. For one, it can be implemented based on the existing interface. Further, its use case is a corner case (I can't even imagine what you need that for).

eekes:
@UlrichEckhardt  The reason I need this is for comparing two big sets of data and searching for inconsistencies (and being able to display how much they differ).

But I didn't know about the `absolute` function, so this can be closed I think, thanks for the info 🙂

Issue #589, opened 2020-01-17
PHP 7.4 support


benr77:
It seems there's not yet any support for PHP 7.4? I'm getting errors such as

`Trying to access array offset on value of type int`

e.g. This is related to src/Formatter/IntlMoneyFormatter.php (line 44) 

`if ($valueBase[0] === '-') {`

PHP 7.4 does not permit accessing integer values with array access any more.

I've tried replacing the above with 

`if (substr($valueBase, 0, 1) === '-') {`

This works, but it feels wrong using a string function on an integer value.

Any thoughts? I'm happy to submit a pull request.

UlrichEckhardt:
I searched https://www.php.net/ChangeLog-7.php#7.4.0 and only came up with 
https://wiki.php.net/rfc/deprecate_curly_braces_array_access, but that doesn't explain what you're seeing...

benr77:
I believe it's the first entry on this page https://www.php.net/manual/en/migration74.incompatible.php

UlrichEckhardt:
Ah, so the problem is that `$valueBase` is of type `int`. Using `substr()` then implicitly causes a conversion to string, so this works again, but that's probably not a "proper" solution. I'd like to understand the context of this error a bit better though.

benr77:
Yes, but it only does a type conversion in order to do the comparison to determine the sign of the number. `$valueBase` is still an integer after the substr statement.

However, inside the `if` statement there's already a `substr()` in operation, and the resulting value of $valueBase is converted to a string anyway, as it is the return value of `substr()` - meaning depending upon the initial sign of $valueBase, after the `ìf` statement it could be either a string or an integer.
```
if ('-' === $valueBase[0]) {
    $negative = true;
    $valueBase = substr($valueBase, 1);
}
```
Surely we can just replace this with integer only processing to the same effect:

```
if ($valueBase < 0) {
    $negative = true;
    $valueBase = $valueBase * -1;
}
```
Here, $valueBase will start and end up as an integer value.

benr77:
Looking across the codebase, it seems all of the cases where some $var is accessed as an array, it is already a string, or is treated as a string immediately afterwards.

I can't see any good reason not to just switch to using `substr()` to determine the negativity in all these cases. 

Any thoughts?

UlrichEckhardt:
[...]

PR #595, opened 2020-02-12
Cast integer amount to string

toooni:
`$valueBase` is used as a string. To be strict we have to cast the amount to string first.


UlrichEckhardt:
That's wrong, `Money::getAmount()` returns a string.

UlrichEckhardt:
Maybe https://github.com/moneyphp/money/issues/589 is similar to what you're doing, but at least a few things are discussed there in depth.

toooni:
You're right. Sorry 👍

PR #610, opened 2020-06-21
Simplify constructor, whitespace not allowed in amount, fixes #575

### LINKED ISSUE #575: Wrong amount set causes unexpected comparison behaviour
Hey there!
Thank you for this library, it is really great!

I got into something really subtle.

```php
$money1 = Money::EUR("5000\n");
$money2 = Money::EUR("5000");
var_dump($money1->greaterThan($money2)); // prints 1
```

These 3 lines works and the result is `1`. Is it the expected behaviour or should there be checks while parsing the numeric string representation?

My `moneyphp/money` version is `3.2.1` and the php version is 7.3, but it can be reproduced in prior php versions

frederikbosch:


frederikbosch:
Fixed by #634

PR #621, opened 2020-12-23
Documentation: Add docs for compare() method

### LINKED ISSUE #620: Money->compare() is missing in documentation
This method isn't being documented anywhere:

https://github.com/moneyphp/money/blob/512333ef834bd77f76e865a30eb5963d480e2ed7/src/Money.php#L128-L140

half0wl:
Resolves #620

The lack of documentation and its usage makes me think that this is supposed to be internal, so I'm not too sure about this. Regardless, I think this should be documented since it's a public method and part of the API.

This can be useful over the other (more straightforward) comparison methods if you're doing a sort, especially with PHP8's corresponding deprecation [0]:

>  Sort comparison functions that return true or false will now throw a deprecation warning, and should be replaced with an implementation that returns an integer less than, equal to, or greater than zero. 

[0] https://www.php.net/manual/en/migration80.deprecated.php

### DIFF
--- a/doc/features/comparison.rst
+++ b/doc/features/comparison.rst
@@ -35,6 +35,29 @@ Equality
     $result = $value1->equals($value2);     // true
     $result = $value1->equals($value3);     // false
 
+.. _compare:
+
+Compare
+-------
+
+``compare()`` returns an integer indicating whether the first Money object is less than,
+equal to, or greater than the second. Both Money objects **must** have the same currency.
+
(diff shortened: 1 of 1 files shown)


half0wl:
Rendered `doc/features/comparison.rst`:

Screenshot 2020-12-23 at 8 17 35 AM

sagikazarmark (approved):
Thanks @half0wl

PR #622, opened 2020-12-30
Add php8 support

### LINKED ISSUE #619: PHP 8 compatibility
PHP 8 is being released today. So far it's not possible to run full test suite for MoneyPHP on PHP 8. I've identified following issues:
* `cache/taggable-cache` is unavailable for PHP 8, looks like there's a read-only repo
* PHPUnit 9 is required, which forces dropping support for PHPUnit 5, and this forces dropping PHP 5.6 (see https://phpunit.de/supported-versions.html)
* waiting for an update of `florianv/swap`

As dropping PHP 5.6 is required, I believe this ticket will be related to #393

sspat:
Closes #619 

- PHP 5.6, 7.0, 7.1 and 7.2 had to be dropped, mostly due to testing tools versions that support 8.0 dropping them, main issues were with phpspec.
- Some dataprovider value types had to be changed from string to float because of a bc break in `sebastianbergmann/comparator`, see https://github.com/sebastianbergmann/phpunit/issues/3185
- Fixed running Psalm in CI using `vimeo/psalm-github-actions`, there were multiple errors: https://github.com/moneyphp/money/runs/1597930547?check_suite_focus=true => https://github.com/sspat/money/runs/1627743511?check_suite_focus=true 
- Symfony PHPUnit bridge was removed because there is no point left in it after dropping the older php versions
- Fixed usage of deprecated methods in phpunit tests


sagikazarmark:
@sspat thanks! I'm gonna have to think about dropping PHP 5.6/7.x support a little (ie. should we make a new major release for that?). Other than that 👍

sspat:
taggable-cache, exchanger, swap - all had major bumps with some api changes and deprecations so even if money will be updated as a minor version change to get php8 support, these dependencies will still need a major update with breaking changes.

PowerKiKi:
> should we make a new major release

No, this should not be a major release, because `moneyphp/money` is not changed by this PR. Composer will take care of selecting the appropriate version of  `moneyphp/money` dependending on PHP version. There is not need to make this a major version.

And since it already is mentioned clearly in the changelog, there is nothing else to do than merge and and release as minor version.

frederikbosch:
Only v4 will support PHP8.0+. Thank you for your effort and I am sorry I did not pay attention to this earlier on.

PR #623, opened 2021-01-21
Rremove unnecessary ratio validation logic.

tabennett:
- Removed ratio validation logic that checks that the sum of ratios is non-negative.
- Removed ratio validation logic that prevented negative ratios.

Both of these methods currently block valid use cases for the allocate method.
In general, it should always be possible to allocate ratios as long as there is at least one non-negative ratio.
Guarding against this invariant should not be the responsibility of the money object.

### DIFF
--- a/spec/MoneySpec.php
+++ b/spec/MoneySpec.php
@@ -335,16 +335,6 @@ final class MoneySpec extends ObjectBehavior
         $this->shouldThrow(\InvalidArgumentException::class)->duringAllocate([]);
     }
 
-    function it_throws_an_exception_when_allocation_ratio_is_negative()
-    {
-        $this->shouldThrow(\InvalidArgumentException::class)->duringAllocate([-1]);
-    }
-
-    function it_throws_an_exception_when_allocation_total_is_zero()
-    {
-        $this->shouldThrow(\InvalidArgumentException::class)->duringAllocate([0, 0]);
-    }
-
     function it_throws_an_exception_when_allocate_to_target_is_less_than_or_equals_zero()
     {
         $this->shouldThrow(\InvalidArgumentException::class)->duringAllocateTo(-1);
--- a/src/Money.php
+++ b/src/Money.php
@@ -352,14 +352,7 @@ final class Money implements \JsonSerializable
         $results = [];
         $total = array_sum($ratios);
 
-        if ($total <= 0) {
-            throw new \InvalidArgumentException('Cannot allocate to none, sum of ratios must be greater than zero');
-        }
-
         foreach ($ratios as $key => $ratio) {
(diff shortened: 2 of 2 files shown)


frederikbosch (approved):

frederikbosch:
@sagikazarmark I am OK with merging this. What is up with the psalm analysis? I only use PHPStan on other projects, so I cannot tell.

tabennett:
I'm not sure what's going on there. I don't use psalm either.

frederikbosch:
Will look at this again after #634 is merged.

tabennett:
@frederikbosch  Is there a specific reason why? That branch doesn't look like it's going to be ready to release anytime soon and I PR'd this over 3 months ago (and that was after first explaining this issue and being asked if I could PR it).

#634 Doesn't fix this issue.  It's still throwing the same exception [here](https://github.com/Ocramius/money/blob/91c8091eb4f02497d225c717b5af90116e757c2f/src/Money.php#L292).

frederikbosch:
@tabennett Would you want to have a look at this again? What do you do when the total of the ratios is zero? Suppose you want to allocate 10  to 0, -1 and 1. What do you expect the outcome to be? I will only consider this when there is no BC break and working tests. This PR is closed until the mentioned criteria are met.

tabennett:
So I've been thinking about this.  It looks like removing [this check](https://github.com/moneyphp/money/blob/59837e34d7476c0f416b84f42e5f790a544de942/src/Money.php#L301) should be all that's necessary.

frederikbosch:
@tabennett Well yes. But can you provide (unit / spec) tests which proves that the system calculates the correct answer. Again, suppose you have `new Money(10, new Currency('EUR'))`. And you call `$money->allocate([0, -1, 1])`, what do you expect the outcome to be?

tabennett:
@frederikbosch I expect that 
```php Money::EUR(10)->allocate([0, -1, 1])```
throws an exception before the suggested changes that I would PR are even executed.  

I'm suggesting removal of lines 301 through 302 only.  This change would have no affect on the example you're asking me about...

frederikbosch:
Well, in the PR you are suggesting more lines to be removed. But suppose you want [only these specific lines to be removed](https://github.com/moneyphp/money/blob/a53900c697dab083d9a0dbac41ad5ddc3d7d4a93/src/Money.php#L301:L303), then I am wondering what the outcome of `$money->allocate([0, -1, 4])` would be. Removing those lines would allow negative ratios while the total still must be higher than zero, as in the example.

If you can explain me a scenario, the logic why that outcome is logic and we can capture that in a test, then I am happy to merge that.

x157660503 (commented):
這是我的 x157660503

PR #625, opened 2021-03-10
Fix Money\Currency Json serialization

jaumebecks:
`Currency` JSON serialization should return an indexed array, as it should be treated as an object in JSON.
Therefore, when serializing a Money object, the result should be like

```json
"moneyInstance": {
    "amount": "1000",
    "currency": { 
        "code": "EUR"
    }
}
```

Instead of 

```json
"moneyInstance": {
    "amount": "1000",
    "currency":  "EUR"
}
```

### DIFF
--- a/src/Currency.php
+++ b/src/Currency.php
@@ -82,7 +82,7 @@ final class Currency implements \JsonSerializable
     public function jsonSerialize()
     {
         return [
-            "code" => $this->code,
+            'code' => $this->code,
         ];
     }
 }
--- a/src/Currency.php
+++ b/src/Currency.php
@@ -77,10 +77,12 @@ final class Currency implements \JsonSerializable
     /**
      * {@inheritdoc}
      *
-     * @return string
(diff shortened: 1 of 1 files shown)


sagikazarmark:
> `Currency` JSON serialization should return an indexed array, as it should be treated as an object in JSON.

I don't think that's necessarily true. The format is arbitrary, nothing says the currency has to be an object. Additionally, this is a breaking change.

jaumebecks:
@sagikazarmark It's true that it is a breaking change, and that shouldn't be an impediment to try to improve the code. 

And following the current Entity structure, the default Object formatting should respect the JSON serialization. Money::$currency is a `Currency` object, and should be reflected as so in JSON, otherwise serialization/deserialization need special serializers/deserializers to manipulate the data. Another thing is that it's intended to apply such change

frederikbosch:
Unnecessary breaking change. Closing.

PR #627, opened 2021-03-11
Update assertRoundingMode function

Glancu:
Add  | to assertRoundingMode function when routingMode is not in array


frederikbosch:
In 4.x this can be validated with a type-checker.

PR #632, opened 2021-04-09
Variadic Money::isSameCurrency()

gsteel:
I frequently need to check that multiple money instances are all the same currency - making `isSameCurrency()` variadic seemed like an innocuous and safe addition to me! 🤞

### DIFF
--- a/doc/features/comparison.rst
+++ b/doc/features/comparison.rst
@@ -8,7 +8,7 @@ A number of built in methods are available for comparing Money objects.
 Same Currency
 -------------
 
-``isSameCurrency()`` compares whether two Money objects have the same currency.
+``isSameCurrency()`` compares whether two or more Money objects have the same currency.
 
 .. code-block:: php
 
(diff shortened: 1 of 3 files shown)


frederikbosch:
Looks good, will be merged for 4.0, after #634 is merged. I am wondering though, should we not expect at least one other `Money` object and thus make the first argument required? I believe the method now works without any method, no?

gsteel:
> Looks good, will be merged for 4.0, after #634 is merged. I am wondering though, should we not expect at least one other `Money` object and thus make the first argument required? I believe the method now works without any method, no?

Do you mean asserting that `count($other) >= 1` ?

I guess that would make it obvious to consumers in the case that they pass an empty iterable, and as it's targeting 4, the BC break of throwing an exception would be no big deal right? In that case, which exception should be thrown? PHP's `\BadMethodCall` or a specific (new) type?

I'd personally like to see this in 3.x because it'll be a while before I'm able to go to v8 in most of the project that use this lib.

frederikbosch:
Never mind. I was thinking of below, but on a second thought, let's leave it like this. I will merge it both v3 and v4 when I start processing PRs.

```php
public function isSameCurrency(Money $first, Money ...$other) {}
```

frederikbosch:
Sorry, I decided not merge this with v3. All new features go to the v4 branch.

frederikbosch:
Thanks for your effort, the feature in merged in v4.

PR #634, opened 2021-04-13
Quick money (butchering this library for the glory of satan, and CPU cycles)

### LINKED ISSUE #7: Deal with decimals using PreciseMoney

### LINKED ISSUE #258: Unspecified behavior of Money::allocate()

### LINKED ISSUE #465: Make Number and/or Calculators internal

### LINKED ISSUE #474: Construct Money from JSON stdClass
The `Money` object can be serialized to JSON, however there is no way to construct it back.

Can we add a method `public static function Money::fromJson(stdClass $object): Money`?

I prefer taking a `stdClass` as the json decode is not a reposibility of `Money` and it may be part of some deeper nested structure.

### LINKED ISSUE #575: Wrong amount set causes unexpected comparison behaviour

### LINKED ISSUE #583: Update cached currencies
The underlying cache package changed a lot since 0.4.0: https://github.com/php-cache/taggable-cache

We should probably relocate the implementation to an external package as well, since it has an external dependency.

### LINKED ISSUE #615: Allow null amount
A null amount is already allowed by lack of type enforcing, but phpdoc is suggesting that is not.
I think it should be allowed, so when types will be enforced this can be taken into account.

### LINKED ISSUE #619: PHP 8 compatibility

### LINKED ISSUE #633: `moneyphp/money` declares compatibility with PHP >5.6, but cannot guarantee compatibility with PHP 9 or later
Discovered by digging around that the version compatibility range for `php` allows for versions that are un-tested, as well as potentially just broken:

https://github.com/moneyphp/money/blob/a44e2f2bb6bf048596af1649cb7c51dd421a979a/composer.json#L27

Ocramius:
This is a draft: still playing around with it.

Ultimately, I'm mostly toying around with the library, removing bits that can be removed without substantial BC Breaks (yes, there will be BC breaks) and inlining code that can be inlined (yes, the code will be inlined).

I suggest avoiding an immediate code review until this is ready and I have stable performance results (most of the benchmarks were performed while my PC was under high load, so the values in the commit messages are not reliable).

In addition to all this, I still need to check if the GMP calculator can be used instead of BcMath, in order of priority, since GMP seems to generally be a bit quicker.

## Notable improvements:

 * [x] full type coverage (100% type coverage by `vimeo/psalm`)
 * [x] 80%+ mutation test coverage (`roave/infection-static-analysis-plugin`)
 * [x] removal of a lot of internal API
 * [x] roughly 50% performance improvements in happy-path usage of `Money`, `Number` and `Currency`
 * [x] removal of `Calculator` discovery by-default (you want a custom calculator? configure it manually!)

## Benchmark results

```
❯ php -dopcache.enable_cli=1 ./vendor/bin/phpbench run --ref=ref_original_5 --retry-threshold=3 --iterations=15 --revs=1000 --warmup=2
PHPBench @git_tag@ running benchmarks...
with configuration file: /home/ocramius/Documents/moneyphp/money/phpbench.json
with PHP version 8.0.3, xdebug ❌, opcache ✔

\Benchmark\Money\NumberInstantiationBench

    benchConstructorWithZeroIntegerAmount...R1 I2 - [Mo0.210μs vs Mo0.163μs] +28.42% (±1.29%)
    benchConstructorWithPositiveIntegerAmou.R1 I8 - [Mo0.308μs vs Mo0.626μs] -50.87% (±0.82%)
    benchConstructorWithNegativeIntegerAmou.R2 I2 - [Mo0.305μs vs Mo0.683μs] -55.40% (±0.90%)
    benchConstructorWithZeroAndFractionalAm.R1 I7 - [Mo0.316μs vs Mo0.576μs] -45.04% (±0.78%)
    benchConstructorWithFractionalAmount....R1 I14 - [Mo0.436μs vs Mo1.144μs] -61.85% (±0.80%)
    benchConstructorWithNegativeFractionalA.R1 I7 - [Mo0.435μs vs Mo1.200μs] -63.75% (±1.15%)
    benchFromStringWithZeroIntegerAmount....R1 I3 - [Mo0.319μs vs Mo0.208μs] +53.54% (±0.99%)
    benchFromStringWithPositiveIntegerAmoun.R1 I4 - [Mo0.426μs vs Mo0.676μs] -37.02% (±1.05%)
    benchFromStringWithNegativeIntegerAmoun.R1 I14 - [Mo0.423μs vs Mo0.731μs] -42.22% (±0.91%)
    benchFromStringWithZeroAndFractionalAmo.R2 I8 - [Mo0.476μs vs Mo0.689μs] -30.97% (±1.00%)
    benchFromStringWithFractionalAmount.....R2 I14 - [Mo0.609μs vs Mo1.257μs] -51.53% (±0.79%)
    benchFromStringWithNegativeFractionalAm.R1 I13 - [Mo0.612μs vs Mo1.304μs] -53.11% (±1.06%)

\Benchmark\Money\MoneyOperationBench

    benchAdd................................R1 I14 - [Mo1.244μs vs Mo1.338μs] -7.07% (±1.45%)
    benchSubtract...........................R1 I14 - [Mo1.254μs vs Mo1.315μs] -4.65% (±1.36%)
    benchMultiply...........................R3 I10 - [Mo1.297μs vs Mo2.346μs] -44.70% (±0.96%)
    benchDivide.............................R3 I14 - [Mo2.105μs vs Mo3.074μs] -31.52% (±1.03%)
    benchSum................................R1 I2 - [Mo2.231μs vs Mo3.630μs] -38.56% (±0.89%)
    benchMin................................R1 I14 - [Mo1.195μs vs Mo1.317μs] -9.22% (±1.13%)
    benchMax................................R1 I2 - [Mo1.191μs vs Mo1.307μs] -8.83% (±0.59%)
    benchAvg................................R2 I12 - [Mo1.190μs vs Mo1.309μs] -9.12% (±1.10%)
    benchRatioOf............................R2 I14 - [Mo0.783μs vs Mo1.232μs] -36.48% (±0.69%)
    benchMod................................R1 I4 - [Mo0.761μs vs Mo0.781μs] -2.60% (±1.24%)
    benchIsSameCurrency.....................R1 I14 - [Mo0.081μs vs Mo0.107μs] -24.24% (±0.96%)
    benchIsZero.............................R1 I1 - [Mo0.242μs vs Mo0.226μs] +7.02% (±0.96%)
    benchAbsolute...........................R1 I14 - [Mo0.345μs vs Mo0.307μs] +12.60% (±0.99%)
    benchNegative...........................R1 I14 - [Mo1.396μs vs Mo1.595μs] -12.48% (±1.01%)
    benchIsPositive.........................R1 I8 - [Mo0.236μs vs Mo0.219μs] +7.94% (±0.80%)
    benchCompare............................R1 I14 - [Mo0.320μs vs Mo0.361μs] -11.35% (±0.84%)
    benchLessThan...........................R1 I0 - [Mo0.351μs vs Mo0.394μs] -10.89% (±0.46%)
    benchLessThanOrEqual....................R1 I14 - [Mo0.354μs vs Mo0.386μs] -8.06% (±0.73%)
    benchEquals.............................R1 I14 - [Mo0.391μs vs Mo0.141μs] +177.89% (±0.96%)
    benchGreaterThan........................R1 I2 - [Mo0.352μs vs Mo0.386μs] -8.93% (±0.68%)
    benchGreaterThanOrEqual.................R1 I8 - [Mo0.352μs vs Mo0.387μs] -9.02% (±0.63%)

\Benchmark\Money\MoneyInstantiationBench

    benchConstructorWithZeroIntegerAmount...R2 I14 - [Mo0.149μs vs Mo0.175μs] -15.11% (±0.73%)
    benchConstructorWithPositiveIntegerAmou.R1 I0 - [Mo0.174μs vs Mo0.245μs] -29.24% (±1.32%)
    benchConstructorWithNegativeIntegerAmou.R1 I14 - [Mo0.175μs vs Mo0.248μs] -29.54% (±0.77%)
    benchConstructorWithZeroStringAmount....R1 I4 - [Mo0.183μs vs Mo0.162μs] +13.28% (±0.59%)
    benchConstructorWithPositiveStringAmoun.R1 I3 - [Mo0.205μs vs Mo0.180μs] +13.46% (±0.54%)
    benchConstructorWithNegativeStringAmoun.R1 I8 - [Mo0.209μs vs Mo0.185μs] +12.83% (±0.74%)

Subjects: 39, Assertions: 0, Failures: 0, Errors: 0
```

## CI check status:

 * [x] ~~Composer Normalize (pull_request) - Failing after 19s - somehow didn't check my updated `composer.json`?~~ :thinking: 
 * [ ] PrettyCI — Code formatting - no longer relevant - can be disabled?
 * [ ] Build (5.6) Expected - no longer expected - need adjustment of branch protection rules pre-merge
 * [ ] Build (7.0) Expected - no longer expected - need adjustment of branch protection rules pre-merge
 * [ ] Build (7.1) Expected - no longer expected - need adjustment of branch protection rules pre-merge
 * [ ] Build (7.2) Expected - no longer expected - need adjustment of branch protection rules pre-merge
 * [ ] Build (7.3) Expected - no longer expected - need adjustment of branch protection rules pre-merge
 * [ ] Build (7.4) Expected - no longer expected - need adjustment of branch protection rules pre-merge
 * [ ] PHP-CS-Fixer Expected - no longer expected - need adjustment of branch protection rules pre-merge
 * [ ] PHPStan Expected - no longer expected - need adjustment of branch protection rules pre-merge

## Fixes
 
 * [x] fixes #633 (PHP 8 compat is now declared with a stricter range)
 * [x] fixes #619 (PHP 8 compat)
 * [x] fixes #615 (no longer allowing `null` amount)
 * [x] fixes #583 (interface of taggable cache in `1.0.0`)
 * [x] fixes #575 (`"5000\n"` is **NOT** a `numeric-string`, so that value is now considered illegal at type level)
 * [x] fixes #474 (interface is now much stricter: json de-serialization is completely up to the end-user)
 * [x] fixes #465 (`Calculator` and `Number` are now internal)
 * [x] fixes #258 (negative ratio in `Money#allocate()` is unsupported)
 * [x] fixes #7 (internals of `Money` now use `numeric-string` and `BcMathCalculator` by default)


.github/workflows/ci.yml
```diff
@@ -13,9 +13,9 @@ jobs:
 
     steps:
       - name: Set up PHP
-        uses: shivammathur/setup-php@1.7.0
+        uses: shivammathur/setup-php@v2
         with:
-          php-version: '5.6'
```
Ocramius, line 18: Begone, relic of the past!

composer.json
```diff
@@ -24,25 +24,26 @@
...
         "moneyphp/iso-currencies": "^3.2.1",
-        "php-http/message": "^1.4",
-        "php-http/mock-client": "^1.0.0",
-        "phpspec/phpspec": "^3.4.3",
-        "phpunit/phpunit": "^5.7.27 || ^6.5.14 || ^7.5.18 || ^8.5",
-        "psr/cache": "^1.0",
-        "symfony/phpunit-bridge": "^4"
+        "php-http/message": "^1.11.0",
+        "php-http/mock-client": "^1.4.1",
+        "phpbench/phpbench": "1.0.0-beta1@BETA",
```
Ocramius, line 42: Yes, we need to use a beta version of this :-)

phpunit.xml.dist
```diff
@@ -1,22 +1,24 @@
...
-    <filter>
-        <whitelist>
+    <coverage>
+        <include>
             <directory suffix=".php">src</directory>
-        </whitelist>
-    </filter>
-
-    <listeners>
-        <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
```
Ocramius, line 16: Begone, over-complicated and messy dependency!

spec/CurrencySpec.php
```diff
@@ -22,13 +22,6 @@ function it_is_json_serializable()
         $this->shouldImplement(\JsonSerializable::class);
     }
 
-    function it_throws_an_exception_when_code_is_not_string()
-    {
-        $this->beConstructedWith(123);
-
-        $this->shouldThrow(\InvalidArgumentException::class)->duringInstantiation();
```
Ocramius, line 29: Argument is now a `string` as per type declaration, so this can be removed

.php_cs.dist
```diff
@@ -6,6 +6,12 @@ return PhpCsFixer\Config::create()
         'array_syntax' => ['syntax' => 'short'],
         'yoda_style' => false,
         'self_accessor' => false,
+        'global_namespace_import' => [
+            'import_classes' => true,
+            'import_constants' => true,
+            'import_functions' => true,
+        ],
+        'native_function_invocation' => true,
```
Ocramius, line 14: TODO: enforce `declare(strict_types=1);` and see if opcache cares about it (if at all)

src/Calculator/BcMathCalculator.php
```diff
@@ -12,14 +15,11 @@
 final class BcMathCalculator implements Calculator
 {
     /**
-     * @var string
+     * @var int
```
Ocramius, line 18: We kinda need to run psalm over the entire codebase: it's full of these quirks.

src/Money.php
```diff
@@ -62,35 +84,28 @@ final class Money implements \JsonSerializable
...
      * @param int|string $amount Amount, expressed in the smallest units of $currency (eg cents)
      *
-     * @throws \InvalidArgumentException If amount is not integer
+     * @throws InvalidArgumentException If amount is not integer
      */
     public function __construct($amount, Currency $currency)
     {
+        $this->currency = $currency;
+
+        if (is_int($amount)) {
```
Ocramius, line 93: Unsure about this one: the added function call seems to bring little, and perhaps even make things worse than `filter_var()` :-\

src/Money.php
```diff
@@ -122,7 +139,9 @@ private function assertSameCurrency(Money $other)
      */
     public function equals(Money $other)
     {
-        return $this->isSameCurrency($other) && $this->amount === $other->amount;
+        return $this->amount === $other->amount
+            // Note: non-strict equality is intentional here, since `Currency` is `final` and reliable.
```
Ocramius, line 143: Yes, we're YOLO-ing this :-P

src/Money.php
```diff
@@ -210,7 +229,7 @@ public function getCurrency()
     public function add(Money ...$addends)
     {
         $amount = $this->amount;
-        $calculator = $this->getCalculator();
+        $calculator = (self::$calculator ?? self::initializeCalculator());
```
Ocramius, line 232: Ugly but indeed more removed indirection for something that is a local cache

src/Money.php
```diff
@@ -245,40 +264,6 @@ public function subtract(Money ...$subtrahends)
...
     }
 
-    /**
-     * Asserts that the operand is integer or float.
-     *
-     * @param float|int|string $operand
-     *
-     * @throws \InvalidArgumentException If $operand is neither integer nor float
-     */
-    private function assertOperand($operand)
```
Ocramius, line 255: This can be enforced at type-level via psalm, and does not need a runtime check

src/Money.php
```diff
@@ -245,40 +264,6 @@ public function subtract(Money ...$subtrahends)
...
-    }
-
-    /**
-     * Asserts that rounding mode is a valid integer value.
-     *
-     * @param int $roundingMode
-     *
-     * @throws \InvalidArgumentException If $roundingMode is not valid
-     */
-    private function assertRoundingMode($roundingMode)
```
Ocramius, line 269: This can be enforced at type-level via psalm, and does not need a runtime check

src/Number.php
```diff
@@ -45,14 +56,11 @@ public function __construct($integerPart, $fractionalPart = '')
      */
     public static function fromString($number)
     {
-        $decimalSeparatorPosition = strpos($number, '.');
-        if ($decimalSeparatorPosition === false) {
-            return new self($number, '');
-        }
+        $portions = explode('.', $number, 2);
```
Ocramius, line 59: Didn't make a big difference, but still a bit easier to read.

src/Number.php
```diff
@@ -247,17 +255,23 @@ private static function parseIntegerPart($number)
             return '-0';
         }
 
+        // Happy path performance optimization: number can be used as-is if it is within
+        // the platform's integer capabilities.
+        if ($number === (string) (int) $number) {
```
Ocramius, line 260: I'm very puzzled about this: it seems like there's no visible performance impact, although it **should** be faster, since the algorithm below is `O(n)`, but in userland :thinking: 

My thought was to keep things as "simple integers" where possible in internals, but that's probably going to complicate the code by a lot.

src/Parser/DecimalMoneyParser.php
```diff
@@ -47,7 +56,7 @@ public function parse($money, $forceCurrency = null)
          */
         $currency = $forceCurrency;
         if (!$currency instanceof Currency) {
-            @trigger_error('Passing a currency as string is deprecated since 3.1 and will be removed in 4.0. Please pass a '.Currency::class.' instance instead.', E_USER_DEPRECATED);
```
Ocramius, line 50: Begone, hack that makes everything so much messier @_@ 

Heck, we may honestly just get rid of it overall, if we decide to break BC

frederikbosch:
@sagikazarmark In my opinion these changes should target MoneyPHP 4.0 with PHP8.0+. This would mean that @Ocramius has the freedom to remove deprecation warnings and add types where possible. So then we have Money 3 for <= PHP7.4. I would then also create a separate branch for the development of 4.0. Would you agree?

frederikbosch:
@Ocramius That's why I asked the other maintainer @sagikazarmark to join the discussion, and see what he thinks. I would suggest to break dependencies. I know there are places in the library that still use floats for calculations; they should be removed in this cycle too. We could indeed branch off to a 3.x branch.

Moreover I want to discuss how we should store numbers internally. There is this `Number` object I created which is instantiated too many times I believe, right? (forgive me, it's been a long time I really worked on this lib). We might want to change this.

Ocramius:
Working on this again today - will consider the changes as if I were on v4, so there will be BC breaks.

 > There is this `Number` object I created which is instantiated too many times I believe, right? (forgive me, it's been a long time I really worked on this lib). We might want to change this.

Yes, overall something I'd like to change: use `Number` only when user input is involved, while operating with `string` internally at all times, when possible.

frederikbosch:
Wonderful. My suggestion is to go for PHP8.0 and above, but I hope @sagikazarmark has the opportunity to confirm my suggestion.

Ocramius:
> My suggestion is to go for PHP8.0 and above

Works for me, and also makes my life easier with the whole type improvements. I don't think I'll wait for @sagikazarmark's input: that's mostly because I can (like everyone else) only dedicate limited time to this anyway, and I since v3 already supports PHP 7, there is no need to support it in v4.

frederikbosch:
Alright, go for it then. You are the one willing to spend time on this huge improve, whereas @sagikazarmark and I are the ones that do not have this time, so I guess we are happy with the changes you provide!

sagikazarmark:
Sorry for being late to the party. Yeah, I think targeting 4.0 with these changes makes sense. I'm not sure about PHP 8.0, unless there is a specific reason (ie. a feature that we want to use, a dependency that only supports 8.0, etc).

I can create a 3.x branch, so we can manage required checks separately. I wonder if submitting these changes in smaller chunks would be possible (eg. PHP upgrades and build changes then every other change). It would make reviews easier and would make sure that subsequent PRs run the new build pipelines.

We should also discuss the fate of the current 4.0 milestone: https://github.com/moneyphp/money/milestone/4
Is there something from there that needs to be done or can we do those once 4.0 is tagged?

Thanks for putting the effort into this @Ocramius!

Ocramius:
@sagikazarmark I can take care of some of them - will reference them in related commits.

It's gonna be an ugly patch that touches a lot of things, but it also comes down to optimizing available time :-\

It can also live as a `4.x` branch, waiting for people downstream to try it out until we feel like a release is safe to do.

frederikbosch:
@sagikazarmark How much time do you have to spend on this? I do have the time, so I can help Marco to get 4.0 out of the door. After he finishes his work, I will see what other tasks from [the 4.x milestone](https://github.com/moneyphp/money/milestone/4) I can merge it or resolve. 

The reason I suggest PHP8.0 is because A) it would bring us close to current standards reducing future maintenance costs B) we already support PHP7 in version 3.

Ocramius:
@frederikbosch I'll already move to PHP 8 because we need those native union types to remove unnecessary type casts.

frederikbosch:
The 3.x branch is there.

sagikazarmark:
Unfortunately I'm not a time billionaire these days (which is why there are so many outstanding PRs in this project :\), but your and Marco's track record is enough for me: happy to help wherever I can, feel free to include me in big decisions if you need input from me, otherwise feel free to proceed with whatever changes you see fit.

frederikbosch:
@sagikazarmark Wonderful, I will include you in big decisions and when I have doubts on particular issues. Hopefully, with version 4 we can reduce the number of outstanding PRs and issues.

sagikazarmark:
Thank you both.

Ocramius:
Meanwhile, some preliminary results of "the butchering":

```
❯ php -dopcache.enable_cli=1 ./vendor/bin/phpbench run --ref=ref_original_4 --retry-threshold=3 --iterations=15 --revs=1000  --warmup=2 > report.txt
PHPBench @git_tag@ running benchmarks...
with configuration file: /home/ocramius/Documents/moneyphp/money/phpbench.json
with PHP version 8.0.3, xdebug ❌, opcache ✔

\Benchmark\Money\NumberInstantiationBench

    benchConstructorWithZeroIntegerAmount...R1 I5 - [Mo0.212μs vs Mo0.168μs] +25.75% (±1.23%)
    benchConstructorWithPositiveIntegerAmou.R5 I12 - [Mo0.311μs vs Mo0.626μs] -50.28% (±1.26%)
    benchConstructorWithNegativeIntegerAmou.R1 I0 - [Mo0.306μs vs Mo0.690μs] -55.68% (±0.92%)
    benchConstructorWithZeroAndFractionalAm.R1 I1 - [Mo0.319μs vs Mo0.576μs] -44.62% (±1.11%)
    benchConstructorWithFractionalAmount....R2 I9 - [Mo0.435μs vs Mo1.148μs] -62.09% (±0.72%)
    benchConstructorWithNegativeFractionalA.R1 I14 - [Mo0.437μs vs Mo1.211μs] -63.96% (±0.63%)
    benchFromStringWithZeroIntegerAmount....R1 I14 - [Mo0.331μs vs Mo0.209μs] +58.24% (±1.65%)
    benchFromStringWithPositiveIntegerAmoun.R1 I5 - [Mo0.427μs vs Mo0.680μs] -37.26% (±1.21%)
    benchFromStringWithNegativeIntegerAmoun.R1 I13 - [Mo0.425μs vs Mo0.735μs] -42.10% (±0.77%)
    benchFromStringWithZeroAndFractionalAmo.R2 I7 - [Mo0.478μs vs Mo0.697μs] -31.41% (±1.09%)
    benchFromStringWithFractionalAmount.....R1 I1 - [Mo0.608μs vs Mo1.257μs] -51.65% (±0.94%)
    benchFromStringWithNegativeFractionalAm.R1 I13 - [Mo0.611μs vs Mo1.316μs] -53.58% (±1.06%)

\Benchmark\Money\MoneyOperationBench

    benchAdd................................R2 I4 - [Mo1.187μs vs Mo1.338μs] -11.32% (±1.85%)
    benchSubtract...........................R6 I11 - [Mo1.155μs vs Mo1.304μs] -11.39% (±1.19%)
    benchMultiply...........................R1 I5 - [Mo1.136μs vs Mo2.331μs] -51.28% (±1.32%)
    benchDivide.............................R1 I2 - [Mo1.868μs vs Mo3.098μs] -39.69% (±1.21%)
    benchSum................................R1 I11 - [Mo1.930μs vs Mo3.584μs] -46.15% (±1.50%)
    benchMin................................R1 I14 - [Mo0.984μs vs Mo1.303μs] -24.54% (±0.90%)
    benchMax................................R2 I13 - [Mo0.981μs vs Mo1.306μs] -24.88% (±0.99%)
    benchAvg................................R3 I9 - [Mo0.988μs vs Mo1.309μs] -24.56% (±1.44%)
    benchRatioOf............................R2 I11 - [Mo0.648μs vs Mo1.226μs] -47.10% (±0.86%)
    benchMod................................R1 I12 - [Mo0.707μs vs Mo0.778μs] -9.23% (±1.26%)
    benchIsSameCurrency.....................R1 I1 - [Mo0.075μs vs Mo0.104μs] -27.92% (±1.17%)
    benchIsZero.............................R2 I13 - [Mo0.172μs vs Mo0.226μs] -23.74% (±0.79%)
    benchAbsolute...........................R1 I0 - [Mo0.283μs vs Mo0.306μs] -7.43% (±0.70%)
    benchNegative...........................R1 I2 - [Mo1.347μs vs Mo1.620μs] -16.86% (±1.39%)
    benchIsPositive.........................R11 I14 - [Mo0.167μs vs Mo0.218μs] -23.33% (±1.28%)
    benchCompare............................R1 I1 - [Mo0.253μs vs Mo0.363μs] -30.47% (±0.64%)
    benchLessThan...........................R1 I14 - [Mo0.281μs vs Mo0.385μs] -27.15% (±0.74%)
    benchLessThanOrEqual....................R2 I5 - [Mo0.278μs vs Mo0.386μs] -27.89% (±1.29%)
    benchEquals.............................R1 I2 - [Mo0.320μs vs Mo0.141μs] +127.52% (±0.85%)
    benchGreaterThan........................R1 I14 - [Mo0.278μs vs Mo0.387μs] -28.17% (±0.82%)
    benchGreaterThanOrEqual.................R4 I12 - [Mo0.280μs vs Mo0.386μs] -27.42% (±1.03%)

\Benchmark\Money\MoneyInstantiationBench

    benchConstructorWithZeroIntegerAmount...R1 I3 - [Mo0.156μs vs Mo0.181μs] -13.62% (±1.16%)
    benchConstructorWithPositiveIntegerAmou.R7 I12 - [Mo0.175μs vs Mo0.248μs] -29.61% (±1.67%)
    benchConstructorWithNegativeIntegerAmou.R1 I4 - [Mo0.180μs vs Mo0.252μs] -28.76% (±1.45%)
    benchConstructorWithZeroStringAmount....R2 I5 - [Mo0.184μs vs Mo0.168μs] +9.44% (±1.43%)
    benchConstructorWithPositiveStringAmoun.R2 I14 - [Mo0.205μs vs Mo0.189μs] +8.55% (±1.35%)
    benchConstructorWithNegativeStringAmoun.R7 I12 - [Mo0.212μs vs Mo0.190μs] +11.52% (±1.31%)

Subjects: 39, Assertions: 0, Failures: 0, Errors: 0
```

My next step, given my deep dive into the lib, is to make `ext-bcmath` a requirement, and completely get rid of the entire "calculator" concept.

Given that installing extensions nowadays is pretty much trivial (post-Docker world), I think that's a relatively reasonable approach.

Besides that, the `GmpCalculator` and the `PhpCalculator` are full of complexity traps that we should really get rid of :|

src/Money.php
```diff
@@ -204,18 +185,18 @@ public function getCurrency()
...
-    public function add(Money ...$addends)
+    public function add(Money ...$addends): Money
     {
         $amount = $this->amount;
-        $calculator = $this->getCalculator();
 
         foreach ($addends as $addend) {
-            $this->assertSameCurrency($addend);
+            // Note: non-strict equality is intentional here, since `Currency` is `final` and reliable.
+            if ($this->currency != $addend->currency) {
```
Ocramius, line 195: assertions have been inlined - kinda needed for these variadic `O(n)` operations.

src/Money.php
```diff
@@ -227,158 +208,122 @@ public function add(Money ...$addends)
...
-        if ($this->getCalculator()->compare($remainder, '0') === 0) {
+        if (self::$calculator::compare($remainder, '0') === 0) {
             return $results;
         }
 
-        $fractions = array_map(function ($ratio) use ($total) {
-            $share = ($ratio / $total) * $this->amount;
+        $amount    = $this->amount;
+        $fractions = array_map(static function (float|int $ratio) use ($total, $amount) {
+            $share = (float) $ratio / $total * (float) $amount;
```
Ocramius, line 318: I did not really understand all of this logic, but left it as-is.

Seems risky due to divisions/multiplications happening directly here, rather than in the calculators.

Trying to replace it with calculators leads to failed tests, hence why I held back in trying to adjust it.

src/Money.php
```diff
@@ -524,89 +452,38 @@ public static function min(self $first, self ...$collection)
...
-        return $first->add(...$collection)->divide(func_num_args());
+        return $first->add(...$collection)->divide((string) (count($collection) + 1));
     }
 
-    /**
-     * @param string $calculator
-     */
-    public static function registerCalculator($calculator)
+    /** @psalm-param class-string<Calculator> $calculator */
+    public static function registerCalculator(string $calculator): void
```
Ocramius, line 485: Tegistering the calculator now pretty much means using it.

This is by design: preferred to keep the old name to ease transition for those that use custom calculators. Assuming they register a calculator in order to use it, they probably want it to be in use anyway.

src/MoneyFactory.php
```diff
@@ -184,6 +188,7 @@
  * @method static Money ZAR(string|int $amount)
```
Ocramius, line 188: I think these should become `numeric-string` too - still to be adjusted.

src/Number.php
```diff
@@ -1,135 +1,105 @@
...
+use function sprintf;
+use function str_pad;
+use function strlen;
+use function substr;
+
 /**
  * Represents a numeric value.
  *
- * @author Frederik Bosch <f.bosch@genkgo.nl>
+ * @internal this is an internal utility of the library, and may vary at any time. It is mostly used to internally validate
```
Ocramius, line 24: This class is now `@internal` and no longer part of the BC promise of this library.

That is intentional: the amount of voodoo and magic in here is NOT sustainable long-term, and we should put our efforts in trying to get rid of it in future, if possible.

src/Number.php
Ocramius: Verified that there actually **is** a performance impact: comment above can be dismissed.

src/PHPUnit/Comparator.php
```diff
@@ -1,27 +1,33 @@
...
+use function assert;
+
 /**
  * The comparator is for comparing Money objects in PHPUnit tests.
  *
  * Add this to your bootstrap file:
  *
  * \SebastianBergmann\Comparator\Factory::getInstance()->register(new \Money\PHPUnit\Comparator());
+ *
+ * @internal do not use within your sources: this comparator is only to be used within the test suite of this library
```
Ocramius, line 24: Made this `@internal` - it's not supposed to be used by third-parties, and is only really used within our test suite.

It's also not documented, AFAIK

src/Parser/AggregateMoneyParser.php
```diff
@@ -1,51 +1,38 @@
...
+        $this->parsers = $parsers;
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function parse($money, $forceCurrency = null)
+    public function parse(string $money, Currency|null $forceCurrency = null): Money
     {
-        if ($forceCurrency !== null && !$forceCurrency instanceof Currency) {
```
Ocramius, line 44: These notices have been removed from all over the library: much simpler to upgrade the type and let the type-checker tell that there's an incompatibility.

tests/Currencies/CurrencyListTest.php
```diff
@@ -52,45 +61,22 @@ public function it_throws_an_exception_when_providing_subunit_and_currency_is_un
...
         $iterator = $currencies->getIterator();
 
         $this->assertContainsOnlyInstancesOf(Currency::class, $iterator);
     }
 
-    /**
-     * @dataProvider invalidInstantiation
-     * @test
-     */
-    public function it_does_not_initialize_if_array_is_invalid(array $currencies)
```
Ocramius, line 68: No longer valid, checked at type level

tests/Formatter/AggregateMoneyFormatterTest.php
```diff
@@ -1,20 +1,79 @@
...
+use Money\Money;
+use Money\MoneyFormatter;
 use PHPUnit\Framework\TestCase;
 
 final class AggregateMoneyFormatterTest extends TestCase
 {
-    /**
-     * @test
-     */
-    public function it_can_be_initialized_with_empty_array()
```
Ocramius, line 13: No longer valid: checked at type level

tests/Formatter/BitcoinMoneyFormatterTest.php
```diff
@@ -1,36 +1,52 @@
...
+     * @psalm-param positive-int|0 $fractionDigits
+     *
      * @dataProvider bitcoinExamples
      * @test
      */
-    public function it_formats_money($value, $formatted, $fractionDigits)
+    public function itFormatsMoney(int $value, string $formatted, int $fractionDigits): void
     {
-        /** @var Currencies|ObjectProphecy $currencies */
-        $currencies = $this->prophesize(Currencies::class);
```
Ocramius, line 21: I removed prophecy where possible: it was mostly noise

tests/MoneyTest.php
```diff
@@ -53,77 +77,67 @@ public function it_compares_two_amounts($other, $result)
...
-
-        $this->assertInstanceOf(Money::class, $money);
-        $this->assertEquals(10, $money->getAmount());
-    }
-
-    /**
-     * @dataProvider invalidOperandExamples
-     * @test
-     */
-    public function it_throws_an_exception_when_operand_is_invalid_during_multiplication($operand)
```
Ocramius, line 87: Operand now checked at type level

tests/Parser/AggregateMoneyParserTest.php
```diff
@@ -1,20 +1,82 @@
...
+use Money\MoneyParser;
 use Money\Parser\AggregateMoneyParser;
 use PHPUnit\Framework\TestCase;
 
 final class AggregateMoneyParserTest extends TestCase
 {
-    /**
-     * @test
-     */
-    public function it_can_be_initialized_with_empty_array()
```
Ocramius, line 13: Now checked at type level

tests/Parser/DecimalMoneyParserTest.php
```diff
@@ -45,29 +56,18 @@ public function it_throws_an_exception_upon_invalid_inputs($input)
...
-     * @expectedDeprecation Passing a currency as string is deprecated since 3.1 and will be removed in 4.0. Please pass a Money\Currency instance instead.
-     * @test
+     * @psalm-return non-empty-list<array{
+     *     string,
+     *     non-empty-string,
+     *     positive-int|0,
+     *     int
+     * }>
      */
-    public function it_accepts_only_a_currency_object()
```
Ocramius, line 56: Now checked at type level

tests/Parser/IntlLocalizedDecimalParserTest.php
```diff
@@ -80,41 +89,39 @@ public function it_accepts_a_forced_currency()
...
 
         $this->assertSame('100001', $money->getAmount());
     }
 
     /**
-     * @group legacy
-     * @expectedDeprecation Passing a currency as string is deprecated since 3.1 and will be removed in 4.0. Please pass a Money\Currency instance instead.
      * @test
      */
-    public function it_accepts_only_a_currency_object()
```
Ocramius, line 110: Now checked at type level

tests/Parser/IntlMoneyParserTest.php
```diff
@@ -105,33 +112,25 @@ public function it_supports_fraction_digits()
...
     /**
-     * @group legacy
-     * @expectedDeprecation Passing a currency as string is deprecated since 3.1 and will be removed in 4.0. Please pass a Money\Currency instance instead.
-     * @test
+     * @psalm-return non-empty-list<array{
+     *     non-empty-string,
+     *     int
+     * }>
      */
-    public function it_accepts_only_a_currency_object()
```
Ocramius, line 125: Now checked at type level

src/MoneyFactory.php
Ocramius: 0689d3cc0933093144967fa4864aa5092fa29bf6

Ocramius:
@frederikbosch @sagikazarmark now ready for review :-)

psalm.xml
```diff
@@ -1,51 +1,82 @@
 <?xml version="1.0"?>
 <psalm
+    findUnusedPsalmSuppress="true"
+    findUnusedCode="true"
+    findUnusedVariablesAndParams="true"
+    strictBinaryOperands="true"
     totallyTyped="true"
```
bendavies, line 7: https://psalm.dev/docs/running_psalm/configuration/#totallytyped
this is deprecated and should prefer `errorLevel="1"`

resources/generate-money-factory.php
```diff
@@ -3,59 +3,65 @@
...
-    new Currencies\ISOCurrencies(),
-    new Currencies\BitcoinCurrencies(),
-]);
+    $currencies = iterator_to_array(new Currencies\AggregateCurrencies([
+        new Currencies\ISOCurrencies(),
+        new Currencies\BitcoinCurrencies(),
+    ]));
 
-$currencies = iterator_to_array($currencies);
+    usort($currencies, static fn (Currency $a, Currency $b): int => strcmp($a->getCode(), $b->getCode()));
```
Neirda24, line 57: Couldn't this be done as a method Inthe iterator object ? Allow more readability and avoid altogether the need for iterator to array ?

psalm.xml
Ocramius: Until `totallyTyped` is supported, I will stick to that: I find `errorLeve="1"` to be super-confusing and counter-intuitive, as you need to go to the docs to find out what it implies.

Until then, sticking with the deprecated (better) syntax :D

resources/generate-money-factory.php
Ocramius: This was only re-aligned, not changed: will keep it as-is, since it's not a runtime component of the lib, and it has no actual tests besides running it manually.

src/Calculator.php
```diff
@@ -1,127 +1,148 @@
...
-     * @param int|float|string $divisor
+     * @psalm-param numeric-string $amount
+     * @psalm-param numeric-string $divisor
+     *
+     * @psalm-return numeric-string
+     *
+     * @psalm-pure
      *
-     * @return string
+     * @TODO what to do with $divisor == '0'?
```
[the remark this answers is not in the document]
frederikbosch: I'd say it should throw.

src/Calculator/BcMathCalculator.php
```diff
@@ -1,241 +1,212 @@
...
 
-    /**
-     * {@inheritdoc}
-     */
-    public function add($amount, $addend)
+    /** @psalm-pure */
+    public static function multiply(string $amount, string $multiplier): string
     {
-        return (string) Number::fromString(bcadd($amount, $addend, $this->scale));
+        return bcmul($amount, $multiplier, self::SCALE);
```
frederikbosch, line 45: I believe `add` and `substract` could return additional zeros in the fractional part, depending on the scale, and therefore the [Number object was used to get rid of them](https://github.com/moneyphp/money/blob/master/src/Number.php#L55). 

All operations should return a value without additional zeros, otherwise we would have to improve our equality comparison to allow for these zeros. My guess is that performance, apparently of high importance, would suffer more from additional operations during equality comparison than during calculator operations.

src/Calculator/PhpCalculator.php
```diff
@@ -1,152 +1,149 @@
...
+use function floor;
+use function ltrim;
+use function round;
+
+use const FILTER_VALIDATE_INT;
+use const PHP_INT_MAX;
 
 /**
- * @author Frederik Bosch <f.bosch@genkgo.nl>
+ * @internal  this calculator is **NOT** supposed to be used in a production environment. You
```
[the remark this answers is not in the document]
frederikbosch: Agreed. Let's drop it.

src/Converter.php
```diff
@@ -1,45 +1,34 @@
...
+        $ratio        = $this->exchange->quote($baseCurrency, $counterCurrency)->getConversionRatio();
 
-        $baseCurrencySubunit = $this->currencies->subunitFor($baseCurrency);
+        $baseCurrencySubunit    = $this->currencies->subunitFor($baseCurrency);
         $counterCurrencySubunit = $this->currencies->subunitFor($counterCurrency);
-        $subunitDifference = $baseCurrencySubunit - $counterCurrencySubunit;
+        $subunitDifference      = $baseCurrencySubunit - $counterCurrencySubunit;
 
-        $ratio = (string) Number::fromFloat($ratio)->base10($subunitDifference);
+        $ratio = (string) Number::fromFloat((float) $ratio)->base10($subunitDifference);
```
frederikbosch, line 31: While we are at it, you already converted the return type for `CurrencyPair::getConversionRatio()` from [float](https://github.com/moneyphp/money/blob/a44e2f2bb6bf048596af1649cb7c51dd421a979a/src/CurrencyPair.php#L96) to [string](https://github.com/Ocramius/money/blob/feature/quick-money/src/CurrencyPair.php#L92), so let's also change `Number::fromFloat()` to `Number::fromString()` and get rid of the floats.

src/Currencies/AggregateCurrencies.php
```diff
@@ -46,31 +36,24 @@ public function contains(Currency $currency)
...
-    public function getIterator()
+    /** {@inheritDoc} */
+    public function getIterator(): Traversable
     {
-        $iterator = new \AppendIterator();
-
         foreach ($this->currencies as $currencies) {
-            $iterator->append($currencies->getIterator());
+            foreach ($currencies as $currency) {
+                yield $currency;
```
frederikbosch, line 55: Agree here, but isn't this a BC break? From rewindable to non-rewindable iterator?

src/Exchange/ExchangerExchange.php
```diff
@@ -13,35 +15,27 @@
...
             );
-            $rate = $this->exchanger->getExchangeRate($query);
-        } catch (ExchangerException $e) {
+            $rate  = $this->exchanger->getExchangeRate($query);
+        } catch (ExchangerException) {
             throw UnresolvableCurrencyPairException::createFromCurrencies($baseCurrency, $counterCurrency);
         }
 
-        return new CurrencyPair($baseCurrency, $counterCurrency, $rate->getValue());
+        return new CurrencyPair($baseCurrency, $counterCurrency, (string) $rate->getValue());
```
frederikbosch, line 39: What does `$rate->getValue()` return here? If it is, or might be, float to string, then I think we should consider using `Number::fromFloat` or use a similar `sprintf` call as `Number::fromFloat` is using.

src/Exchange/ReversedCurrenciesExchange.php
```diff
@@ -11,34 +13,30 @@
...
         } catch (UnresolvableCurrencyPairException $exception) {
             try {
                 $currencyPair = $this->exchange->quote($counterCurrency, $baseCurrency);
 
-                return new CurrencyPair($baseCurrency, $counterCurrency, 1 / $currencyPair->getConversionRatio());
-            } catch (UnresolvableCurrencyPairException $inversedException) {
+                return new CurrencyPair(
+                    $baseCurrency,
+                    $counterCurrency,
+                    (string) (1.0 / (float) $currencyPair->getConversionRatio())
```
frederikbosch, line 37: I'd love to remove this float here, but I guess this would require the use of a calculator. Unless you see the opportunity to fix this in this PR, this should be a task after merging.

src/Exchange/SwapExchange.php
```diff
@@ -11,32 +13,24 @@
...
         try {
-            $rate = $this->swap->latest($baseCurrency->getCode().'/'.$counterCurrency->getCode());
-        } catch (ExchangerException $e) {
+            $rate = $this->swap->latest($baseCurrency->getCode() . '/' . $counterCurrency->getCode());
+        } catch (ExchangerException) {
             throw UnresolvableCurrencyPairException::createFromCurrencies($baseCurrency, $counterCurrency);
         }
 
-        return new CurrencyPair($baseCurrency, $counterCurrency, $rate->getValue());
+        return new CurrencyPair($baseCurrency, $counterCurrency, (string) $rate->getValue());
```
frederikbosch, line 34: See earlier question.

frederikbosch (commented):
Review finished, I have some remarks. All in all extremely valuable additions to the library. I am very glad you picked it up and gave continuity, thank you very much!

src/Money.php
frederikbosch: Then let's drop it.

src/Money.php
```diff
@@ -1,200 +1,181 @@
...
-     * @return bool
-     */
-    public function equals(Money $other)
-    {
-        return $this->isSameCurrency($other) && $this->amount === $other->amount;
+        if ($this->amount === $other->amount) {
+            return true;
+        }
+
+        // @TODO do we want Money instance to be byte-equivalent when trailing zeroes exist? Very expensive!
```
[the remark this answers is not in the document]
frederikbosch: Checking for equality might be faster without use of a calculator, but I do not know how much performance priority this has.

src/Money.php
frederikbosch: Perfect.

frederikbosch (commented):
LOL, I was interupted during my review and therefore I forgot the most important class to review.

src/Calculator/BcMathCalculator.php
Ocramius: It is possible to re-introduce `Number` here, at the expense of some performance improvements that would still be noticeable.

src/Currencies/AggregateCurrencies.php
Ocramius: Did not think about this: I can revert the code overall, if preferred? Would need to add some type suppressions, but no biggie.

src/Exchange/ExchangerExchange.php
Ocramius: It returns `float` in this case: I don't think this is performance-sensitive though, since we're doing I/O against an external system anyway

src/Exchange/ReversedCurrenciesExchange.php
Ocramius: Initially wanted to use the calculator here, but avoided introducing changes in components that are not performance sensitive. Most changes you see have been done to sustain the improvements in tests and performance of the core `Money` VO.

I would say new issue and defer here.

src/Exchange/SwapExchange.php
Ocramius: Same as https://github.com/moneyphp/money/pull/634#discussion_r617635607

src/Currencies/AggregateCurrencies.php
frederikbosch: To be honest I am often struggling with this. From a writer and reader perspective generators result in code that is easier to write and to easier to read, but it is not infrequent that I hit a generator that is not rewindable. Let's reverse it here.

src/Exchange/ExchangerExchange.php
frederikbosch: My comment was not related to performance, but related to the inaccurate `(string)$float` conversion. I believe `sprintf('%.14F', $float)` is a better way to handle this, see [this discussion we had earlier](https://github.com/moneyphp/money/issues/408).

src/Exchange/ReversedCurrenciesExchange.php
frederikbosch: Makes sense, see #636

frederikbosch:
In my edited review response I also asked the following questions. Maybe you missed it because I was stilling editing and you were already replying.

Am I correct that I do not see a PHP-CS-Fixer Check (github action) anymore in the PR? Why would you leave it out?

Ocramius:
> Am I correct that I do not see a PHP-CS-Fixer Check (github action) anymore in the PR? Why would you leave it out?

It's done by PHPCS now, which is much more restrictive - I see I forgot it in the `.yml` definitions

src/Calculator/PhpCalculator.php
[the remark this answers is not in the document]
Ocramius: dc3ce62d468c1550664d8e3f32a12420d2bb01fd

src/Calculator.php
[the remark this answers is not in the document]
Ocramius: dc3ce62d468c1550664d8e3f32a12420d2bb01fd

src/Calculator/BcMathCalculator.php
```diff
@@ -1,241 +1,212 @@
...
 
-    /**
-     * {@inheritdoc}
-     */
-    public function mod($amount, $divisor)
+    /** @psalm-pure */
+    public static function mod(string $amount, string $divisor): string
     {
-        return bcmod($amount, $divisor);
+        // @TODO: null check needed because `bcmod(_, 0)` fails - should the check be kept in here?
```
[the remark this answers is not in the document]
Ocramius: dc3ce62d468c1550664d8e3f32a12420d2bb01fd

src/Money.php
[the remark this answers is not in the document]
Ocramius: Performance is not the big deal, it was more about object-equality, but then the library provides a PHPUnit comparator, should that be important :+1:

src/Calculator/BcMathCalculator.php
Ocramius: Checked this again: indeed, equality is not as important as general-purpose multiplication/division, and `round()` operations already take care of this.

Marking therefore as resolved.

src/Calculator/PhpCalculator.php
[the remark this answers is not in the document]
Ocramius: 7790e528989283470178ba644cb005e91eb3bfcf

src/Converter.php
Ocramius: eb32c53dbc32ad31a387735efcb2a7e6024f0a0c

src/Exchange/ExchangerExchange.php
Ocramius: 9496d0940a031249a42c03bfab0ac79826ed7a22

Ocramius:
@frederikbosch another round?

src/Number.php
```diff
@@ -58,18 +57,8 @@ public static function fromString(string $number): self
     }
 
     /** @psalm-pure */
-    public static function fromFloat(float $number): self
+    public static function fromNumber(int|string $number): self
     {
-        return self::fromString(sprintf('%.14F', $number));
```
frederikbosch, line 70: While I know you marked this class as internal, and that we are breaking BC, I simply know from issue history that  people are relying on this. I don't think I find removing `Number::fromFloat` completely necessary per se.

src/Exchange/ExchangerExchange.php
```diff
@@ -36,6 +40,10 @@ public function quote(Currency $baseCurrency, Currency $counterCurrency): Curren
             throw UnresolvableCurrencyPairException::createFromCurrencies($baseCurrency, $counterCurrency);
         }
 
-        return new CurrencyPair($baseCurrency, $counterCurrency, (string) $rate->getValue());
+        $rateValue = sprintf('%.14F', $rate->getValue());
+
+        assert(is_numeric($rateValue));
```
frederikbosch, line 45: Why is this `assert` necessary? I read on PHP manual page that from PHP 8 `sprintf` always returns string.

src/Currencies/AggregateCurrencies.php
frederikbosch: What do you think? I would prefer a rewindable iterator here.

Ocramius: Ah yes, I simply forgot to tackle this comment 👍

src/Number.php
Ocramius: I can restore it, and we can avoid using it in the code: not much to revert - will do, but please note that `Number` is now `@internal`, and people shouldn't rely on it anymore

src/Exchange/ExchangerExchange.php
frederikbosch: Sorry, still don't get it. Why would you want it to be type checked when the only possible result is string? Why the `is_numeric` hint?

src/Number.php
frederikbosch: Please do. I understand your remark, and I will take it into account when writing upgrade docs.

src/Exchange/ExchangerExchange.php
Ocramius: In order for this `string` to be acceptable by endpoints that use it as a number, it must be a `numeric-string`.

`string` won't be sufficient.

frederikbosch: Ok, clear.

src/Number.php
Ocramius: 0ca293324ac1033a1357bae564445f3dddac7cea

src/Currencies/AggregateCurrencies.php
frederikbosch: If you could revert this one, I will merge it in!

Ocramius: I did some investigation on this, and it looks kinda broken with `AppendIterator`: https://3v4l.org/sImTa

Is that the expected behavior? :thinking:

Ocramius: Nvm, fixed: 8e0630f367f21af20de3221f7068431f05b07c90

src/Number.php
Ocramius: 756649c07e2bb827e08bcc9b09b3f880d7b0fb73

Ocramius:
@frederikbosch should be OK for another round

frederikbosch:
How about merging in? It's so huge that I have to work with it in order to find improvements.

Ocramius:
> How about merging in? It's so huge that I have to work with it in order to find improvements.

Given the previous review, I suggest reviewing the individual (new) commits one-by-one, since I try to keep commit messages as relevant as possible.

Ocramius:
@frederikbosch just realized that yours may be a typo in the message: from my PoV, this is mergeable, and I have decent confidence in the amount of checks I've added here to avoid major regressions.

Minor regressions are unavoidable, but that's normality in OSS: https://xkcd.com/1172/

Ocramius:
I'm releasing this on a fork to internally test it on a customer project: IMO this is mergeable here too :+1:

frederikbosch:
Merged, @Ocramius thanks a lot for all your work here. It is much appreciated by us maintainers, but I guess also by many users. I will try my best to come up with a release - first alpha/beta - as soon as possible.

frederikbosch:
@Ocramius You added `roave/infection-static-analysis-plugin` to this package in this PR. Yesterday the MSI reached a point below the 78% that you have set. I have no idea how to approach this, other than lowering this number. Is there any advice you can give in this regard?

Ocramius:
>  Yesterday the MSI reached a point below the 78% that you have set.

How did it reach it? What was the change? Was there a dependency change?

frederikbosch:
I added `ext-filter` and `ext-json` after running maglnet/composer-require-checker.

Ocramius:
@frederikbosch do you have a diff somewhere? Perhaps `composer update` changed something else?

frederikbosch:
[Yes](https://github.com/moneyphp/money/pull/661/files).

Ocramius:
@frederikbosch that bosh does a general `composer update`, which also happened to update `infection/infection` from `0.20.2` to `0.23.0`.

Newer versions of `infection/infection` apply **MORE** mutators, and therefore need more effort in testing (especially around regular expressions, for which I think new mutators were introduced).

My suggestion is to:

 * revert #661
 * re-introduce the `composer.json` (only) changes from #661, run a `composer update nothing`, which does **NOT** alter `composer.lock`
 * configure @dependabot to upgrade individual dependencies one by one, instead of a one-shot `composer update`

If you need any help with it, I can probably hop on a call this afternoon CEST.

frederikbosch:
`composer update nothing`? Never heard of before, very nice! See #663. Thanks, for the comprehensive remarks. I will try to add dependabot later today.

bcremer:
```

**Comment by bcremer** [comment]
I'm optimizing an application that deals with lots of MoneyPHP instances and operations (50k+ instances).

The MoneyPHP instantiation and operations sum to a few hundred milliseconds for a web request so we needed to optimize that.

With the benchmarks provided by @Ocramius I discovered that the `add` and `subtract` orations are considerable faster with the `GmpCalculator`. Most other operations are faster with the default `BcMathCalculator`.

To get the most performance in our application I wrote a delegating calculator that uses both implementations.


```php
/**
 * Optimized `Money\Calculator` that delegates to either 
 * `BcMathCalculator` or `GmpCalculator` depending on the operation.
 *
 * `GmpCalculator` is approx 25% faster for add and subtract.
 * `BcMathCalculator` is faster for all decimal operations.
 */
class DelegatingCalculator implements Calculator
{
    public static function compare(string $a, string $b): int
    {
        return BcMathCalculator::compare($a, $b);
    }

    public static function add(string $amount, string $addend): string
    {
        return GmpCalculator::add($amount, $addend);
    }
    
[...]

Money::registerCalculator(DelegatingCalculator::class);
```

Sorry for responding in this old PR but maybe some one else finds this useful.

frederikbosch:
Do you have some benchmarks? Maybe create a PR with a `PerformanceOptimizedDelegateCalculator` class? Tests are easy since you can simply extend the [abstract CalculatorTestCase](https://github.com/moneyphp/money/blob/master/tests/Calculator/CalculatorTestCase.php).

PR #638, opened 2021-04-24
Fix #637: Ensure Currency::$code is uppercase

### LINKED ISSUE #637: Currency: ensure the code is uppercase (or lowercase, depending on the decision in this issue)
Currently the `Currency` object doesn't ensure consistency on `$code` property.

```php

$upperCurrency = new Currency('EUR');
$lowerCurrency = new Currency('eur');

echo $upperCurrency->getCode(); // Returns 'EUR'
echo $lowerCurrency->getCode(); // Returns 'eur'
```

This is not convenient as it doesn't ensure consistency and may cause some issues.

The first issue is on identity:

```php
var_dump($lowerCurrency->getCode() === $upperCurrency->getCode()); // bool(false), while it should be true
```

Directly descending from the first issue, there is a database issue.

I've created a Doctrine type that uses the `Currency` object to create a `currency` database field type.

Unfortunately, as the object doesn't ensure consistency, if I pass it `eur` (lowercase), it cannot find entites that have the currency set as uppercase (`EUR`).

For sure, I could ensure consistency on myself, but simply adding a function call to the `Currency` object, all will be more solid:

```diff
final class Currency implements \JsonSerializable
{
    /**
     * Currency code.
     *
     * @var string
     */
    private $code;

    /**
     * @param string $code
     */
    public function __construct($code)
    {
        if (!is_string($code)) {
            throw new \InvalidArgumentException('Currency code should be string');
        }

        if ($code === '') {
            throw new \InvalidArgumentException('Currency code should not be empty string');
        }

-        $this->code = $code;
+        $this->code = strtoupper($code); // or strtolower($code), depending on the preference of the lead devs of this package
    }
...
```

I proposed a PR using `strtoupper()` but, if you prefer, also `strtolower()` can be used.

Aerendir:
Fix #637: Ensure `$code` is uppercase.

I proposed a PR using `strtoupper()` but, if you prefer, also `strtolower()` can be used.

### DIFF
--- a/src/Currency.php
+++ b/src/Currency.php
@@ -33,7 +33,7 @@ final class Currency implements \JsonSerializable
             throw new \InvalidArgumentException('Currency code should not be empty string');
         }
 
-        $this->code = $code;
+        $this->code = strtoupper($code);
     }
 
     /**


frederikbosch:
Fixed by #639

PR #645, opened 2021-05-03
Remove mention of older PHP versions

PowerKiKi:
`composer.json` is the source of truth for PHP requirements, and should
not be duplicated to reduce maintenance burden.


frederikbosch:
Thanks, good suggestions and good findings in the tests.

PR #646, opened 2021-05-03
Only register calculator in money class

frederikbosch:
Supersedes #458.

This PR simplifies working with calculators. By calling `Money::registerCalculator()` you register a calculator for all calculations in this library.


frederikbosch:
@Ocramius I would be happy if you also give your feedback on this one too. It will be the last PR I will ask you to review. Then we are very close to releasing a beta version.

PR #648, opened 2021-05-17
Allow integer as multiplier for `Money::multiply` and divisor for `Money::divide`

ruudk:
Our application calculates a lot with `Money::multiply` and `Money::divide`. When trying to upgrade to [Money v4.0.0](https://github.com/moneyphp/money/releases/tag/v4.0.0) I needed to add `(string)` casts to almost all of the multiplier/divisors. It makes the code harder to read. 

I wonder if it's really needed, as `Money::__construct` also takes integers now. Can't we do the same for `Money::multiply` and `Money::divide`?

What do you think?


frederikbosch:
I think this is very helpful indeed. Maybe investigate whether more methods have this problem.

frederikbosch:
Great, thanks!

frederikbosch:
I will wait until the end of the week with tagging 4.0.1 (please remind me if I forgot ...)

frederikbosch:
Tagged, specifically to fix this issue, thanks again!

PR #651, opened 2021-05-19
Update wording about minimum PHP version

BackEndTea:
This should make it clearer that PHP 8 is the minimum.
Original wording looked more like it was now supported


frederikbosch:
Merci!

PR #653, opened 2021-05-27
Remove ext-json from required in composer.json

tuqqu:
Extension JSON is [part](https://wiki.php.net/rfc/always_enable_json) of PHP core since version 8, there is no need to specify it in the composer.json file any longer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,8 +25,7 @@
     ],
     "require": {
         "php": "^8.0",
-        "ext-bcmath": "*",
-        "ext-json": "*"
+        "ext-bcmath": "*"
     },
     "require-dev": {
         "ext-gmp": "*",


frederikbosch:
@tuqqu This looks fine, but just curious: how did you find out about this?

tuqqu:
:) I was going through the [changelog](https://www.php.net/ChangeLog-8.php) for PHP8 when noticed there was a dedicated RFC for this.

frederikbosch:
Thanks, attentive! Since this change does not have any actual substance, I won't create a tag for this.

Ocramius:
Note: even if it is part of core (and bundled with it), it is still a dependency.

frederikbosch:
@Ocramius And therefore it should be readded as a dependency?

Ocramius:
Maybe: no strong feelings though.

Adding maglnet/composer-require-checker would expose this and other hidden
dependencies.

frederikbosch:
Created issue #658 for this, thanks for the suggestion.

PR #668, opened 2021-09-09
Bring @gmponos PR forward with master

jameswilddev:
This is identical to #520 but caught up with master.


jameswilddev:
Right, got that to pass everything but the mutation tests.  I don't really know how to fix those (a lot of what it flags isn't stuff I've changed).

frederikbosch:
Great work @jameswilddev and @gmponos (!). To what extend is the exponential parser also able to parse normal numbers?

jameswilddev:
I'll be honest I haven't looked at the base PR in depth; I've only checked out the branch, merged master and fixed CI failures.

In my case I'm maintaining a product which uses Money and found a bug where entering "1e1" would get past validation, but then cause a parser error and figured this could well be fix.

frederikbosch:
What is the use again of this Parser/Formatter? Maybe it is time to merge this in after 4 years..... sorry guys!

gmponos:
> What is the use again of this Parser/Formatter? Maybe it is time to merge this in after 4 years..... sorry guys!

Hi, haven't fully understood the question.

- We haven't submitted a formatter.. 
- The use cases are on the tests.

Not sure if I miss something.. sorry

PR #669, opened 2021-09-15
jsonSerialize() method made compatible with PHP 8.1

daniser:
This library fails to work in PHP 8.1 environment because signatures of internal methods has changed
(see https://wiki.php.net/rfc/internal_method_return_types).

This PR adds #[ReturnTypeWillChange] attribute to jsonSerialize() methods to suppress deprecation warning on PHP 8.1.

### DIFF
--- a/src/Currency.php
+++ b/src/Currency.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Money;
 
 use JsonSerializable;
+use ReturnTypeWillChange;
 
 use function strtoupper;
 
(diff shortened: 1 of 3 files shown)


frederikbosch:
This will not be merged before the arrival of PHP 8.1, and not before PHP 8.1 has been added to the CI environment (Github Actions).

frederikbosch:
I'll close this in favor of the PR by @driesvints that addresses more 8.1 issues, and has another fix for the serialization problem.

PR #670, opened 2021-10-08
Full PHP 8.1 support

driesvints:
~Blocked by https://github.com/phpspec/phpspec/pull/1397 a new tag on PhpSpec.~

Issue #656, opened 2021-06-16
JSON serialization return type warning under PHP 8.1


Firehed:
When starting to test on the 8.1 alpha builds, I encountered the following deprecation warnings from this library:

```
PHP Deprecated:  Declaration of Money\Currency::jsonSerialize() should be compatible with JsonSerializable::jsonSerialize(): mixed in /var/www/html/vendor/moneyphp/money/src/Currency.php on line 86
PHP Deprecated:  Declaration of Money\Money::jsonSerialize() should be compatible with JsonSerializable::jsonSerialize(): mixed in /var/www/html/vendor/moneyphp/money/src/Money.php on line 521
```

I'm currently still on `3.x`, but it looks like the issue is present on `4.x` and HEAD.

It looks like adding native return types (replacing the docblock annotations) fixes the issue, and as the two classes are declared `final` I don't think it should create any BC issues.

Happy to send a PR to fix if you'd like - let me know!

frederikbosch:
Reopen when 8.1 is out of the door and issue still there. We will not create fixes for alpha releases of PHP.

Firehed:
Will do :)

someniatko:
@frederikbosch Hi, it is still the same for the PHP 8.1RC6, which is the last pre-release version.

peterdevpl:
@Firehed still interested in submitting the PR? :)

jaikdean:
I've confirmed this to also be present under the GA release of PHP 8.1.0.

Firehed:
I'm more than happy to create a PR to fix this, but I see at least one open that's trying to do the same thing. Can any maintainer comment on how they'd like to proceed?

frederikbosch:
PR #670 is the way forward. At the moment incompatible dependencies - phpspec in particular - are blocking the merge of this PR now.

AlexandruGG:
> PR #670 is the way forward. At the moment incompatible dependencies - phpspec in particular - are blocking the merge of this PR now.

hey @frederikbosch I also opened #675, could you please have a look? I think dev dependencies don't need to block this as it's quite a small change

frederikbosch:
Support is there, see version 4.0.3. Thanks for the help.

PR #675, opened 2021-11-30
Add `jsonSerialize` Return Types

AlexandruGG:
Add `jsonSerialize` return types
- fixes PHP 8.1 deprecations
- shows array shapes where applicable

Addresses: https://github.com/moneyphp/money/issues/656
Related to PR: https://github.com/moneyphp/money/pull/670

While we can't say that Money fully supports 8.1 until the workflows are updated to run on it, the benefit of this is we don't have to wait for PHPSpec fixes from https://github.com/phpspec/phpspec/pull/1397. Once that is fixed we can have CI running on 8.1 as well. 

Also note that these changes are completely safe even without running workflows on 8.1.

### DIFF
--- a/src/Currency.php
+++ b/src/Currency.php
@@ -56,10 +56,8 @@ final class Currency implements JsonSerializable
 
     /**
      * {@inheritdoc}
-     *
-     * @return string
      */
-    public function jsonSerialize()
+    public function jsonSerialize(): string
     {
         return $this->code;
     }
--- a/src/CurrencyPair.php
+++ b/src/CurrencyPair.php
@@ -107,9 +107,9 @@ final class CurrencyPair implements JsonSerializable
     /**
      * {@inheritdoc}
      *
-     * @return array
+     * @psalm-return array{baseCurrency: Currency, counterCurrency: Currency, ratio: numeric-string}
      */
-    public function jsonSerialize()
+    public function jsonSerialize(): array
     {
         return [
             'baseCurrency' => $this->baseCurrency,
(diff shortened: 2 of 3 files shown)


AlexandruGG:
hey @sagikazarmark @driesvints 👋 

What do you think of this one? Would allow users to update to PHP 8.1 without the deprecations

driesvints:
I don't think it's safe to merge this without also running the builds. You'll be left in the dark if anything goes amis.

driesvints:
Another option is to drop PhpSpec altogether if we don't want to wait and convert all tests to PHPUnit tests.

AlexandruGG:
> I don't think it's safe to merge this without also running the builds. You'll be left in the dark if anything goes amis.

I'm not suggesting to merge this without running the _existing_ CI builds.

If you're referring to running the workflows on 8.1 like [here](https://github.com/moneyphp/money/pull/670/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR39), I think that's definitely something to aim for but I see it as unrelated to this PR. The changes here are quite small, adding those return types shouldn't make any difference in terms of functionality in either 8.0 or 8.1.

driesvints:
@AlexandruGG I meant the PHP 8.1 builds. I don't think it's safe to merge 8.1 support if there's no actual builds to verify the implementation.

driesvints:
@AlexandruGG Ah I finally see what you mean. Yes, you're correct, that sounds reasonable 👍

frederikbosch:
@AlexandruGG Thank you, your commit was merged in the final PR that also fixed the CI pipeline, see #676.

PR #678, opened 2022-01-11
Reduce PHP support

frederikbosch:
Following community discussions, I think we should opt-out to automatic support for every PHP 8 version. For now we fully support PHP 8.0, and have preliminary support for PHP 8.1. Since PHP does not follow semantic versioning (minor versions can have breaking changes), we should not communicate the support.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,13 +1,12 @@
 {
     "name": "moneyphp/money",
     "description": "PHP implementation of Fowler's Money pattern",
+    "license": "MIT",
     "keywords": [
         "money",
         "vo",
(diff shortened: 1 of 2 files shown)


frederikbosch:
If @Ocramius does have any comment on this, I'd love to hear from him. He uplifted the lib most recently and is the inspiration for this change.

Ocramius (approved):
👍

frederikbosch:
Merci

frederikbosch:
@sagikazarmark If you agree, I will merge this in.

PR #685, opened 2022-03-21
Conversion ratio is a string

jongotlin:


frederikbosch:
You have to pass the ratio as a string, not as a float, to the constructor. This is something that has changed since version 4. I am updating the docs in #686. In the same PR I will update this CHANGELOG too. For converting the float to a string, you can use something like `sprintf('%.14F', $float)`.

jongotlin:
Ok! Thank you. 
Note, you have the same casting in https://github.com/moneyphp/money/blob/master/src/Exchange/SwapExchange.php#L34 if you want to switch to sprintf.

PR #688, opened 2022-04-14
Allow divisor in mod to be a number as well

### LINKED ISSUE #687: Why is divisor argument in mod method of type Money
Hey, forgive me if this is a dummy question. I might be missing something totally obvious.

But I was asking myself, why the argument for the [`mod` method](https://github.com/moneyphp/money/blob/2fa15bbdfafe9be39ff57e4d3c64a18add719f73/src/Money.php#L275) is of type `Money`. Shouldn't the` $divisor` argument here be the same as for the [`divide` method](https://github.com/moneyphp/money/blob/2fa15bbdfafe9be39ff57e4d3c64a18add719f73/src/Money.php#L259)?

Thanks!

helhum:
Fixes #687

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -272,14 +272,22 @@ final class Money implements JsonSerializable
      * the remainder after dividing the value by
      * the given factor.
      */
-    public function mod(Money $divisor): Money
+    public function mod(Money|int|string $divisor): Money
     {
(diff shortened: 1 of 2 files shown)


frederikbosch:
I am not against this. It seems fair you can use a number too.

w3bdesign (approved):

frederikbosch:
Thanks, took more than a year, but I have merged it!

PR #689, opened 2022-04-18
Public method

ricardogobbosouza:
Sometimes we need all the currency information

Resolves https://github.com/cknow/laravel-money/issues/110


frederikbosch:
Won't do, you can talk to the repository [files included in the library](https://github.com/moneyphp/money/blob/master/resources/currency.php). Or if you need different formats, use the [iso-currencies package](https://github.com/moneyphp/iso-currencies).

PR #690, opened 2022-04-21
clarify exactly what the BC Break is

cabbey:
Documentation only change to make it clear what folks need to look for.


frederikbosch:
Thank you!

PR #698, opened 2022-07-04
Fix: ISO parsing invalid assertions for CurrencyPair

### LINKED ISSUE #695: Unit Test for CurrencyPair - iso string parse FAILS
Running the unit tests against the built lib, all pass but 1. 

```
There was 1 failure:

1) Tests\Money\CurrencyPairTest::it_parses_an_iso_string
assert(is_numeric($matches[2])) in /app/src/CurrencyPair.php:65

FAILURES!
Tests: 1790, Assertions: 2599, Failures: 1.
```

Should be a fairly straight forward fix to get this to pass. I'll take a look myself if I get time and submit a PR. Otherwise leaving this here for a contributor to pick up!

Notable Files@Methods:

src/CurrencyPair@createFromIso
tests/CurrencyPairTest@it_parses_an_iso_string

Thanks :)

sudkumar:
With iso as `EUR/USD 1.2500`, we will get `$matches = ['EUR/USD 1.2500', 'EUR', 'USD', '1.2500']` after `pref_match` but the assertion expects `$matches[2]` to be numeric which seems to be incorrect. Just updated the assertion to not expect an empty string.

```diff
assert($matches[1] !== '');
- assert(is_numeric($matches[2]));
+ assert($matches[2] !== '');

```

fixes #695

> **Warning** : Not ready for merge. 
> The changes have been reverted to showcase the issue with existing implementation.

### DIFF
--- a/spec/CurrencyPairSpec.php
+++ b/spec/CurrencyPairSpec.php
@@ -27,6 +27,14 @@ final class CurrencyPairSpec extends ObjectBehavior
         $this->shouldImplement(JsonSerializable::class);
     }
 
+    public function it_parses_an_iso_string(): void
+    {
+        $this->equals(
+            new CurrencyPair(new Currency('EUR'), new Currency('USD'), '1.250000'),
+            CurrencyPair::createFromIso('EUR/USD 1.250000')
+        );
+    }
+
     public function it_has_currencies_and_ratio(): void
     {
         $this->beConstructedWith($base = new Currency('EUR'), $counter = new Currency('USD'), $ratio = '1.0');
--- a/src/CurrencyPair.php
+++ b/src/CurrencyPair.php
@@ -62,7 +62,7 @@ final class CurrencyPair implements JsonSerializable
         }
 
         assert($matches[1] !== '');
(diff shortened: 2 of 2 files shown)


frederikbosch:
Can you include a test on the example that fails?

sudkumar:
@frederikbosch I have reverted the implementation and added a test to show the failure.

frederikbosch:
@sudkumar The tests pass. So the question is: why does it fail on your end?

frederikbosch:
What is inside $matches when you execute the test? Can you show the result of the following var_dump here?
```php
var_dump($iso, $matches);
```

sudkumar:
What do you get with `var_dump($iso, $matches);` ?

sudkumar:
CleanShot 2022-08-10 at 20 12 09

Am I missing something obvious here!! 😄

frederikbosch:
@sudkumar Thank you very much, I have fixed the issue. `zend.assertions` was not enabled in CI. That's why it was not throwing. Your observation was correct.

sudkumar:
So glad we found the issue. Thank you for looking into it.

PR #702, opened 2022-08-17
fix decimal money parser to accept leading zeros

frederikbosch:
includes test for leading zeros in all parsers

PR #704, opened 2022-08-25
Accept leading zeros

PowerKiKi:
This is a direct follow-up to 2f1dbf1afde48dbe98402e5e8aa450434def234f, from #702,
that started to allow leading zeros in some places, but did not actually allow them everywhere.

### DIFF
--- a/doc/getting-started.rst
+++ b/doc/getting-started.rst
@@ -19,7 +19,7 @@ See :doc:`features/parsing` for additional ways to instantiate a Money object fr
 
 Accepted integer values
 -----------------------
-The Money object only supports integer(ish) values on instantiation. The following is (not) supported. When a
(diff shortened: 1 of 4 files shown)


PowerKiKi:
@frederikbosch, it would probably be best to merge this PR before cutting a release, otherwise the PR #702 would be not be super useful

frederikbosch:
@PowerKiKi I would prefer to only accept leading zero's in the parsers, and not in the Number/Money objects.

PowerKiKi:
You know what ? I think you are right. Parsing with `DecimalMoneyParser` or creating a new `Money` object is not the same use-case. And it makes sense that `Money` might be stricter than `DecimalMoneyParser`.

I wished I had thought a bit more about that before creating this PR 😅

PR #706, opened 2022-09-04
fix deprecation on jsonSerialize

### LINKED ISSUE #705: Deprecated warning on php 8.1.8 for different signature of `jsonSerialize`
`PHP Deprecated:  Return type of Money\Money::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in Money.php on line 521`

Found this thread on a different package that might help https://github.com/ramsey/uuid/issues/381#issuecomment-1035461859

I think that to avoid a breaking change - we can just add the `#[ReturnTypeWillChange]` attribute to the method?

garak:
Fix #705

### DIFF
--- a/src/Currency.php
+++ b/src/Currency.php
@@ -79,6 +79,7 @@ final class Currency implements \JsonSerializable
      *
      * @return string
      */
+    #[\ReturnTypeWillChange]
     public function jsonSerialize()
     {
         return $this->code;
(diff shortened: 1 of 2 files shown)


frederikbosch:
I rather add the real return-type instead of the attribute, so: `: string` for `Currency` and `: array` for `Money`.

frederikbosch:
Oh, now I see, this is for Money v3. No, there will not be support for PHP8+ in Money v3. Please use Money v4.

PR #710, opened 2022-09-13
Add underlying provider name to the currency pair

morloderex:
Hey MoneyPHP team.

I am doing some currency conversations using the swap implementation as the baseline, which internally uses the exchange framework behind the scenes using its chain functionality.

I have been asked to save the conversion rate and the provider that was used to fetch that conversion rate somewhere. Which brings up a question that i am trying to solve "How can i get the provider that was used to fetch the conversion rate?"

I believe i have managed to solve that issue for the different implementations that we have for our exchange functionality by also allowing the "CurrencyPair" to hold that information for us.

I hope this will make it in so i can get the context i need.

If changing method signatures is considered a breaking change i would be open for changing that around.

### DIFF
--- a/spec/CurrencyPairSpec.php
+++ b/spec/CurrencyPairSpec.php
@@ -14,7 +14,7 @@ final class CurrencyPairSpec extends ObjectBehavior
 {
     public function let(): void
     {
-        $this->beConstructedWith(new Currency('EUR'), new Currency('USD'), '1.250000');
+        $this->beConstructedWith(new Currency('EUR'), new Currency('USD'), '1.250000', 'fixed');
     }
 
     public function it_is_initializable(): void
@@ -47,5 +47,10 @@ final class CurrencyPairSpec extends ObjectBehavior
     public function it_throws_an_exception_when_iso_string_cannot_be_parsed(): void
     {
         $this->shouldThrow(InvalidArgumentException::class)->duringCreateFromIso('1.250000');
+    }
+
+    public function it_returns_the_provider_name_when_asked_for(): void
+    {
+        $this->providerName()->shouldReturn('fixed');
     }
 }
--- a/spec/Exchange/FixedExchangeSpec.php
+++ b/spec/Exchange/FixedExchangeSpec.php
@@ -17,7 +17,7 @@ final class FixedExchangeSpec extends ObjectBehavior
     {
         $this->beConstructedWith([
             'EUR' => ['USD' => '1.25'],
-        ]);
+        ], 'provider-name');
     }
 
     public function it_is_initializable(): void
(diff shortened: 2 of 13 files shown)


frederikbosch:
You should solve this on your own end. Create a new (value) object and your own exchange.

```php
<?php
use Exchanger\Contract\ExchangeRateProvider;
use Exchanger\CurrencyPair as ExchangerCurrencyPair;
use Exchanger\Exception\Exception as ExchangerException;
use Exchanger\ExchangeRateQuery;
use Money\Currency;
use Money\CurrencyPair;
use Money\Exception\UnresolvableCurrencyPairException;
use Money\Exchange;

use function assert;
use function is_numeric;
use function sprintf;

final class ProvidedCurrencyPair 
{
    private CurrencyPair $pair;
    private string providerName;

    public function __construct(CurrencyPair $pair, string $providerName) 
    {
        $this->pair = $pair;
        $this->providerName = $providerName;
    }

    public function getPair(): CurrencyPair
    {
         return $this->pair;
    }

    public function getProviderName(): string
    {
         return $this->providerName;
    }
}

final class MyExchange implements Exchange 
{
    private ExchangeRateProvider $exchanger;

    public function __construct(ExchangeRateProvider $exchanger)
    {
        $this->exchanger = $exchanger;
    }

    public function quoteWithProviderName(Currency $baseCurrency, Currency $counterCurrency): CurrencyPair
    {
        try {
            $query = new ExchangeRateQuery(
                new ExchangerCurrencyPair($baseCurrency->getCode(), $counterCurrency->getCode())
            );
            $rate  = $this->exchanger->getExchangeRate($query);
        } catch (ExchangerException) {
            throw UnresolvableCurrencyPairException::createFromCurrencies($baseCurrency, $counterCurrency);
        }

        $rateValue = sprintf('%.14F', $rate->getValue());

        assert(is_numeric($rateValue));

        return new ProvidedCurrencyPair(
            new CurrencyPair($baseCurrency, $counterCurrency, $rateValue), 
            $rate->getProviderName()
        );
    }

    public function quote(Currency $baseCurrency, Currency $counterCurrency): CurrencyPair 
    {
        return $this->decoratedExchange->quoteWithProvider($baseCurrency, $counterCurrency)->getPair();
    }
}

frederikbosch:
Not dropping the `final` keyword, anywhere. You don't need to extend, you need a new class, e.g. `ProvidedCurrencyPair` as I called in the example.

morloderex:
@frederikbosch  Okay can you give me an example where i can get the providerName from calling. 

`[$converted, $pair] = (new \Money\Converter())->convertAndReturnWithCurrencyPair(new \Money\Money(1000, new \Money\Currency('EUR')), new \Money\Currency('EUR'))`

Without having to reinvent the `\Money\Converter` class also sense i cannot see how i can `just` decorate it?

PR #715, opened 2022-09-29
Support PHP 8.2

rvanlaak:
Seems PHP 8.2 is supported without having to make any changes.

phpspec is not supporting 8.2 yet, but [work is in progress](https://github.com/phpspec/phpspec/pull/1435). This is blocking us from adding php 8.2 support for now with the current set of requirements.

PR #717, opened 2022-10-18
Support Installation on PHP 8.2

gsteel:
Prophecy appears to be the only dependency that explicitly declares it's not compatible with 8.2 - adding `--ignore-platform-reqs` allows CI to pass, but feel free to close if you're waiting for all deps to be compatible first 👍


gsteel:
Sorry - Closing as duplicate of #715

Issue #671, opened 2021-10-09
Composer.lock is commited


sagikazarmark:
I'm trying to understand why the `composer.lock` file is committed.

The change was made by @Ocramius here: https://github.com/moneyphp/money/commit/0d1ec59efd43e66d2f3d671eb390d110055944f2

The reason being that it will lock the versions for tools like Psalm, PHPCS and PHPStan.

My questions are:
- Why not just pin the versions in `composer.json`?
- Why not use the [composer-bin-plugin](https://github.com/bamarni/composer-bin-plugin) for these dependencies?

My primary concern (apart from the potential git conflicts) is that people will start bumping dependencies and it will get much harder to review what changed compared to a single version update or psalm dedicated composer.json/lock.

@Ocramius @frederikbosch any thoughts?

frederikbosch:
I prefer not to commit `composer.lock` but that is mostly based on arguments that were valid years ago. In defense of not committing it, I would think - but @Ocramius might reject my argument for valid reasons - that not committing `composer.lock` leaves contributors with the problem of creating a valid PR that passes all checks. And when a PR results in an error with tools like Psalm, PHPCS and PHPStan due to a version upgrade, that it is up to the contributor to solve it. The contributor can choose to narrow a version, or to fix the issue with the tool at hand. 

Maybe @Ocramius is willing to explain his reasons, or to link to a blog/article/text that contains the reason.

Ocramius:
I explained this in more detail at https://github.com/mariosimao/notion-sdk-php/pull/6#discussion_r744177586

Quoting:

> 
> Yes, `locked` means using a committed `composer.lock`.
> 
> You can put your `composer.lock` in `export-ignore`, and it will be ignored regardless by downstream `composer install` and `composer update` operations.
> 
> The reasons why `composer.lock` is valuable is that tools like `vimeo/psalm`, `phpunit/phpunit` and `phpcs/php_codesniffer` often contain bugfixes and tiny improvements that completely destabilize the whole downstream consumer CI pipelines. This is also not fixable on their end: if `vimeo/psalm` slightly improves the type inference of an internal PHP function, for example, you might get reported `RedundantCondition` or similar errors in your pipeline.
> 
> By running them **locked**, you can come back to this project one year from now:
> 
>  * `latest` pipeline will have "discovered" any changes that might be problematic
>  * `lowest` pipeline might highlight how some dependencies no longer work with your code changes (signatures, mostly)
[...]

PR #722, opened 2022-12-15
Delete composer.lock

rogervila:
Hello,

Packages usually do not contain a lock file since it can cause unexpected collisions with other packages installed on a project.

Also, it makes it more difficult to support multiple PHP versions since it ties dependencies to specific versions, which might not support newer PHP versions.


rogervila:
@frederikbosch this PR would solve the issue on #715

rvanlaak:
Worth knowing is that composer solely uses the root project's lock file, and does not consider lock file of dependencies: https://getcomposer.org/doc/02-libraries.md#lock-file

sagikazarmark:
BTW this has been discussed here: #671

I don't really care, but going back and forth isn't that great IMO.

rvanlaak:
About [the related commit message](https://github.com/moneyphp/money/commit/0d1ec59efd43e66d2f3d671eb390d110055944f2):

> PHPCS, PHPStan, Psalm and similar tools are too unstable to lock at minor/major release,
therefore we need a `composer.lock` to keep things frozen during builds.

The CI jobs indeed do run via the dependency versions as pinned down in the lock files, [except for the "test lowest" CI job](https://github.com/moneyphp/money/blob/master/.github/workflows/ci.yml#L29).

frederikbosch:
Sorry guys, I just accepted the change. Maybe I should have left it in. Maybe I should have opened discussion. Don't know. I think it is still rather unusual to commit the `composer.lock`. It was not the first time a PR came with a changed lock file.

Also considering that amount of PRs is not that much - this library is not in phase of creating features - that chances are when a PR is there, it is when a major dependency has changed - with PHP version being the most likely one. This most probably will lead to updating dev dependencies. This was the case first with the 4.1.0 release, required updating PHP spec and then Psalm. With PHP 8.1 support we were also waiting for a PHP spec update.

 So I decided remove composer.lock but tighten Psalm to `~5.3.0`, which I think should lead to stable builds too. While going and forth is not that great - agreed - I suggest to see what happens.

PR #723, opened 2022-12-15
Fix psalm errors

rogervila:
This PR removes some invalid Psalm suppressions, and ignores `composer.lock`


rogervila:
@frederikbosch this PR solves the issue. Anyway, I will work on a new one to upgrade Psalm from v4 to v5, after migrating the PHP 8.2 PR.

PR #725, opened 2022-12-15
Psalm 5

rogervila:
This PR upgrades Psalm from v4 to v5 and adds some fixes raised by the new version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 
 doc/_build/
 composer.lock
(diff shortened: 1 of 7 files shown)


rogervila:
@frederikbosch Should PHP 8.2 support be released v4.0.6, or will it be a new major version? 
If so, this branch could be merged before releasing it. Otherwise, it does not matter when this branch is merged.

Thank you!

PR #726, opened 2022-12-15
add crypto currencies

frederikbosch:
This PR adds a `CryptoCurrencies` class, quite similar to `ISOCurrencies`, but now crypto. We are using our [`moneyphp/crypto-currencies`](https://github.com/moneyphp/crypto-currencies) repository for this.


frederikbosch:
@slunak WDYT?

resources/generate-money-factory.php
```diff
@@ -49,15 +48,21 @@ public static function __callStatic(string $method, array $arguments): Money
 
     $methodBuffer = '';
 
-    $currencies = iterator_to_array(new Currencies\AggregateCurrencies([
+    $iterator = new Currencies\AggregateCurrencies([
         new Currencies\ISOCurrencies(),
         new Currencies\BitcoinCurrencies(),
```
rogervila, line 53: Maybe BitcoinCurrencies should be deprecated/removed since the CryptoCurrencies class already handles BTC

rogervila (changes requested):

resources/generate-money-factory.php
slunak: I agree with deprecating BitcoinCurrencies.

As for XBT, I would suggest to discuss here:

https://github.com/moneyphp/crypto-currencies/issues/2

src/MoneyFactory.php
```diff
@@ -54,7 +54,6 @@
  * @method static Money ATOM(numeric-string|int $amount)
  * @method static Money AUCTION(numeric-string|int $amount)
  * @method static Money AUD(numeric-string|int $amount)
- * @method static Money AUD(numeric-string|int $amount)
```
slunak: How about removing fiat currencies in https://github.com/moneyphp/crypto-currencies?

I think it is more elegant to remove them at that stage because crypto-currencies library should provide only crypto currencies. The reason it provides some fiat currencies at the moment is that because they are listed listed on Binance. I can ignore them and provide only crypto currencies.

resources/generate-money-factory.php
```diff
@@ -49,15 +48,21 @@ public static function __callStatic(string $method, array $arguments): Money
...
 
+    $currencies = array_unique([...$iterator]);
     usort($currencies, static fn (Currency $a, Currency $b): int => strcmp($a->getCode(), $b->getCode()));
 
     /** @var Currency[] $currencies */
     foreach ($currencies as $currency) {
+        if (is_numeric($currency->getCode()[0])) {
+            // skip currencies that start with a number, as php function has to start with a letter or underscore
+            continue;
+        }
```
slunak, line 65: There is currently only one crypto currency in resources starting with an integer, 1INCH. It would be good to include it. Perhaps I can rename it to ONEINCH.

src/MoneyFactory.php
frederikbosch: Please do.

resources/generate-money-factory.php
frederikbosch: Codes starting a number are now transformed to start with the number expressed as a word.

src/MoneyFactory.php
frederikbosch: I have created [an issue in the crypto currencies lib](https://github.com/moneyphp/crypto-currencies/issues/3). Do you time to fix this @slunak?

slunak: Yeah, I will look after it. @frederikbosch I was looking into more options how to fetch crypto currencies from other available sources. Ideal scenario would be to fetch them from cryptomarketcap platform. The only caveat I see is that a person that fetches them will need to have a free accout with cryptomarketcap and use their API key. As I said it is ideal scenario because the platform has all the crypto currencies needed and does not include fiat currencies. What do you think about this scenario @frederikbosch ? If you are ok with it, then I will explore it more, otherwise I will just fix our current source, Binance. Downside of this approach is that in case Binance adds more fiat currencies then we will have to manually fix them too.

frederikbosch: I am not against that. I have not worked with secrets in Github actions, I guess there must an option for that.

slunak: Ok, let me fix Binance first and then we call look into coinmarketcap option. P.S. I said cryptomarketcap in my previous comment, but I made a mistake, the platform is called coinmarketcap.

resources/generate-money-factory.php
rogervila: @frederikbosch BitcoinCurrencies will be integrated inside CryptoCurrencies so the feature will be duplicated. 

Marking it as deprecated will help developers to move to the new CryptoCurrencies without adding any breaking changes. 

Do you agree on deprecating it on the next major version at least?

frederikbosch: @rogervila the code `XBT` is not included in `CryptoCurrencies`, as it is included with the code `BTC`. Hence, I would like to leave it in. But even if it would contain `XBT`, deprecation suggests future removal, and I would only want to disturb people with refactoring their code that depends on this library if we have a technical reason for it. And I believe that reason is not there.

rogervila: Ok. Approving the PR.

rogervila (approved):

frederikbosch:
@rogervila Thanks for your help here!

frederikbosch:
Thanks @rogervila @slunak !

Issue #728, opened 2022-12-16
FixedExchange currency conversion problem with ROUND_UP


tashik:
```
$exchange = new FixedExchange([
      'USD' => [
              'KZT' => "374.3",
       ]   
]);
$converter = new Converter(new ISOCurrencies(), $exchange);
$amount = $converter->convert(new Money(800, new Currency('USD')), new Currency('KZT'), Money::ROUND_UP);
print($amount->getAmount());
```
>>> 299441
instead of 29940. 
As you can see, I pass a string "374.3" when creating and exchange instance, but in your code (CurrencyPair __construct) you convert my correct string to incorrect float. Then in convert method of Converter class you try to get a string from that float - and fail to get correct string.
The problem is that currency conversion ratio before multiplying becomes 374.3000000001.
Please, advise, why so and how to fix this?

We use php/money 3.x since our project is PHP 7.2.4

frederikbosch:
We only support [versions supported by PHP](https://www.php.net/supported-versions.php), hence no support for PHP 7.2.

PR #729, opened 2022-12-17
fix issue #728

tashik:
for those who are still at PHP 7.x, fix currency conversion issue with round_up rounding mode (remove float typing)


frederikbosch:
There is no more work on MoneyPHP 3.x. I made an exception for adding attributes/comments. Please upgrade to 4.x.

PR #731, opened 2023-01-13
numeric-string DecimalMoneyFormatter return type

rogervila:
`DecimalMoneyFormatter` format method returns always a `numeric-string` value.


rogervila:
Well, It needs additional Psalm config. @frederikbosch before going on with this PR, do you agree with the change?

frederikbosch:
I think the new docblock is correct indeed, decimal formatter returns a numeric-string.

frederikbosch:
I am not a Pslam guru, but if that error is fixed, I can merge this.

rogervila:
Thank you @frederikbosch, I have fixed the PR, let me know if there is any additional change needed.

PR #735, opened 2023-02-04
Money::ratioOf throws exception if currencies don't match

### LINKED ISSUE #732: ratioOf works incorrect for currencies with different minor units
`ratioOf` assumes that the given amount has the same minor units as the current amount, otherwise the result is incorrect:

```php
Money::EUR(20_00)->ratioOf(Money::JPY(1000))
```
expected: 0.02
actual: 2.00

Probably `ratioOf` should throw an exception when currencies don't match to avoid unexpected results.

peterdevpl:
I also added a test case for `Money::mod` which already performs currency check.

fixes #732

### DIFF
--- a/tests/MoneyTest.php
+++ b/tests/MoneyTest.php
@@ -339,7 +339,6 @@ final class MoneyTest extends TestCase
 
         $this->expectException(InvalidArgumentException::class);
 
-        /** @psalm-suppress UnusedMethodCall this method throws */
         $money->ratioOf($rightMoney);
     }
 
--- a/tests/MoneyTest.php
+++ b/tests/MoneyTest.php
@@ -260,6 +260,20 @@ final class MoneyTest extends TestCase
     /**
      * @test
      */
+    public function itThrowsWhenCalculatingModulusOfDifferentCurrencies(): void
+    {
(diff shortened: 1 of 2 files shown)


tests/MoneyTest.php
```diff
@@ -315,6 +329,20 @@ public function itThrowsWhenCalculatingRatioOfZero(): void
...
+     */
+    public function itThrowsWhenCalculatingRatioOfDifferentCurrencies(): void
+    {
+        $money      = new Money(self::AMOUNT, new Currency(self::CURRENCY));
+        $rightMoney = new Money(self::OTHER_AMOUNT, new Currency(self::OTHER_CURRENCY));
+
+        $this->expectException(InvalidArgumentException::class);
+
+        /** @psalm-suppress UnusedMethodCall this method throws */
+        $money->mod($rightMoney);
```
smelesh, line 343: it looks like `ratioOf` should be there

smelesh (changes requested):

tests/MoneyTest.php
peterdevpl: Good catch, fixed!

frederikbosch:
Nice catch, will merge when green checks. Thanks!

Issue #740, opened 2023-02-24
`roundToUnit` does not handles amounts less than 1 unit


bendavies:
Hi there,

Given:

```php
$currencies = new \Money\Currencies\ISOCurrencies();
$money = \Money\Money::USD(10); //10 cent

$money->roundToUnit($currencies->subunitFor($money->getCurrency()));
```

should return  `\Money\Money::USD(0)`

but will throw:

```
Fatal error: Uncaught InvalidArgumentException: Leading zeros are not allowed in /vendor/moneyphp/money/src/Number.php:216
Stack trace:
#0 /vendor/moneyphp/money/src/Number.php(44): Money\Number::parseIntegerPart('000')
#1 /vendor/moneyphp/money/src/Number.php(53): Money\Number->__construct('000', '')
#2 /vendor/moneyphp/money/src/Money.php(81): Money\Number::fromString('000')
#3 /vendor/moneyphp/money/src/Money.php(410): Money\Money->__construct('000', Object(Money\Currency))
#4 /test.php(10): Money\Money->roundToUnit(2)
#5 {main}
```

Thanks

frederikbosch:
I am not able to reproduce this. What should I change in PR #741?

bendavies:
looking...

bendavies:
@frederikbosch example updated from 1 cent to 99 cents.

PR #741, opened 2023-02-28
Try reproduce bug #740

frederikbosch:
This PR tries to reproduce issue #740


tests/MoneyTest.php
```diff
@@ -584,6 +584,7 @@ public function roundToUnitExamples(): array
             [1250, 2, 1200, Money::ROUND_DOWN],
             [1250, 2, 1300, Money::ROUND_HALF_POSITIVE_INFINITY],
             [1250, 2, 1200, Money::ROUND_HALF_NEGATIVE_INFINITY],
+            [1, 2, 0, Money::ROUND_HALF_UP],
```
bendavies, line 587: try any value from 10-99 cents

frederikbosch: 10-49 indeed fails, 99 does work.

bendavies: erg. i should try my examples before posting :D

PR #742, opened 2023-03-06
abstract CurrencyPair to an interface

morloderex:
Currently we can create our own Exchange objects, but we are forced to use this library's CurrencyPair ValueObject.

This Pull Request tries to lift that restriction by extracting the methods inside the CurrencyPair into an interface that can be used as a return value while implementing our own Exchange values. 

The use-case is to allow calling converter#convertAndReturnWithCurrencyPair and have it return my custom implementation back to me if i implement my own Exchange implementation. 

Previously it was impossible to retrieve any other information from the CurrencyPair other than the currencies involved and the Rate for it. With this proposal a Developer would be able to provide their own implementation and that implementation could consist of e.g. the date or the underlaying provider used for the exchange. 

Superseds #710.

Maybe we should even provide a trait for very basic logic that currently goes into creating the CurrencyPair ValueObject?

What do you think of this?

### DIFF
--- a/doc/features/currency-conversion.rst
+++ b/doc/features/currency-conversion.rst
@@ -186,3 +186,14 @@ A currency pair can be used to convert an amount.
     $converter = new Converter(new ISOCurrencies(), $exchange);
     $eur100 = Money::EUR(100);
     $usd125 = $converter->convertAgainstCurrencyPair($eur100, $pair);
+
+Implementing your own CurrencyPair
+------------
+Sometimes you would like to have retrieve more data than our CurrencyPair provides
+perhaps you would like to store which date the currency conversion was made.
+
+In order to do so you can create your very own CurrencyPair valueObject.
+
+That object needs to implement \Money\Exchange\CurrencyPair.
+
+Then you can create your own exchange and have it return that CurrencyPair instead of the default \Money\CurrencyPair object.
--- a/src/Converter.php
+++ b/src/Converter.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Money;
 
 use InvalidArgumentException;
+use Money\Exchange\CurrencyPair;
 
 use function sprintf;
 
--- a/src/Exchange.php
+++ b/src/Exchange.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Money;
 
 use Money\Exception\UnresolvableCurrencyPairException;
+use Money\Exchange\CurrencyPair;
 
 /**
  * Provides a way to get exchange rate from a third-party source and return a currency pair.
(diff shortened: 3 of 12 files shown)


frederikbosch:
While I get where you are going with this, I wonder why you must talk to the `Exchange` interface defined in this package.

```php
namespace Vendor;

interface Exchange {
  public function quote(Currency $baseCurrency, Currency $counterCurrency): MyReturnType;
}

final class MyExchange implements Exchange {
  public function __construct(private Money\Exchange $moneyExchange) {}

  public function quote(Currency $baseCurrency, Currency $counterCurrency): MyReturnType
  {
       $currencyPair = $this->moneyExchange->quote($baseCurrency, $counterCurrency);
       return $this->convertToOrAddSpecificsAboutMyReturnType($currencyPair);
  }
}

morloderex:
@frederikbosch We are very deep in the code but the basic idea is to not having to totally rewrite the `\Money\Converter` class entirely if a developer needs more information than the `\Money\currencyPair` gave you which is what i am currently doing.

I from a techical  care about how the actual `\Money\Money` object is constructed with the counter currency i just care about

frederikbosch:
I disagree.

```php
namespace Vendor;

interface MyExchange {
  public function quote(Currency $baseCurrency, Currency $counterCurrency): MyReturnType;
}

final class MyConverter {
  public function __construct(private MyExchange $myExchange, private Money\Converter $moneyConverter) {}

  /** @return array{0: Money, 1: MyReturnType} */
  public function convertAndReturnWithMyReturnType(Money $money, Currency $counterCurrency, int $roundingMode = Money::ROUND_HALF_UP): array
  {
      $myPair = $this->myExchange->quote(
          $money->getCurrency(),
          $counterCurrency
      );

      $moneyCurrencyPair = new CurrencyPair($myPair->getBaseCurrency(), $myPair->getCounterCurrency(), $myPair->getConversionRatio());

      return [$this->converter->convertAgainstCurrencyPair($money, $moneyCurrencyPair, $roundingMode), $myPair];
  }
}

morloderex:
@frederikbosch That's an interesting way of decorating the money converter yes. 

But i still feel like it its somewhat over over engineering this whole thing, but it might just be me.

I'll close this if you really don't like my design idea.

frederikbosch:
@morloderex I'd rather not provide an interface for value objects. Then the value object loses its purposes.  `Money\Exchange` and `MoneyConverter` are tied to `Money\CurrencyPair` by design. If it does not fit you, you can create own exchanges, converters and currency-pair-alike objects. Or, as I have suggested, have your own exchange and converter next to ours. Per use-case you can decide what to inject `Money\Converter`, with limited conversion meta data, or `Vendor\Converter` with extended meta data. I do this quite often in my own applications.

Issue #750, opened 2023-05-16
Bring back HRK currency


gjuric:
v4.1.1 removed Croatian Kuna from the list of currencies. 

Although Croatia has switched to Euro as of January 1th 2023, there is still obligation to show both prices till the end of this year.

Not to mention that historical data is going nowhere.

Therefore I would request that HRK is reinstated as a currency.

frederikbosch:
The ISO Currencies class is linked to the [moneyphp/iso-currencies package](https://github.com/moneyphp/iso-currencies), which in turn uses [a third party list](https://github.com/moneyphp/iso-currencies/blob/master/bin/fetch-update.php). That is not going to change.

Use `AggregateCurrencies` to support unlisted currencies.

```php
$currencies = new AggregateCurrencies([
    new ISOCurrencies(),
    new CurrencyList([
        'HRK' => 2
    ])
]);
``

gjuric:
HRK  is still in https://www.six-group.com/dam/download/financial-information/data-center/iso-currrency/lists/list-three.xml that you fetch using that package but it gets dumped to resources/historic.*

Is there a chance that you include historic currencies as well in the list of ISOCurrencies()?

frederikbosch:
Sorry, we [do not maintain history](https://github.com/moneyphp/money/issues/500). You will have to create your own currencies class for this.

PR #751, opened 2023-05-16
Bring back HRK currency

gjuric:
Brings back Croatian Kuna as explained in #750.


gjuric:
It looks like my editor settings removed extra whitespace from currency.php let me know if I should fix this.

PR #760, opened 2023-11-17
Support PHP 8.3

frederikbosch:


frederikbosch:
Also need to get these docs fixed.

ruudk:
I wish this project did not depend on PHPSpec. Every year it's the same blocker 😢 
Wouldn't it make sense to use plain PHPUnit? https://tomasvotruba.com/blog/2019/03/21/how-to-instantly-migrate-phpspec-to-phpunit

frederikbosch:
Yeah, I agree. That's why I said I would even merge it despite lack of phpspec support. I would not mind getting rid of it at all. If you see time to come up with a PR, I would very much welcome it. 

With rector it might be less work than we think. Although, moving to tests folder might conflict with our current phpunit tests.

ruudk:
What if we drop the specs completely? Isn't it covered by unit tests already?

frederikbosch:
I would only do such a thing if we would know, rather than think it does.

frederikbosch:
I have increased the test coverage of the unit tests. Now every file is tested, and has a coverage of above 80%, with the far majority having a coverage of 90%+. Maybe we should indeed drop the phpsec. I am giving it another thought later this week, before really removing it.

frederikbosch:
I think this is ready to go

ruudk:
Wow, great that PHPSpec has been removed. This will make future maintenance and contributions easier! 🎉

frederikbosch:
We could also drop PHP 8.0, maybe in a separate PR. [Security support for 8.0 will be dropped in a few days](https://www.php.net/supported-versions.php). If people want to use MoneyPHP with PHP 8.0 they can use <= ~4.2.0, while PHP 8.3 would require ^4.3.0.

ruudk:
If a PHP version is EOL, we can drop it indeed. But please do it in a separate PR / release 😁

Issue #650, opened 2021-05-17
Mutliplier + Divisor - Why not a float?


internalsystemerror:
When using the `Money::multiply` and `Money::divide` methods, thanks to the changes in https://github.com/moneyphp/money/pull/648 we can pass an integer as well as a numeric string. It should also be possible to pass a float to these methods such as when calculating tax:
`$incTax = $money->multiply(1.2);`.

If you agree then I can submit a PR to make the necessary adjustments.

frederikbosch:
You would have to convert the float to string before passing it to `multiply` or `divide`. Converting floats to string has the risk of inaccuracy which of the risk of the caller. Run the following in your terminal.

```php
$a = '35';
$b = '-34.99';
echo ($a + $b);
```

Now suppose you call `$money->multiply($a + $b)`. A float passed is not to be trusted. From version 4.0 we force callers to use int and string only, doing the conversion themselves.

frederikbosch:
I will leave this open for now because this not documented well enough in the CHANGELOG.

pmochine:
@frederikbosch It would be cool if you would show this in the documentation. https://www.moneyphp.org/en/stable/features/operation.html#multiplication-division 

Now after coming back to a project float is not working

PR #762, opened 2023-12-12
Support (multiply|divide) for float numbers

ibra-hassan:
Hi team, we have a problem with multiplying or dividing by float numbers example:
```
$value = Money::EUR(800);       // €8.00
$result = $value->multiply(2.5);  // €16.00
```

Here it will just multiply by 2 only and ignore the float, but if we pass the number as a string it will work fine.
```
$value = Money::EUR(800);       // €8.00
$result = $value->multiply("2.5");  // €20.00
```

I'm using this package for a huge project and using multiply and divide functions a lot, this modification will help us to upgrade much more easily.

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -16,6 +16,7 @@ use function count;
 use function filter_var;
 use function floor;
 use function is_int;
+use function is_float;
 use function max;
 use function str_pad;
 use function strlen;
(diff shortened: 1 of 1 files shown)


UlrichEckhardt:
Just one suggestion up front: Use a linter, like e.g. Psalm or PHPStan. It would have told you that your code isn't clean. Also, use `declare(strict_types=1)` in your sources.

That said, what should `$value->multiply(1.1)` do? 1.1 can't be represented in binary without rounding. It's effectively like one third can't be represented exactly in decimal. Not accepting floating point values here actually tries to help you not make mistakes. It might work okay with 1.1 even, but there are values where the conversion to a string doesn't give the expected result. I'm not a maintainer here though, just an interested reader, so don't take this as authoritative opinion.

BTW: Your PR lacks necessary adjustments to the `@psalm-param` annotations.

HTH

Uli

UlrichEckhardt:
Just noticed, https://github.com/moneyphp/money/blob/master/doc/concept.rst#type-safety supports some of the above.

frederikbosch:
No float, [it has been explained before](https://github.com/moneyphp/money/issues/650).

PR #764, opened 2023-12-26
fix type hint for multiply and divide

### LINKED ISSUE #763: Multiply when float will going wrong result
When I use float to be a multiplier, then it will be converted to int and got wrong answer
```php
$amount = new Money(10000);
dd($amount->multiply(0.3)); // Money(0)
```

I change the type hint from int|string to int|string|float and it work functional again

from
```php
public function multiply(int|string $multiplier, int $roundingMode = self::ROUND_HALF_UP): Money
```
to
```php
public function multiply(int|string|float $multiplier, int $roundingMode = self::ROUND_HALF_UP): Money
```

ivan88217:
fix type hint for multiply and divide when use float as multiplier or divide

fix #763

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -238,7 +238,7 @@ final class Money implements JsonSerializable
      * @psalm-param int|numeric-string $multiplier
      * @psalm-param self::ROUND_*  $roundingMode
      */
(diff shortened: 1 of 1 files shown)


frederikbosch:
No float, [it has been explained before](https://github.com/moneyphp/money/issues/650).

Issue #765, opened 2023-12-26
 Incorrect ‘minorUnit’ Value for Iranian Rial (IRR)


AmirMehrabi:
Hello,

I came across this library which solves a big issue for me. I am developing a multilingual application and we needed such a library. Really thrilled to find this.

Anyways, I've come across an issue with the 'minorUnit' value specified for the Iranian Rial (IRR) in the currency array. The current implementation of the library specifies the 'minorUnit' for IRR as 2. This value implies that the currency should be divisible into 100 smaller units, which would be dinars.

However, this is not reflective of the practical usage and financial transactions within Iran. Due to inflation and the devaluation of the Iranian Rial over time, the dinar is no longer used in everyday transactions (I would say it hasn't been used in more than +100 years at least), and prices are not calculated to two decimal places. Instead, prices are rounded to the nearest Rial, and the Toman (which is equal to 10 Rials) is used informally for larger denominations (I would say even Rial is becoming obsolete and Toman has been the default currency for the past +5 decades).

The Issue:

- 'minorUnit' for IRR is set to 2, which is correct in theory but not in line with the practical use of the currency.

Proposed Change:

- The 'minorUnit' value for IRR should be changed to accurately reflect its current use in the economy. It may be more accurate to set the 'minorUnit' to 0 as transactions are no longer fragmented into 100 subunits.

By addressing this, the moneyphp/money library will provide more accurate financial information for developers working with IRR. 

Thank you for taking the time to consider this issue. I look forward to your response and the resolution.

Best regards,
Amir

frederikbosch:
The currencies are derived from the package [moneyphp/iso-currencies](https://github.com/moneyphp/iso-currencies), which uses [this data source](https://www.six-group.com/dam/download/financial-information/data-center/iso-currrency/lists/list-one.xml). You can create your own implementation of the `Currencies` interface to overwrite this.

PR #766, opened 2023-12-26
Update currency.php with the correct minorUnit value for IRR

AmirMehrabi:
This is a proposed fix for #765


frederikbosch:
The currencies are derived from the package [moneyphp/iso-currencies](https://github.com/moneyphp/iso-currencies), which uses [this data source](https://www.six-group.com/dam/download/financial-information/data-center/iso-currrency/lists/list-one.xml). You can create your own implementation of the `Currencies` interface to overwrite this.

PR #767, opened 2024-01-08
Update checks.ymlm

Tunechi2323:


frederikbosch:
What's the function of `check` here?

Tunechi2323:
> What's the function of `check` here?

I'm sorry come again

frederikbosch:
You have created a PR that adds the word `check` inside `checks.yml`. Why do we need that?

Tunechi2323:
My phone was messing up so I had to use my tablet I actually didn't know that

Ocramius:
@frederikbosch this account was created 4h ago and is already spamming everywhere: reporting it.

frederikbosch:
After those two stupid replies, I reported it immediately. Long live AI...

PR #777, opened 2024-01-26
tests: mark test as skippable if locale is not available

Chris53897:
This tests fails for me locally, because i do not have the locale installed.
I have to run tests one by one instead of `composer test`
With this change it makes it easier to test local.


frederikbosch:
But, if in CI, the test must fail if the locale is not there. So this solution is a little to easy.

Chris53897:
ok, i will think about a more robust solution.

Chris53897:
Is it valid to change it to a more commonly used Locale?
I guess US is comma seperated and the sense of the test ist still correct.
Please check this.

frederikbosch:
I rather leave it as is.

PR #779, opened 2024-01-26
feat: add un/serialize for currency

Chris53897:
We use https://github.com/doctrine-extensions/DoctrineExtensions for Versioning.
I came across the problem, that versioning does not work for Curreny.
Symfony via https://github.com/TheBigBrainsCompany/TbbcMoneyBundle

We use this Attribut.

```
#[ORM\Column(options: ['default' => 'EUR'])]
#[Gedmo\Versioned]
public $currency = null;
```

```
public function setCurrency(Currency $currency): self
    {
        $this->currency = $currency;
        return $this;
    }

    public function getCurrency(): Currency
    {
        return new Currency($this->currency);
    }
```

This PR solves the problem for us.
But i am not sure if we use it, how it should be used.
Sideeffekts of this PR?

### DIFF
--- a/src/Currency.php
+++ b/src/Currency.php
@@ -67,7 +67,7 @@ final class Currency implements JsonSerializable, Serializable
         return ['code' => $this->code];
     }
 
-    # TODO Correct PSALM?
+    // TODO Correct PSALM?
     public function __unserialize(array $data): void
     {
         $this->code = (string) $data['code'];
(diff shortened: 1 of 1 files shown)


frederikbosch:
We will not implement. We have parsers and formatters. Regarding ORM mapping: this should be fixed inside the ORM layer.

PR #780, opened 2024-01-27
chore: upgrade PHPUnit 9 => 10

Chris53897:
changes for sebastian/comparator

https://github.com/moneyphp/money/pull/746 rebased against master.
Migrate Config PHPUnit 9 => 10
Only Allow PHPUnit 10.

I did not have a closer look to the comperator changes.

### DIFF
--- a/tests/Calculator/LocaleAwareBcMathCalculatorTest.php
+++ b/tests/Calculator/LocaleAwareBcMathCalculatorTest.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Tests\Money\Calculator;
 
+use function setlocale;
+
 use const LC_ALL;
 
 /** @covers \Money\Calculator\BcMathCalculator */
@@ -13,6 +15,6 @@ final class LocaleAwareBcMathCalculatorTest extends BcMathCalculatorTest
     {
         parent::setUp();
 
-        $this->setLocale(LC_ALL, 'ru_RU.UTF-8');
+        setlocale(LC_ALL, 'ru_RU.UTF-8');
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         "php-http/message": "^1.16.0",
         "php-http/mock-client": "^1.6.0",
         "phpbench/phpbench": "^1.2.5",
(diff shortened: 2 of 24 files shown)


Chris53897:
Tests are failing, because of
DeprecatedMethod: The method PHPUnit\Framework\TestCase::setLocale has been marked as deprecated (see https://psalm.dev/001)

Chris53897:
I replaced it with direct calls to `setlocale()`.
But i am not sure if this is enough.

tests/Calculator/LocaleAwareBcMathCalculatorTest.php
```diff
@@ -13,6 +15,6 @@ public function setUp(): void
     {
         parent::setUp();
 
-        $this->setLocale(LC_ALL, 'ru_RU.UTF-8');
+        setlocale(LC_ALL, 'ru_RU.UTF-8');
```
frederikbosch, line 18: I would add a file called `tests/LocaleTest.php`, with the following content.

```php
declare(strict_types=1);

namespace Tests\Money;

trait LocaleTest
{
    public static function setLocale(int $category, string $locale, \Closure $callback)
    {
        $currentLocale= \setlocale($category, 0);
        try {
            \setlocale($category, $locale);
            $callback();
        } finally {
            \setlocale($category, $currentLocale);
        }
    }
}
```

And then remove `setlocale` from `setUp` but simply use it in the test file as follows

```php
use Tests\Money\LocaleTest;

public static function the_test()
{
    LocaleTest::setLocale(LC_ALL, 'ru_RU.UTF-8', function () {
        // test content here
    });
}

```

frederikbosch (commented):
This looks good. Only with regard to `setlocale` I would solve that differently

tests/Calculator/LocaleAwareBcMathCalculatorTest.php
Chris53897: I followed your suggestion. But i did need to make some changes to make it work.
The name setLocale is already taken with other parameters.

Filename changed from LocaleTest to Locale

I am not sure if i did everythink correct. Please have a look.

PR #783, opened 2024-01-28
Update currencies

slunak:


frederikbosch:
Already done in the last commit. Thanks for noticing though!

PR #793, opened 2024-07-13
:book: Improve descriptions for amount field and getter

andrew-demb:
This will help to know what to expect to get from the `getAmount` method

Actually I mostly copy-pasted description from the constructor


frederikbosch:
Why not, thanks

PR #795, opened 2024-09-11
Allow PHP 8.4

ruudk:
I don’t get it why this library forcefully disallows installation on higher / unsupported PHP versions.

It’s one of the few dependencies that I know that do this. 

Can we please relax this so that people can start running their CI on PHP 8.4 already to spot issues and report / fix them here.


ruudk:
Handled in https://github.com/moneyphp/money/pull/796

PR #796, opened 2024-09-11
Add PHP 8.4 to test matrix

ruudk:
I don’t get it why this library forcefully disallows installation on higher / unsupported PHP versions.

It’s one of the few dependencies that I know that do this. 

Can we please relax this so that people can start running their CI on PHP 8.4 already to spot issues and report / fix them here.

Added PHP 8.4 to the test matrix, and it just works fine. 

These need to be merged and tagged too:
- https://github.com/moneyphp/crypto-currencies/pull/11
- https://github.com/moneyphp/iso-currencies/pull/18

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.1'
+          php-version: '8.3'
           extensions: bcmath, gmp, intl, dom, mbstring
 
       - uses: "ramsey/composer-install@v3"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,7 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         php:
           - '8.1'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,8 +25,9 @@ jobs:
       - name: Set up locales
         run: ./hack/setup-locales.sh
 
-      - name: Download dependencies
-        run: composer update --prefer-stable --prefer-dist --no-interaction --no-progress --no-suggest --prefer-lowest --classmap-authoritative
+      - uses: "ramsey/composer-install@v3"
+        with:
+          dependency-versions: "lowest"
 
       - name: Run tests
         run: composer test
@@ -58,8 +59,7 @@ jobs:
       - name: Set up locales
         run: ./hack/setup-locales.sh
 
-      - name: Download dependencies
-        run: composer install --classmap-authoritative
+      - uses: "ramsey/composer-install@v3"
 
       - name: Run tests
         run: composer test
@@ -78,8 +78,7 @@ jobs:
           php-version: '8.1'
           extensions: bcmath, gmp, intl, dom, mbstring
 
-      - name: Download dependencies
-        run: composer install --classmap-authoritative
+      - uses: "ramsey/composer-install@v3"
 
       - name: Psalm
         run: vendor/bin/psalm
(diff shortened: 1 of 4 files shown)


ruudk:
@frederikbosch this one is ready for review.

composer.json
```diff
@@ -24,7 +24,7 @@
     ],
     "homepage": "http://moneyphp.org",
     "require": {
-        "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
+        "php": "^8.1",
```
frederikbosch, line 27: Please keep the strategy we used before, explicit support of versions. Just add `|| ~8.4.0`

frederikbosch (commented):
Looks okay. One remark, we are using the strategy to explicitly state which PHP versions we support. No auto-support of unknown versions.

composer.json
ruudk: But then I have to do this every year. It's so annoying, like I stated in the OP. This is one of the few packages that does this. Why.

ruudk: Even the creator, Composer, doesn't do this: https://github.com/composer/composer/blob/e4038e5e3bffe22cb815dc1663ea76be96d4222e/composer.json#L25
Or PHPStan https://github.com/phpstan/phpstan-src/blob/2251258cdac7ac0aa36cdf88e5924780bc230f70/composer.json#L8

frederikbosch: I don't think it's annoying. Especially with people looking for early support of newer PHP versions, like yourself, I find it useful that I know for sure that this package is delivering full-support for it. Moreover, it creates a moment where we can reconsider previous decisions. Maybe it's overly conservative, I am fine with that.

.github/workflows/ci.yml
```diff
@@ -58,8 +60,10 @@ jobs:
       - name: Set up locales
         run: ./hack/setup-locales.sh
 
-      - name: Download dependencies
-        run: composer install --classmap-authoritative
+      - uses: "ramsey/composer-install@v3"
+        with:
+          # We need to ignore PHP because Psalm blocks installing it
```
frederikbosch, line 65: I cannot merge with this being part of the PR.

frederikbosch (commented):
We cannot merge with ignore platform reqs.

.github/workflows/ci.yml
ruudk: So what do you suggest?

ruudk: We can remove psalm + the plugin before running tests on PHP 8.4. There is no point for the dependency in that job anyway.

frederikbosch: I suggest we wait until [they resolve the issue](https://github.com/vimeo/psalm/pull/10928). I subscribe for that PR. Once support is there, we can move forward here.

ruudk: But this is not a blocker for PHP 8.4. Psalm is only executed on a separate job that runs with PHP 8.1:
https://github.com/moneyphp/money/blob/76a94ba21bd649adbcbe3d5b0646cadb8da3d458/.github/workflows/ci.yml#L78

frederikbosch: First, it's not a bad idea to change that into the latest stable version (8.3). Secondly, since Psalm uses the strategy as this package, we cannot install dependencies, which means we have to wait for them to solve the issue. I don't see a problem here. We are not even in the RC phase of PHP8.4. We wait.

frederikbosch: > We can remove psalm + the plugin before running tests on PHP 8.4. There is no point for the dependency in that job anyway.

Let's go for this one then.

frederikbosch:
So, we will work from this PR, remove psalm in 8.4 and now we are at it, I would also like to move the default PHP version for Psalm and PHPCS to 8.3.

ruudk:
@frederikbosch Oke, this one is ready. Had to make a few changes, hope you agree.

frederikbosch:
Could you add a phpbench run to a new file bench.yml?

Chris53897:
Thanks

PR #801, opened 2024-11-22
feat: allow PHP 8.4, temp remove Vimeo

Chris53897:
https://github.com/moneyphp/money/pull/796

@ruudk @frederikbosch 
Could you please have a look if this is a solution for a pre-release to help People to move on with PHP 8.4 ?

There is one Deprecation about a vendor-library. but this needs to be fixed there.
1) ...money/vendor/florianv/swap/src/Swap.php:80
Swap\Swap::quote(): Implicitly marking parameter $date as nullable is deprecated, the explicit nullable type must be used instead

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,8 +25,9 @@ jobs:
       - name: Set up locales
         run: ./hack/setup-locales.sh
 
-      - name: Download dependencies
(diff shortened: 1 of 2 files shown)


frederikbosch:
I am willing to pull some PR to a different branch in this project and then release an alpha/beta/RC version (4.6.0-xxx). Would that help?

Chris53897:
I guess that would help

PR #802, opened 2024-11-22
Auto generated TellerFactory

ruudk:
This way, the static typing for Teller works the same as for Money.


frederikbosch:
Nice one, thanks.

PR #803, opened 2024-11-22
Switch from Psalm to PHPStan level 6

### LINKED ISSUE #798: Switch Psalm with PHPStan
Currently, Psalm blocks testing this on PHP 8.4 #796 .

I also wonder why the package uses Psalm, as [PHPStan feels much more alive](https://github.com/phpstan/phpstan-src/pulse/monthly) than [Psalm](https://github.com/vimeo/psalm/pulse/monthly).

ruudk:
Needs https://github.com/moneyphp/money/pull/802

Fixes #798


ruudk:
@frederikbosch I was able to bring it to level 6. I cannot really spend more time on it, but I think this is a good starting point. 

I replaced all `@psalm-` specific prefixes and replaced them with `@phpstan-` specific ones, but ideally, we remove that prefix completely. There is no need for those vendor prefixes anymore, sine it's widely supported by PHPStorm et all.

frederikbosch:
Damn, that must have been quite some work. Fantastic work. I will schedule my review soon.

sagikazarmark:
I'm not really the right person to approve this these days, so I'll just leave a couple emojis here to salute your effort

✅ 🎉 :shipit: 🍾 🥂 ❤️

frederikbosch:
I couldn't find any real code changes that were required for this change. Changes that were included were code removals. So I believe there is no reason why not merging this PR, is there?

src/Currencies/AggregateCurrencies.php
```diff
@@ -47,13 +48,10 @@ public function subunitFor(Currency $currency): int
...
     public function getIterator(): Traversable
     {
-        /** @psalm-var AppendIterator&Traversable<int|string, Currency> $iterator */
         $iterator = new AppendIterator();
 
         foreach ($this->currencies as $currencies) {
-            $currencyIterator = $currencies->getIterator();
-            /** @psalm-var AppendIterator&Traversable<int|string, Currency> $currencyIterator */
-            $iterator->append($currencyIterator);
+            $iterator->append(new IteratorIterator($currencies->getIterator()));
```
ruudk, line 54: @frederikbosch This was the only code change, but I think it's fine, as the test also passes.

ruudk:
@frederikbosch There is only 1 line changed, but I think that's fine. Let's ship this :shipit:

PR #806, opened 2024-12-23
Update Copyright

florianhofsaessC24:
I know - looks like a stupid change.

When I was looking for some kind of lib, I was concerned reading the docs, if the project is maintained. 
GitHub showed me it is.

But I guess it should be visible from the docs as well, that it's not outdated


Jamesking56 (approved):

frederikbosch:
Fixed in 7e86cc541698156633ed02a6858af4fa4c9b43a8

PR #809, opened 2025-03-21
Introduce subtypes of InvalidArgumentException

renanbr:
This PR introduces subtypes of `InvalidArgumentException`.

This improves the log analysis and allow users to rely on type instead of exception message when handling specific errors.

### DIFF
--- a/src/Exception/InvalidArgumentException.php
+++ b/src/Exception/InvalidArgumentException.php
@@ -7,17 +7,26 @@ namespace Money\Exception;
 use InvalidArgumentException as CoreInvalidArgumentException;
 use Money\Exception;
 
-final class InvalidArgumentException extends CoreInvalidArgumentException implements Exception
+/**
+ * @internal
+ */
+class InvalidArgumentException extends CoreInvalidArgumentException implements Exception
 {
     /** @phpstan-pure */
-    public static function divisionByZero(): self
+    public static function divisionByZero(): DivisionByZeroException
     {
-        return new self('Cannot compute division with a zero divisor');
+        return new DivisionByZeroException('Cannot compute division with a zero divisor');
     }
 
     /** @phpstan-pure */
-    public static function moduloByZero(): self
+    public static function moduloByZero(): ModuleByZeroException
     {
-        return new self('Cannot compute modulo with a zero divisor');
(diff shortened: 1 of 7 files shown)


frederikbosch:
I am pretty alone here managing this package, but personally I think this is a good PR. There is no BC-break and I can see why the specific exceptions add value.

@ruudk Since you have actively helped getting PHP8.4 support in, would you be willing to provide some feedback?

src/Exception/InvalidArgumentException.php
```diff
@@ -7,17 +7,26 @@
...
-    public static function divisionByZero(): self
+    public static function divisionByZero(): DivisionByZeroException
     {
-        return new self('Cannot compute division with a zero divisor');
+        return new DivisionByZeroException('Cannot compute division with a zero divisor');
     }
 
     /** @phpstan-pure */
-    public static function moduloByZero(): self
+    public static function moduloByZero(): ModuleByZeroException
```
UlrichEckhardt, line 22: Careful: "module" and "modulo" are not the same!

ruudk (approved):
Looks like a good change to me, except for the typo in ModuleByZeroException class name. From what I can tell, this won't have a BC impact. The newly more specific exceptions that are thrown in Money.php still extend from `\InvalidArgumentException`.

src/Exception/InvalidArgumentException.php
renanbr: handled in b91e101118de04390056431090778134c62cf64c

UlrichEckhardt: Looking better now, no objections from my side!

UlrichEckhardt (approved):

frederikbosch:
There is a PHPStan error. If someone is able to see what is causing this, and resolve it, I can merge it. Otherwise I will have a look myself, but I don't know exactly when I have time for that.

renanbr:
@frederikbosch https://github.com/moneyphp/money/pull/810 may solve the PHPStan error

frederikbosch:
Now this one can be rebased against current master and then merged.

frederikbosch:
Great, thanks for the work. Nice addition.

Issue #807, opened 2025-02-19
Reconsider strict currency comparison for zero amounts


bmoll-axiomconnected:
I'm hoping the decision to enforce strict currency comparison when the amount is zero can be reconsidered. Currently, the library requires both operands to have the same currency for arithmetic operations, even when one of them is 0. This makes it unnecessarily tedious to perform calculations when a value might be null.

For example, given two Money instances, $first and $second, we can simply write:
```php
$total = $first->add($second);
```

However, if either $first or $second could be null, we must provide a default Money instance.

```php
$total = ($first ?? Money::USD(0))->add($second ?? Money::USD(0));
```

This introduces a significant issue:
- We must assume a currency (e.g., USD in the example above).
- If $second is in a different currency (e.g., EUR), the operation will fail, even though adding zero should logically not affect the total.
- This results in additional conditional logic to ensure that a default Money instance has the correct currency, which increases complexity and verbosity:

```php
if ($first !== null) {
    $currency = $first->getCurrency();
} elseif ($second !== null) {
    $currency = $second->getCurrency();
} else {
    $currency = new Currency('USD'); // Arbitrary default, which may be incorrect
}

$total = ($first ?? new Money(0, $currency))->add($second ?? new Money(0, $currency));
```

This complexity arises only because the library enforces currency comparison for zero values. In reality, adding zero does not affect the sum, making this restriction unnecessary. Consider this: If you have 5 M&Ms and you add 0 Skittles, you still have 5 M&Ms, right? The fact that Skittles are a different type of candy doesn’t matter because you’re adding zero of them. The same logic should apply to money—if you add zero of a different currency, the total should remain unchanged, and strict currency enforcement shouldn’t apply in this case.

I understand the importance of strict currency enforcement in general, but relaxing this rule for zero would improve usability without introducing ambiguity. If there is interest from the maintainers, I would be happy to work on a PR.

Side note: It has occurred to me that we could extend the Money class and override the methods we want to, but even this is not a possibility, as Money is a final class.

BenMakesGames:
If adding/subtracting different units is unpalatable, a couple alternatives come to mind surrounding unitless zeros:

- A special `Money::ZERO` object - example use: `$total = $first->add($second ?? Money::ZERO);`
- Allowing adding/subtracting the integer `0` - example use: `$total = $first->add($second ?? 0);`

[...]

PR #811, opened 2025-04-21
Allow operations on Money objects with different currencies when one value is zero

bmoll-axiomconnected:
This PR updates the behavior of Money to allow certain operations to proceed between objects with differing currencies **when one of the operands has a value of zero**.

### Motivation

As outlined in #807, enforcing strict currency equivalence for all comparisons and arithmetic operations—even when one operand is zero—can introduce unnecessary friction, particularly in scenarios involving default or placeholder values (e.g. initial states or empty totals). In these cases, the currency mismatch is inconsequential and may be safely ignored.

### Changes

- Introduced a new private method `isCompatibleCurrency()` which returns true if:
    - Either Money object is zero, or
    - Their currencies match
- Replaced existing strict currency checks in `equals()`, `compare()`, `add()`, `subtract()`, `mod()`, and `ratioOf()` with this new compatibility check
- Adjusted the currency used for return values in `add()` and `subtract()` to favor the non-zero operand’s currency
- Added test cases to validate behavior when one or both operands are zero but have different currencies

### Impact

This change preserves strict currency checking for non-zero values while making operations involving zeros more permissive and ergonomic. This should help avoid unnecessary exceptions in common use cases without compromising correctness.

### ⚠️ Backwards Compatibility Note

This change introduces behavior that may be considered a breaking change: previously, operations involving Money instances with an amount of zero would still enforce strict currency matching and throw an exception on mismatch. With this update, such checks are relaxed for zero values.

While it’s difficult to envision a legitimate use case where an exception should be expected when combining or comparing zero-valued money objects of differing currencies, existing code that implicitly relies on that behavior would be affected.

To support a smoother upgrade path, one potential approach considered was the introduction of an optional `$strictCurrency` argument to the constructor, indicating that the instance should enforce strict currency checks even for zero amounts. This would allow existing behavior to be preserved, however, this was not implemented in this PR to avoid complicating the constructor unless there is clear demand for it.

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -93,13 +93,23 @@ final class Money implements JsonSerializable
     public function isSameCurrency(Money ...$others): bool
     {
         foreach ($others as $other) {
-            // Note: non-strict equality is intentional here, since `Currency` is `final` and reliable.
-            if ($this->currency != $other->currency) {
+            if (! $this->currency->equals($other->currency)) {
                 return false;
             }
         }
 
         return true;
+    }
+
+    /**
+     * Checks whether a Money object has a Currency that is compatible for the sake of comparisons and calculations.
+     *
+     * @phpstan-pure
+     */
+    private function isCompatibleCurrency(Money $other): bool
+    {
+        // @phpstan-ignore possiblyImpure.methodCall
+        return $this->isZero() || $other->isZero() || $this->currency->equals($other->currency);
     }
 
     /**
@@ -109,8 +119,7 @@ final class Money implements JsonSerializable
      */
     public function equals(Money $other): bool
     {
-        // Note: non-strict equality is intentional here, since `Currency` is `final` and reliable.
-        if ($this->currency != $other->currency) {
+        if (! $this->isCompatibleCurrency($other)) {
             return false;
         }
 
@@ -133,8 +142,7 @@ final class Money implements JsonSerializable
      */
     public function compare(Money $other): int
     {
-        // Note: non-strict equality is intentional here, since `Currency` is `final` and reliable.
-        if ($this->currency != $other->currency) {
+        if (! $this->isCompatibleCurrency($other)) {
             throw InvalidArgumentException::currencyMismatch();
         }
 
@@ -204,19 +212,26 @@ final class Money implements JsonSerializable
      */
     public function add(Money ...$addends): Money
     {
-        $amount = $this->amount;
+        $amount   = $this->amount;
+        $currency = $this->currency;
 
         foreach ($addends as $addend) {
             // Note: non-strict equality is intentional here, since `Currency` is `final` and reliable.
-            if ($this->currency != $addend->currency) {
+            if (! $this->isCompatibleCurrency($addend)) {
                 throw InvalidArgumentException::currencyMismatch();
+            }
+
+            // We need to use the currency from the non-zero Money object
+            if (! $addend->isZero()) {
+                // @phpstan-ignore possiblyImpure.methodCall
+                $currency = $addend->getCurrency();
             }
 
             // @phpstan-ignore impure.staticPropertyAccess, possiblyImpure.methodCall
             $amount = self::$calculator::add($amount, $addend->amount);
         }
 
-        return new self($amount, $this->currency);
+        return new self($amount, $currency);
     }
 
     /**
@@ -227,19 +242,25 @@ final class Money implements JsonSerializable
      */
     public function subtract(Money ...$subtrahends): Money
     {
-        $amount = $this->amount;
+        $amount   = $this->amount;
+        $currency = $this->currency;
 
         foreach ($subtrahends as $subtrahend) {
-            // Note: non-strict equality is intentional here, since `Currency` is `final` and reliable.
-            if ($this->currency != $subtrahend->currency) {
+            if (! $this->isCompatibleCurrency($subtrahend)) {
                 throw InvalidArgumentException::currencyMismatch();
+            }
+
+            // We need to use the currency from the non-zero Money object
+            if (! $subtrahend->isZero()) {
+                // @phpstan-ignore possiblyImpure.methodCall
+                $currency = $subtrahend->getCurrency();
             }
 
             // @phpstan-ignore impure.staticPropertyAccess, possiblyImpure.methodCall
             $amount = self::$calculator::subtract($amount, $subtrahend->amount);
         }
 
-        return new self($amount, $this->currency);
+        return new self($amount, $currency);
     }
 
     /**
@@ -289,8 +310,7 @@ final class Money implements JsonSerializable
     public function mod(Money|int|string $divisor): Money
     {
         if ($divisor instanceof self) {
-            // Note: non-strict equality is intentional here, since `Currency` is `final` and reliable.
-            if ($this->currency != $divisor->currency) {
+            if (! $this->isCompatibleCurrency($divisor)) {
                 throw InvalidArgumentException::currencyMismatch();
             }
 
@@ -383,8 +403,7 @@ final class Money implements JsonSerializable
             throw new InvalidArgumentException('Cannot calculate a ratio of zero');
         }
 
-        // Note: non-strict equality is intentional here, since `Currency` is `final` and reliable.
-        if ($this->currency != $money->currency) {
+        if (! $this->isCompatibleCurrency($money)) {
             throw InvalidArgumentException::currencyMismatch();
         }
 
@@ -458,9 +477,12 @@ final class Money implements JsonSerializable
 
     /**
      * Checks if the value represented by this object is zero.
+     *
+     * @phpstan-pure
      */
     public function isZero(): bool
     {
+        // @phpstan-ignore impure.staticPropertyAccess, possiblyImpure.methodCall
         return self::$calculator::compare($this->amount, '0') === 0;
     }
 
--- a/src/Money.php
+++ b/src/Money.php
@@ -216,7 +216,6 @@ final class Money implements JsonSerializable
         $currency = $this->currency;
 
         foreach ($addends as $addend) {
-            // Note: non-strict equality is intentional here, since `Currency` is `final` and reliable.
             if (! $this->isCompatibleCurrency($addend)) {
                 throw InvalidArgumentException::currencyMismatch();
             }
--- a/src/Money.php
+++ b/src/Money.php
@@ -212,25 +212,25 @@ final class Money implements JsonSerializable
      */
     public function add(Money ...$addends): Money
     {
-        $amount   = $this->amount;
-        $currency = $this->currency;
+        $amount            = $this->amount;
+        $currencyReference = $this;
 
         foreach ($addends as $addend) {
-            if (! $this->isCompatibleCurrency($addend)) {
+            if (! $currencyReference->isCompatibleCurrency($addend)) {
                 throw InvalidArgumentException::currencyMismatch();
             }
 
             // We need to use the currency from the non-zero Money object
             if (! $addend->isZero()) {
-                // @phpstan-ignore possiblyImpure.methodCall
-                $currency = $addend->getCurrency();
+                $currencyReference = $addend;
             }
 
             // @phpstan-ignore impure.staticPropertyAccess, possiblyImpure.methodCall
             $amount = self::$calculator::add($amount, $addend->amount);
         }
 
-        return new self($amount, $currency);
+        // @phpstan-ignore possiblyImpure.methodCall
+        return new self($amount, $currencyReference->getCurrency());
     }
 
     /**
@@ -241,25 +241,25 @@ final class Money implements JsonSerializable
      */
     public function subtract(Money ...$subtrahends): Money
     {
-        $amount   = $this->amount;
-        $currency = $this->currency;
+        $amount            = $this->amount;
+        $currencyReference = $this;
 
         foreach ($subtrahends as $subtrahend) {
-            if (! $this->isCompatibleCurrency($subtrahend)) {
+            if (! $currencyReference->isCompatibleCurrency($subtrahend)) {
                 throw InvalidArgumentException::currencyMismatch();
             }
 
             // We need to use the currency from the non-zero Money object
             if (! $subtrahend->isZero()) {
-                // @phpstan-ignore possiblyImpure.methodCall
-                $currency = $subtrahend->getCurrency();
+                $currencyReference = $subtrahend;
             }
 
             // @phpstan-ignore impure.staticPropertyAccess, possiblyImpure.methodCall
             $amount = self::$calculator::subtract($amount, $subtrahend->amount);
         }
 
-        return new self($amount, $currency);
+        // @phpstan-ignore possiblyImpure.methodCall
+        return new self($amount, $currencyReference->getCurrency());
     }
 
     /**
--- a/tests/MoneyTest.php
+++ b/tests/MoneyTest.php
@@ -10,6 +10,7 @@ use Money\Exception\CurrencyMismatchException;
 use Money\Money;
 use PHPUnit\Framework\TestCase;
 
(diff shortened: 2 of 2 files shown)


src/Money.php
```diff
@@ -204,19 +212,26 @@ public function getCurrency(): Currency
...
     public function add(Money ...$addends): Money
     {
-        $amount = $this->amount;
+        $amount   = $this->amount;
+        $currency = $this->currency;
 
         foreach ($addends as $addend) {
             // Note: non-strict equality is intentional here, since `Currency` is `final` and reliable.
-            if ($this->currency != $addend->currency) {
+            if (! $this->isCompatibleCurrency($addend)) {
```
BenMakesGames, line 220: if `$this->amount === 0`, I think this check will ALWAYS pass, giving a bad result?

instead of `$currency`, up top, replace with something akin to:

`$relevantMoneyObject = array_find($addends, (Money $m) => $m->amount !== 0) ?? $this;`

and compare `$relevantMoneyObject->isCompatibleCurrency($addend)`? lines 224-229 would be deleted as well, and 234 would use `$relevantMoneyObject->currency`

bmoll-axiomconnected: > if $this->amount === 0, I think ^this check will ALWAYS pass, giving a bad result?

This behavior is intentional and produces a correct result, not a bad one.

Here’s how currency selection works across all relevant cases:
| $this->amount | $addend->amount | currencies compatible? | currency to use |
| -------- | -------- | --------- | --------- |
| 0 | 0 | always (by definition) | $this |
| 0 | non-zero | always | $addend |
| non-zero | 0 | always | $this |
| non-zero | non-zero | only if currencies match | $addend (same currency as $this) |

The condition `if (! $addend->isZero())` ensures that we update the `$currency` variable to the non-zero operand’s currency, which is appropriate because:
	•	It guarantees the final Money object reflects the meaningful (non-zero) currency.
	•	In case 4, we already verified currencies match, so reassigning is harmless and keeps the logic consistent.

Replace the logic with `array_find(...)` adds overhead without improving correctness. The current approach makes the intent obvious: we scan through the addends, updating the currency when we encounter the first non-zero value. That keeps the logic straightforward, avoids unnecessary function calls, and handles all edge cases explicitly.

BenMakesGames: if `$this->amount === 0`, I believe `$this->isCompatibleCurrency` will pass for _all_ `$addends`, regardless of their currency types.

`0USD add(10USD, 5YEN)` _should_ fail, but I believe the logic written here will return 15YEN

BenMakesGames: here are unit tests to illustrate. the first & third tests fail, because they do not throw exceptions, but _should_ throw for the first dataset from `addingIncompatibleCurrenciesExamples`:

```php
    /**
     * @dataProvider addingIncompatibleCurrenciesExamples
     * @param Money[] $moneysToAdd
     * @test
     */
    public function itThrowsAnExceptionWhenAddingIncompatibleCurrencies(Money $initialMoney, array $moneysToAdd): void
    {
        $this->expectException(CurrencyMismatchException::class);

        $initialMoney->add(...$moneysToAdd);
    }

    /**
     * @dataProvider addingCompatibleCurrenciesExamples
     * @param Money[] $moneysToAdd
     * @test
     */
    public function itDoesNotThrowAnExceptionWhenAddingCompatibleCurrencies(Money $initialMoney, array $moneysToAdd)
    {
        $this->expectNotToPerformAssertions();

        $initialMoney->add(...$moneysToAdd);
    }

    /**
     * @dataProvider addingIncompatibleCurrenciesExamples
     * @param Money[] $moneysToSubtract
     * @test
     */
    public function itThrowsAnExceptionWhenSubtractingIncompatibleCurrencies(Money $initialMoney, array $moneysToSubtract)
    {
        $this->expectException(CurrencyMismatchException::class);

        $initialMoney->subtract(...$moneysToSubtract);
    }

    /**
     * @dataProvider addingCompatibleCurrenciesExamples
     * @param Money[] $moneysToSubtract
     * @test
     */
    public function itDoesNotThrowAnExceptionWhenSubtractingCompatibleCurrencies(Money $initialMoney, array $moneysToSubtract)
    {
        $this->expectNotToPerformAssertions();

        $initialMoney->subtract(...$moneysToSubtract);
    }

    /**
     * @phpstan-return non-empty-list<array{
     *     Money,
     *     Money[]
     * }>
     */
    public static function addingIncompatibleCurrenciesExamples(): array
    {
        return [
            [
                new Money(0, new Currency(self::CURRENCY)),
                [
                    new Money(self::AMOUNT, new Currency(self::CURRENCY)),
                    new Money(self::AMOUNT, new Currency(self::OTHER_CURRENCY)),
                ]
            ],
            [
                new Money(self::AMOUNT, new Currency(self::CURRENCY)),
                [
                    new Money(self::AMOUNT, new Currency(self::CURRENCY)),
                    new Money(self::AMOUNT, new Currency(self::OTHER_CURRENCY)),
                ]
            ],
            [
                new Money(self::AMOUNT, new Currency(self::CURRENCY)),
                [
                    new Money(0, new Currency(self::CURRENCY)),
                    new Money(self::AMOUNT, new Currency(self::OTHER_CURRENCY)),
                ]
            ],
        ];
    }

    /**
     * @phpstan-return non-empty-list<array{
     *     Money,
     *     Money[]
     * }>
     */
    public static function addingCompatibleCurrenciesExamples(): array
    {
        return [
            [
                new Money(self::AMOUNT, new Currency(self::CURRENCY)),
                [
                    new Money(self::AMOUNT, new Currency(self::CURRENCY)),
                    new Money(self::AMOUNT, new Currency(self::CURRENCY)),
                ]
            ],
            [
                new Money(self::AMOUNT, new Currency(self::CURRENCY)),
                [
                    new Money(0, new Currency(self::CURRENCY)),
                    new Money(0, new Currency(self::OTHER_CURRENCY)),
                ]
            ],
            [
                new Money(0, new Currency(self::CURRENCY)),
                [
                    new Money(0, new Currency(self::CURRENCY)),
                    new Money(self::AMOUNT, new Currency(self::OTHER_CURRENCY)),
                ]
            ],
        ];
    }
```

bmoll-axiomconnected: Yes, you are correct. My implementation wasn't accounting for this case:

$this | $addend[0] | $addend[1]
-- | -- | --
0 USD | 0 USD | 100 EUR

It was comparing each addend against `$this`’s currency instead of checking the currency against non-zero items.

I’ve fixed that and added tests to cover currency compatibility expectations in the add and subtract methods.

tests/MoneyTest.php
```diff
@@ -448,6 +449,132 @@ public function itThrowsWhenComparingDifferentCurrencies(): void
...
+            [[Money::USD(100), Money::EUR(0)], false, Money::USD(100)],
+            [[Money::USD(100), Money::USD(100)], false, Money::USD(0)],
+            [[Money::USD(100), Money::EUR(100)], true, null],
+            [[Money::USD(0), Money::USD(100), Money::USD(100)], false, Money::USD(-200)],
+            [[Money::EUR(0), Money::USD(100), Money::USD(100)], false, Money::USD(-200)],
+            [[Money::USD(0), Money::USD(100), Money::EUR(100)], true, null],
+            [[Money::USD(0), Money::EUR(100), Money::EUR(100)], false, Money::EUR(-200)],
+            [[Money::USD(0), Money::EUR(100), Money::USD(100)], true, null],
+            [[Money::USD(100), Money::USD(100), Money::USD(100)], false, Money::USD(-100)],
+            [[Money::USD(100), Money::USD(100), Money::EUR(100)], true, null],
```
bmoll-axiomconnected, line 574: This expected behavior of this last case is somewhat ambiguous, as it can be looked at in two different ways:
1. An exception should be thrown because all values are non-zero, so currency compatibility should be enforced across the board.
2. No exception should be thrown since 100 USD - 100 USD results in 0, and then 0 - 100 EUR simply becomes -100 EUR.

My guess is that the first interpretation is more in line with what most people would expect.

frederikbosch:
I don't know if I like the feature itself nor the implemented solution.

With regards to the implemented solution, I rather have zero checks in all methods than the current `isCompatibleCurrency`. For example, I'd rather see a solution like this.

```php
public function add(Money ...$addends): Money
{
    $amount = $this->amount;

    foreach ($addends as $addend) {
        if ($added->isZero()) {
            // nothing to add, no currency check, skip to next amount
            continue;
        }

        // Note: non-strict equality is intentional here, since `Currency` is `final` and reliable.
        if ($this->currency != $addend->currency) {
            throw InvalidArgumentException::currencyMismatch();
        }

        // @phpstan-ignore impure.staticPropertyAccess, possiblyImpure.methodCall
        $amount = self::$calculator::add($amount, $addend->amount);
    }

    return new self($amount, $this->currency);
}
```

This is way more explicit than the now proposed currency checks, namely it shows into code that zero is an exceptional amount that allows to be handled differently than other amounts.

ben-at-axiom:
> This is way more explicit than the now proposed currency checks, namely it shows into code that zero is an exceptional amount that allows to be handled differently than other amounts.

* always comparing against `$this->currency` won't work: just as the working total's may change at every step, so might the currency. (amount and currency are both components of the running total.)
* `if($amount === 0)`, the loop can also be `continue`d.

bmoll-axiomconnected:
@frederikbosch My approach to `isCompatibleCurrency` accounts for the case where both `$this` and `$addend` are zero. If we only checked whether the addend was zero, we’d miss the scenario where `$this` is zero and the addend is not.

I also made sure `isCompatibleCurrency` was private and had a clearly defined responsibility. That said, I can see why you might prefer to be more explicit in each method. Would something like this work better for you?

```php
public function add(Money ...$addends): Money
{
    $amount = $this->amount;
    $currencyReference = $this;

    foreach ($addends as $addend) {
        // @phpstan-ignore possiblyImpure.methodCall
        if (! $currencyReference->isZero() && ! $addend->isZero() && ! $currencyReference->currency->equals($addend->currency)) {
            throw InvalidArgumentException::currencyMismatch();
        }

        // Use the currency from the non-zero Money instance
        if (! $addend->isZero()) {
            $currencyReference = $addend;
        }

        // @phpstan-ignore impure.staticPropertyAccess, possiblyImpure.methodCall
        $amount = self::$calculator::add($amount, $addend->amount);
    }

    // @phpstan-ignore possiblyImpure.methodCall
    return new self($amount, $currencyReference->getCurrency());
}
```

frederikbosch:
I dont understand any of your comments. Why not skipping currency check all together when the amount is zero, and make it a noop as suggested in the example above?

bmoll-axiomconnected:
I think there's a misunderstanding about the edge case I'm trying to handle. Let me explain more clearly:

There are two zero-amount scenarios this PR aims to handle:

1. When an **addend** is zero (which your solution correctly handles by skipping with `continue`)
2. When `$this` (the current Money object) is zero (which your solution doesn't account for)

Consider this example:

```php
Money::USD(0)->add(Money::EUR(100));
```

In this case, we're adding a non-zero EUR amount to a zero USD amount. With your implementation, this would still throw a currency mismatch exception, because the loop assumes `$this` is always the correct currency anchor. But if `$this` is zero, there’s no meaningful “owning” currency, and we can safely adopt the non-zero addend’s currency.

To clarify, here are the specific behaviors this PR supports:

```php
Money::USD(100)->add(Money::EUR(0)); // evaluates to 1.00 USD
Money::USD(0)->add(Money::EUR(100)); // evaluates to 1.00 EUR
Money::USD(0)->add([Money::EUR(0), Money::EUR(100)]); // evaluates to 1.00 EUR
Money::USD(0)->add([Money::USD(0), Money::EUR(100)]); // evaluates to 1.00 EUR
Money::USD(100)->add(Money::EUR(100)); // throws currency mismatch exception
```

The key idea is this:

- If either side is zero, currency mismatch is irrelevant.
- If both are non-zero, we enforce strict matching (as before).

That’s why the check in my example looks like this:

```php
if (! $currencyReference->isZero() && ! $addend->isZero() && ! $currencyReference->currency->equals($addend->currency)) {
    throw InvalidArgumentException::currencyMismatch();
}
```

It avoids throwing when either side is zero, and falls back to enforcing matching currencies otherwise. I’m happy to inline this logic into each method for clarity (as you’ve suggested), but I want to make sure we’re on the same page about why the check is more involved than just `if ($addend->isZero()) continue;`.

frederikbosch:
@Ocramius do you mind if I ask your opinion on this one?

ben-at-axiom:
> I am pretty convinced it's not a good idea to return a new currency on already instantiated `Money` objects even if its amount is zero.

for me, the code base often has nullable `Money` objects moving around, so for summations it'd be great to be able to null-coalesce into a 0 `Money` objects without worrying "but what currency?" (they _should_ all be the same, so I can _probably_ just use the currency of the first object, but assumptions are how bugs are born). when I'm working in a method which takes an array of nullable `Money` objects, I can't make assumptions about how many, or which ones, may or may not be `null`.

`array_reduce`, with a `Money::AnyCurrency(0)` as the seed value, would be useful in these cases, too.

frederikbosch:
I can see that. Can't we then start with allowing zero of any currency in `Money::sum`, `Money::avg`, `Money::min` and `Money::max`? I still feel that we should not return a new currency from an instance method. So changing with the static methods is an acceptable start.

tscni:
Changing the currency of an existing instance sounds like a terrible idea.
Taking e.g. this example, where `$sum` is intentionally initialized to have a `EUR` currency, it randomly changing to `USD`, instead of raising an error, is exactly the kind of thing I would not want to happen.
```php
function getMoneysFromSomewhere() {
    return [
        rand(0,1) ? Money::EUR(1) : Money::USD(1),
        rand(0,1) ? Money::EUR(1) : Money::USD(1),
    ];
}

$sum = Money::EUR(0);
foreach (getMoneysFromSomewhere() as $money) {
    $sum = $sum->add($money);
}
```

Very similarly, using an aggregate method should really never allow mixing currencies / lead to an unexpected result instead of raising an error.
```php
function getMoneysFromSomewhere() {
    yield rand(0,1) ? Money::EUR(1) : Money::USD(1);
    yield rand(0,1) ? Money::EUR(1) : Money::USD(1);
}

// prevent a negative EUR amount, but we still want EUR
$result = Money::max(Money::EUR(0), getMoneyFromSomewhere());
```

Or just taking the sum example again:
```php
function getMoneysFromSomewhere() {
    return [
        rand(0,1) ? Money::EUR(1) : Money::USD(1),
        rand(0,1) ? Money::EUR(1) : Money::USD(1),
    ];
}

// This should never be anything but EUR
$sum = Money::sum(Money::EUR(0), ...getMoneysFromSomewhere());
```

I'm not certain if I'd even want this to not raise an error. It seems very error prone, especially when considering that these amounts are rarely hardcoded and could just occasionally be 0.
```php
Money::EUR(1)->add(Money::USD($someValueThatMightOccasionallyBe0));
```

I'd argue that the root issue here (making it easier to handle "nothing") should at most be solved with an entirely separate type, that makes it impossible to ever mix currencies. But is it worth the complexity?
It'd actually be interesting to see some real-world use cases with this issue. The PR and issue only seem to highlight some small cases that could be handled much more cleanly without the need for this.

```php
$total = ($first ?? Money::USD(0))->add($second ?? Money::USD(0));
->
$total = $first && $second ? $first->add($second) : ($first ?? $second);

$total = Money::sum(array_map(fn($v) => $v ?? Money::USD(0), [...]));
->
$moneys = array_filter(array_map(fn($v) => $v, [...]));
$total = $moneys ? Money::sum($moneys) : null;

$total = array_reduce([...], fn($c, $v) => ($c ?? Money::USD(0))->add($v ?? Money::USD(0)));
->
$total = array_reduce([...], fn($c, $v) => $c && $v ? $c->add($v) : ($c ?? $v));
```

frederikbosch:
@tscni Thank you, agree, case closed.

PR #817, opened 2025-10-23
Add PHP version 8.5 to CI workflow

ruudk:


frederikbosch:
@ruudk and so we meet again. How about #818? We add a specific job to CI for early versions (and php-src main when there is no new version?) that ignores platform reqs and allows to fail?

frederikbosch:
@ruudk With new setup we always test dev version of PHP, but allow it to fail (`|| true`). This way we could see early on if there are problems with a new version. Once we can add `|| new-version` to composer.json we can add full support to the version and tag a new release. Hopefully this is an improvement over what we had before.

frederikbosch:
@ruudk And v4.8.0 has just been tagged, with support for PHP8.5. Thanks for bringing it up and suggestion to improve the setup.

ruudk:
@frederikbosch Sounds like a great workflow going forward! This makes me really happy.   🎉 Thanks

PR #826, opened 2026-04-30
Fork 202604

kdocki:
Usually when we allocate it is all positive, example break apart $100 ... 50/50 (which is $50 / $50). This fork was made to handle an real world edge case we ran into where need to allocate across "debt & credits owed."

So imagine you have $100 and you want to allocate it [2, -1]? This may seem strange but effectively this is saying, I'll give you $200 and you owe me $100 (mathematically $100 = $200 - $100). It is still a *valid* allocation although a bit of a brain twister until you release that it's dealing with credit/debt so not all the money has been paid between 2 different parties that are settling an account balance.


frederikbosch:
Please include some tests and rebase.

PR #830, opened 2026-06-21
drop useless ext-json requirement

### LINKED ISSUE #829: Drop ext-json from composer.json
Since PHP 8.0, the JSON functionality has been moved into the core, so ext-json is no longer an actual extension.

garak:
Fix #829


frederikbosch:
Thanks.

PR #831, opened 2026-08-31
`roundToUnit()` with single digit may return 100 depending on roundingMode

PowerKiKi:
```php
Money::EUR(1)->roundToUnit(2, Money::ROUND_UP);
```

incorrectly returned 0, instead of 100.

Similar mistakes might have happened with `Money::ROUND_DOWN`.

This PR makes the radical choice to recycle `$this->divide` for `roundToUnit`, instead of re-implementing manually similar calculation. It is much easier to read, and hopefully this is acceptable.

### DIFF
--- a/src/Money.php
+++ b/src/Money.php
@@ -17,9 +17,7 @@ use function filter_var;
 use function floor;
 use function is_int;
 use function max;
-use function str_pad;
-use function strlen;
-use function substr;
+use function pow;
 
 use const FILTER_VALIDATE_INT;
 use const PHP_ROUND_HALF_DOWN;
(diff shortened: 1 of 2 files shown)


sambaptista:
This happens only with 1 digit, and more than 1 decimal.

This also happens with :

```php
Money::EUR(-1)->roundToUnit(2, Money::ROUND_DOWN);
```

Returns 0 instead of -100.

frederikbosch:
@PowerKiKi This indeed makes it much more readable. I don't see why this should not be acceptable. Maybe we should extend the tests and include larger string amounts too.

